PR #6, opened 2016-05-07
mithril-v0.2.4, cycle.js with jsx and some fixes on the reports builder

daiyam:


daiyam:
Hi Stefan,

I've added TSERS 1.0.0.
I've also split the platforms and benchmarks so we can run a single benchmark on a single platform (for testing) and still be able to generate the `table.html`.

I've fixed the bug of mithril. The attributes of a node have to be not `null`. `null` been the default with `jsx`. They won't fix this issue telling me that it's working as intended...

PR #8, opened 2016-07-08
Add Mithril v1pre

pygy:
This is a straight port of the v0.2.5 version. There are a few UglifyJS warnings of no consequences perf-wise (the same than for v0.2.5).

I've updated the main `index.html` and `App.java`, but I may have missed other places where it needs to be added.

I don't have selenium installed, so I didn't run the systematic benchmarks. Still, it is much faster than the stable version in the manual tests I did. Probably too fast in one case: the `clear` benchmark runs in 2ms (for 1000 items) where the vanilla  JS one takes 50ms... I suppose the `done()` function is called before the nodes are removed. Maybe @lhorie can help correcting this.


pygy:
There's at least another issue, when clicking "Append..." => "Create..." => "Update...", the list vanishes. I'll need to revisit this later...

PR #11, opened 2016-07-09
Bobril implementation

Bobris:
I was not able to run selenium benchmark because it looks like expects just Mac. Hoping it will work ok.
index.html is actually regenerated by build, but I commit it anyway.


krausest:
Hi Bobris, thanks for your pull request it's now merged. 
When playing with bobril I found a strange bug - can you help?
Click as fast as you can on the cross icon on the first row to remove and remove as many rows as fast as possible.
At some point it stops removing rows. Any other click also on other button (e.g on clear) after that is ignored.

Bobris:
It is double and multi click detection. These are different events in Bobril than onClick. It is hard to say if it is really bug, no user will probably do that ... But of course I can add event handlers for these as well if you want.

Bobris:
Thought about it, will need to test it more. It could be regression.

krausest:
Preliminary results are [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/f7e3f9336c8b20bea2f2073f8814dee7022bd3bf/webdriver-java/table_update.html)
Bobril looks good!

PR #12, opened 2016-07-10
Aurelia tuning

jdanyow:
* optimized html template
* moved store to separate module and ensured code matched the inferno version

### DIFF
--- a/aurelia-v1.0.0-rc1.0.0/src/app.html
+++ b/aurelia-v1.0.0-rc1.0.0/src/app.html
@@ -7,34 +7,34 @@
                 </div>
                 <div class="col-md-6">
                     <div class="col-sm-6 smallpad">
-                        <button type="button" class="btn btn-primary btn-block" id="run" click.trigger="run()">Create 1,000 rows</button>
+                        <button type="button" class="btn btn-primary btn-block" id="run" click.delegate="run()">Create 1,000 rows</button>
                     </div>
                     <div class="col-sm-6 smallpad">
-                        <button type="button" class="btn btn-primary btn-block" id="runlots" click.trigger="runLots()">Create 10,000 rows</button>
+                        <button type="button" class="btn btn-primary btn-block" id="runlots" click.delegate="runLots()">Create 10,000 rows</button>
                     </div>
                     <div class="col-sm-6 smallpad">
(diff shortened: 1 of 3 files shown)


jdanyow:
thanks for merging @krausest - can't wait to see the updated benchmark results!  Thank you for putting this together.

krausest:
I ran into a strange bug. 
Click on "Create 1,000 rows". Clicking on "Swap rows", "Update every 10th row" or selecting a row by clicking on the text label will not cause a repaint (if you remove a row it'll all get updated). Do you have any idea how to fix this?

jdanyow:
my bad- here's the fix: https://github.com/krausest/js-framework-benchmark/pull/14

krausest:
Thanks, that works (I just changed class-name.bind to class.bind). Can you point me to documentation why this is the right way? Most of all I don't understand why assigning a new object to this.data[i] doesn't cause an update to the binding, but updating the property of that object does.

jdanyow:
I had changed `class.bind` to `class-name.bind` because it performs better.  With `class.bind` aurelia is going to make sure it doesn't clobber any existing classes- for example, if someone had done `<div class="foo" class.bind="bar">`.  This comes at a cost.  Using `class-name.bind` is faster because it blindly assigns the element's `className` property.  There's no docs for this yet (coincidentally I'm currently working on the binding docs so I'll be sure to add this).

Aurelia doesn't use zones or dirty-checking for the most part.  It observes properties by wrapping them in a getter/setter and it observes array/map/set mutation by intercepting calls to methods that would mutate the array/map/set like push/pop/splice/shift.  Indexed assignment of an array can't be observed using this technique (eg `myArray[0] = something`).  I switched all the indexed assignments to the splice equivalent so it would be observed.

Let me know if `class-name.bind` doesn't work.  It will impact the benchmark negatively if you use `class.bind`. 

this will update the aurelia bits but it looks like you have the RC so I'm not sure why class-name.bind isn't working...
```shell
npm install aurelia-animator-css aurelia-bootstrapper aurelia-fetch-client aurelia-framework aurelia-history-browser aurelia-loader-default aurelia-logging-console aurelia-router aurelia-templating-binding aurelia-templating-resources aurelia-templating-router
```

krausest:
It works now - seems that I had a version problem with jspm. I'll report the numbers soon.

PR #16, opened 2016-07-13
add vue.js v2.0.0-beta.1 test

coffeedeveloper:
Copy code from vue-v1.0.26, using the vue.js 2 latest version .
In my test case, very much faster than vue.js 1.


krausest:
Preliminary results are [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/f7e3f9336c8b20bea2f2073f8814dee7022bd3bf/webdriver-java/table_update.html)
Looks very good and the code is nice!

PR #20, opened 2016-07-14
add domvm v1.2.9

leeoniya:
@krausest 

here's a domvm implementation.

sorry, I don't work with webpack/babel/node too much, so feel free to add the needed config files to build the dist; domvm is up on both NPM [1] and Github [2]. i included the full stable build (with router, html generator, observers/ajax helpers). you can do a custom build of just the view core [3] and shave about 30% off the minified size if you want, since nothing else is needed for this bench.

cheers!

[1] https://www.npmjs.com/package/domvm
[2] https://github.com/leeoniya/domvm
[3] https://github.com/leeoniya/domvm#modules-building


krausest:
Thanks! I added webpack and run a comparison with your version (labeled "domvm-nowebpack"). 
Results are [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/69f49f53b4d4573288cb738e01756009a6d9a788/webdriver-java/table_update.html)
Webpack returns an error when running uglify.js on the minimized version so I'm not using an alias and let uglify.js optimized the source build of domvm. 
I'll keep the webpack version in the repository.

leeoniya:
thanks!

boussou:
Great!
I was about to write it.
;-)

boussou:
so DomVM is very good as I expected

PR #21, opened 2016-07-14
Up vidom to 0.3.14

dfilatov:


krausest:
Preliminary results are [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/1f3600fc0721626c3b2e60dfa4e315ede82424bb/webdriver-java/table_update.html)
Clear 2nd time is now much better. For some reasons create rows was a bit slower (but might be random).

dfilatov:
@krausest Thanks!

PR #23, opened 2016-07-14
rewrote to more idiomatic MobX

mweststrate:
Most benchmarks are faster now. Somehow it didn't print the timing results after this modification for `updateData` and `select`. But it seems faster. If that is an issue when running the real benchmark I'll look into that further.

### DIFF
--- a/react-v15.2.0-mobX-v2.3.3/package.json
+++ b/react-v15.2.0-mobX-v2.3.3/package.json
@@ -33,8 +33,8 @@
     "babel-core": "^6.3.15",
     "babel-loader": "^6.2.0",
     "jsx-loader": "^0.13.2",
-    "mobx": "^2.3.3",
-    "mobx-react": "^3.4.0",
+    "mobx": "^2.3.7",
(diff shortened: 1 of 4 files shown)


mweststrate:
The slow clearance (currently the only surprising outlier) will be addressed by MobX 2.4, but I will submit a separate PR for that if needed.

krausest:
Hi Michel, thanks for your PR. I ran the benchmark with your PR, but it didn't perform well. So I rerun with the original code, your PR with mobx 2.3.3 and your PR with mobx 2.3.7
Results are [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/84617b54f444520831e73b878ab6d578734e4327/webdriver-java/table_update.html)
The performance of the original code is consistent with the last published results.
Any ideas?

> Somehow it didn't print the timing results after this modification for updateData and select
This is because componentDidUpdate for Main isn't called anymore. I added the stopMeasure call directly in the two methods.

mweststrate:
Thanks for the heads up! I'll investigate

PR #24, opened 2016-07-15
domvm impl w/subviews & sCU

leeoniya:
@krausest 

the original domvm implementation was simple. this one goes a bit further and componentizes the rows, giving each an sCU condition. it's not pure perf gains across the board but cumulatively should do a bit better i think.

### DIFF
--- a/domvm-v1.2.9/src/view.es6.js
+++ b/domvm-v1.2.9/src/view.es6.js
@@ -31,82 +31,102 @@ function View(vm, store) {
 		}
 	});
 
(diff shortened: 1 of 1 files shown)


krausest:
The result column is labeled domvm v1.2.9-new in the updated [results](https://cdn.rawgit.com/krausest/js-framework-benchmark/fac0c48fcb9f04ed7ed9e48d80d3835fa927c79b/webdriver-java/table_update.html).

leeoniya:
thanks!

hmm, i guess no need to merge if it's slower :/

krausest:
I've reverted the commit.

PR #26, opened 2016-07-17
Updates: Latest results

gyandeeps:
**Results for commit:** 7866d6131055bb958dd0e02092726dd84cad6602
**Output:** https://cdn.rawgit.com/gyandeeps/js-framework-benchmark/latest-results/webdriver-java/table.html

**Fun Facts :** My machine ran for ~8 hrs :sweat_smile: . I would remove cyclejs 6.x version as it took ~2 hrs :scream:

**OS Info:**

* **Type:** Window 10 64-bit
* **Processor:** Intel i7 4500 CPU @1.80ghz
* **RAM**: 8gb


trueadm:
@gyandeeps Are we really including betas on this list now? I don't think it's right to promote something that isn't yet fully stable.

krausest:
@trueadm: I absolutely intend to include betas. Or should we eliminate all 0.x versions also?
(Just joking, inferno is doing a very good job!)
The latest aurelia version also used some not yet documented features, so why not allow an outlook for beta versions?

krausest:
@gyandeeps: Thanks for measuring. I hope you don't mind if i update the table with a version from my MacBook, this would make the historic results more comparable. You certainly had to make lots of modifications for running on windows. I'd love to get a PR for that.

The append rows benchmark is strange. On average my MacBook seems to be faster and it takes 231 msecs for me. The vanilla.JS version on your machine claims to take only 46 msecs, vanillajs2 takes more realistic 275 msecs. What does it print on the console? Have you checked the timeline if that number is valid? Maybe my webdriver timeline hack isn't measuring correctly in that case. 
I guess the geometric mean isn't really authoritative due to that.
I've updated the preliminary results [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/eadb35df43074e10cba30d5a52152b0de3f6886a/webdriver-java/table.html)

gyandeeps:
@krausest Thats fine. I was just playing around on saturday with ur awesome lib. On windows I just had to change the [chrome exe path](https://github.com/krausest/js-framework-benchmark/blob/master/webdriver-java/src/main/java/net/stefankrause/App.java#L23) and [remove the version logic](https://github.com/krausest/js-framework-benchmark/blob/master/webdriver-java/src/main/java/net/stefankrause/App.java#L491-L501) inside app.java file. Nothing special.

I didnt go inside digging deep inside the result folder which gets created. I just saw the final report.

trueadm:
@krausest 0.x versions doesn't always mean they are beta. Inferno is completely stable at 0.7.x :) it's just the public API is likely to change between 0.7.x and 1.0.0. That's fine if you want to support beta versions – thanks again for the benchmark!

leeoniya:
@trueadm semver defines all pre-1.0 as unstable:

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

The fact that and api has not changed in "a while" is not something that can be inferred from any 0.x version. For all intents and purposes, i think 0.7 is in fact your 1.0. Perhaps you're waiting for a "flawless" api to emerge from iteration to tag 1.0, but i don't think this is representative of reality. There will always be some improvement waiting in the pipeline.

trueadm:
@leeoniya Inferno follows semver and 0.7 is a legit usage of it given I plan to break Inferno away from the React chains for 1.0.0 so Inferno can be consumed by other frameworks/libraries including React :)

leeoniya:
it is legit, yes. but saying that you follow semver and stating that 0.7 is "completely stable" is factually incorrect. the fact that thousands of people us this api, i think, makes it a 1.0. it is fine for the next version to be called 2.0 ;)

trueadm:
@leeoniya don't be silly, the next version should be 8.0.0 :P

PR #30, opened 2016-07-25
Fix missing aria-hidden attribute in Inferno

localvoid:


krausest:
Thanks for fixing that.
I've updated the [results](https://cdn.rawgit.com/krausest/js-framework-benchmark/9a70d5e60da8fc96acb5857fa477829819ff53b6/webdriver-java/table.html).
The version with the aria-hidden attribute is called "inferno v0.7.13-2"
Luckily it has almost no influence on the result - I don't think the difference is significant (create 1,000 is a bit faster, but create 10,000 marginally slower).

PR #32, opened 2016-07-25
Cleaned up double MobX submission. Upgraded to MobX 2.4

mweststrate:
Should result in faster clear, other things mostly as-is. Both MobX 2.3.3 and 2.3.7 where in the repo, removed the first.


krausest:
Thanks! Results are [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/8daa626de88b9a6fa49acfd10060e5f7a44cdf22/webdriver-java/table.html)
Indeed much faster for "clear rows" and "clear rows a 2nd time" (~same as react).

krausest:
Updated to react v 15.3 and mobX 2.4.2. [Now](https://cdn.rawgit.com/krausest/js-framework-benchmark/956b068f03ef96a9c08d209ffadb0947cd56edc6/webdriver-java/table.html) the worst cases for clearing rows are gone.

mweststrate:
@krausest thanks! was already wondering why the unmounting was relatively a lot slower than in my other benchmark projects :)

PR #34, opened 2016-07-27
Update and fix classNames in kivi

localvoid:
- update kivi library to 0.11.0-alpha.3, update contains small change to
sync children algorithm, it delays children remove ops and then perform
all of them in a batch.
- fix invalid className in rows, copy-pasted this typo from Inferno
implementation.
- remove unnecessary classNames in menu, didn't know that ids are
required, so instead of unique classNames for event delegation it should
use this ids.

### DIFF
--- a/kivi-v0.11.0/main.ts
+++ b/kivi-v0.11.0/main.ts
@@ -157,7 +157,7 @@ const Row = new ComponentDescriptor<RowProps, void>()
   .update((c, props) => {
     const data = props.data;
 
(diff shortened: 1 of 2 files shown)


krausest:
Thanks for the fix. Results have been [updated](https://cdn.rawgit.com/krausest/js-framework-benchmark/4e47158a10e52122bf4244724cf99b6de4ef7ac1/webdriver-java/table.html).

PR #44, opened 2016-09-02
Add Ractive edge

evs-chris:
Ractive edge is a good bit faster than 0.7.3 in most cases. If you don't mind, I'd like to see edge in the comparison table.

The template and supporting methods aren't exactly ideal examples of Ractive, but I don't suppose it's really possible to collect times _and_ have ideal examples.


krausest:
Nice to see ractive catching up. Results are [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/403df8825e2cc48502c3caa693fdc8ee692fb43b/webdriver-ts/table.html)

evs-chris:
Thanks! Looks like if we can get creation and removal a big chunks of DOM handled a bit better we'll be close to the head of the pack for these types of use-cases.

PR #46, opened 2016-09-10
Update Inferno 0.7.26 with Inferno 1.0.0-alpha3

trueadm:
Removed old Inferno 0.7.26, replaced with Inferno 1.0.0-alpha3. Updated the implementation to be more optimised and removed usage of keys (so loss of DOM state on patching the rows).


trueadm:
@krausest thanks, how does it perform for you?

krausest:
Completely [insane](https://cdn.rawgit.com/krausest/js-framework-benchmark/26367bcbe364ef6a6cde7cf0400009e617805efb/webdriver-ts/table.html) :boom:
Congratulations!
Took me a while to make vanillajs as fast again... The largest difference was when removing a row. Now I shift all rows one up and delete the last one. Directly removing doesn't get near neither to inferno nor vue due to the cost of recalculating the style.

krausest:
I tried with all childrenType attributes removed in the JSX pages. I'd say both perform the same (both score 1.07). Is that what you'd expect?

trueadm:
@krausest the JSX compiler will auto add many of them anyway, I've manually added them to some places where it cannot infer the type. It won't change a lot in performance in your benchmark as there's not a lot of complexity to it and it's not got a deep tree of mixed nodes.

PR #47, opened 2016-09-11
Updated Preact to v6.0.2

struckm:
Noticed that Preact had been upgraded to v5.7.0 recently but it's dependency was still pointing at v4.8.0. Instead of fixing, I just upgraded to latest version.


krausest:
Thanks for spotting that and the pull request!
I'm sorry to say that the results don't look too [good](https://cdn.rawgit.com/krausest/js-framework-benchmark/415e6d2873642e42d4f3896e939eccdf60305357/webdriver-ts/table.html) for preact...

developit:
Looking into it now! Thanks guys :)

PR #50, opened 2016-09-14
Update controller.jsx

eshell:
fixed L95 jsx "this doesn't work here'. works on my machine


eshell:
For mithril v1.0

Ubuntu 16
Chrome

eshell:
I finally saw the problem when using jsx to render components. This doesn't work. 😞

PR #53, opened 2016-09-15
Add DIO

thysultan:
Adding [dio.js](https://github.com/thysultan/dio.js)

slowdown geometric mean

[DIO: 1.11](https://rawgit.com/thysultan/js-framework-benchmark/master/dio-v1.2.1/index.html)
vanillajs: 1.01
inferno v1.0.0-alpha7: 1.08


krausest:
I'm even getting only an slowdown of 1.08 on my machine (1.07 for inferno) - very impressive.
The source code looks this way amazingly simple and short, but I'm afraid doesn't really fit well to the others (build.js won't handle dio correctly). I think I'll convert your code such that it pulls dio.js as a npm dependency, put the javascript code into a separate file and build with webpack if you don't mind (maybe this evening or tomorrow). Thanks for your PR!

thysultan:
@krausest Ok, thanks.

krausest:
I ran into two problems:
1. (major) Deleting the rows doesn't work properly. 
Click on the delete button of the first row. The row is removed. Click on the button of the then first row. It prints "delete took ..." on the console, but the row is not removed. (Happens for dio 1.2.3 from npm also)
2. (minor) I manage to import via webpack works only like that
var dio = require("imports?define=>false!dio.js").dio
Do you know of any simpler way?

thysultan:
@krausest updated the Store Class delete() method, 1. should work now. 2. var {h, dio} = require('dio.js');

krausest:
Sorry to say: Something still seems wrong - maybe I just don't understand. If I add some logging like that
```javascript
				function onDelete () {
					console.log("onDelete ", props.key, props.data.id, props.data.label);
					props.onDelete(props.data.id);
				}
```
and click delete on the first row it prints: "onDelete  1 1 (some text)", then click on the row labeled "2" and it prints the same line with id and key 1! I would expect to see "onDelete  2 2 (some text)". Now click on the row labeled "3" and no selection will show up. Seems to me like the state is not properly updated.

thysultan:
@krausest Thanks, will look into this, must be because event listeners are not diff'ed

thysultan:
@krausest Thanks for catching this, turns out state was being shared between components, anyway this is fixed with v2.0.0.

https://rawgit.com/thysultan/js-framework-benchmark/master/dio/index.html

stats on my machine: dio - 1.10 vs inferno(1.0 alpha) - 1.08

krausest:
@thysultan: Sorry: If I create 1000 rows, delete the second row, then select the row with id 4, the row with id 3 is highlighted.

thysultan:
@krausest thanks, fixed. https://rawgit.com/thysultan/js-framework-benchmark/master/dio/index.html

stats on my machine: dio - 1.07 vs inferno(1.0 alpha) - 1.09

krausest:
Results are now [available](https://cdn.rawgit.com/krausest/js-framework-benchmark/d09dab709d8f81d7593581b9f3734264a6bb2943/webdriver-ts/table.html).
Really not bad!

PR #54, opened 2016-09-18
Angular 2.0.0 + Ahead of Time compilation

robwormald:
This updates the 2.0.0 demo app to Angular 2.0.0 final, and switches on the ahead of time template compiler.


krausest:
Thanks a lot. The size reduction is impressive (from 719 KB to 376 KB uncompressed). Performance seems the same for jit and aot, but slightly slower than rc5. So far I couldn't find the reason (app source code is identical, maybe typescript or webpack related?)
bildschirmfoto 2016-09-18 um 20 13 21

robwormald:
Sent a followup to remove webpack from the equation https://github.com/krausest/js-framework-benchmark/pull/56 

I would expect to see roughly the same performance from AoT vs JiT, as the code is generated using the same internal mechanics (whether at runtime w/ eval or written to disk) - my testing shows a huge reduction in memory allocation for AoT vs JiT. 

Lots more to come - thanks for putting this together!

PR #56, opened 2016-09-18
chore(angular2): move to rollup + closure compiler

robwormald:
makes a nice difference to final bundle size + startup time (@krausest it might be cool to show those two things in the table - bundle(s) sizes + time to interactive?

TODO: currently, we're not minifying the HTML before it hits the precompiler, which leads to a significant amount of extra execution work dealing with text nodes in AoT mode. improves perf significantly, but feels like cheating till its in the pipeline proper :D


krausest:
Bundle size is now 272 KB (old values: 719 KB without AOT, 376 KB with AOT and webpack). Nice.
I had to add typings and yargs as a dependency.
I really consider adding startup time if I find a good metric I can reliably extract from the timeline. The bundle size would be interesting too, but I doubt I'll add that as it would turn into a rollup vs. webpack battle and I don't know if the benchmark is really a good measure for the framework size.
It's interesting to see that it even got faster (from 1.93 slowdown to 1.87)
bildschirmfoto 2016-09-19 um 20 47 12

krausest:
And yes - memory usage is pretty cool too!

PR #62, opened 2016-10-07
polymer-v1.7.0: added

d4l3k:
WIP.

This adds [Polymer](https://www.polymer-project.org/1.0/) benchmarks.

Haven't actually rebuilt the selenium tests yet since my dev box is headless and can't run Chrome.


polymer-v1.7.0/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "js-framework-benchmark-polymer",
+  "version": "1.0.0",
+  "description": "Boilerplate for Polymer",
+  "scripts": {
+    "postinstall": "bower install",
+    "build-dev": "polymer build",
+    "build-prod": polymer build""
```
peterwmwong: Looks like a typo

krausest:
Thanks - certainly an interesting addition.
Merging worked fine, but my testdriver has a hard time running the benchmark due to shadow dom. The driver currently depends on finding the buttons by their id and xpath expressions (and both are hidden by shadow dom).
I found a way to query for the id if restrict the find invocation on the return value of the following method:
function getShadowRootAccess(driver: WebDriver, element: webdriver.WebElement): webdriver.promise.Promise<webdriver.WebElement>  {
  return driver.executeScript('return arguments[0].shadowRoot', element);
}
But for xpath expressions I'm getting the following error:
NotSupportedError: Failed to execute 'evaluate' on 'Document': The node provided is '#document-fragment', which is not a valid context node type.

Dou you have any suggestions for testing shadow dom with webdriver?

d4l3k:
That sounds like a huge pain, sorry! You could try turning off shadow dom, but that seems to slow certain operations down immensely.

I'm assuming you Googled it and found the several stack overflow posts about it (ex http://stackoverflow.com/a/37253205). 

The Polymer thread also isn't very helpful. https://github.com/Polymer/polymer/issues/948

You could try something like
```ts
function getShadowRootAccess(driver: WebDriver, element: webdriver.WebElement, query: string): webdriver.promise.Promise {
return driver.executeScript('return arguments[0].shadowRoot.querySelector(arguments[1])', element, query);
}
```

polymer-v1.7.0/package.json
d4l3k: Fixed

Saulis:
@krausest @d4l3k I'd say the buttons need to go outside the shadow root. The `tap` listeners need to be then configured manually, but shouldn't matter the overall performance in anyway.

krausest:
Thanks I merged polymer and adapted the webdriver to fake id access and xpath expressions for shadow dom.
If I don't get too many complaints I'll keep that ;-)
Preliminary results are [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/bc144b7f5e97d58b75a7b021ccde14ce402d3238/webdriver-ts/table.html)

d4l3k:
Cool! Any reason you squashed the commit and authorship history?

krausest:
@d4l3k Sorry, that wasn't intended. I had a merge conflict and somehow I messed it up.

PR #66, opened 2016-10-17
Add AppRun

yysun:
[AppRun](https://github.com/yysun/apprun) is a library to help implementing the elm/ Redux model-view-update architecture in plain JavaScript or TypeScript code.

Added tests to compare AppRun with other frameworks and libraries.


krausest:
Thanks for the PR - I've found two issues:
* Click create 1000 rows
* Select the row with id 5
* Click swap rows
=> Two rows are highlightes

* Removing a row doesn't work

Due to the bugs I couldn't measure with webdriver-ts yet, but from the console output I'm not too optimistic concerning the performance (e.g. the update each 10th row is as slow as replacing all rows). What kind of performance would you expect?

yysun:
Thanks for reviewing the PR. I have fixed the bugs.

Tested on my machine, I expect its performance is similar to angular 2.0.

krausest:
Hi, here's another issue:
Click "create 1,000 rows", select row with id 1, then delete that row.
Notice that now the row with id 2 is highlighted (which is wrong). Select row with id 3. Now rows with id 2 and 3 are highlighted (also wrong).

krausest:
@yysun Sorry to say: There's another issue with highlighting: Select row with id 2, then delete row with id 1.
After that the rows with id 2 and 3 are highlighted.

yysun:
@krausest I am really sorry wasting of your time on the bugs. I will close this PR for now.
Thank you for this tool. It helps a lot.

PR #68, opened 2016-10-20
Updated Inferno to 1.0.0-beta5

trueadm:


krausest:
Now that chromedriver works again...
Results are [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/d43a0057820aac05073b917eef6317a3bd91fde8/webdriver-ts/table.html)
Once again faster.

PR #70, opened 2016-10-20
domvm v2.0-dev

leeoniya:
Hey @krausest

Here's a domvm v2.0-dev implementation, which should do quite a bit better than the current v1.2.10. Since 2.0 won't be up on NPM till beta I just included a min build of the view layer in this commit.

I'd be interested to see the difference rendered in the perf table, even if you choose not to merge this code till 2.0 is released. Thanks in advance!


krausest:
Had some trouble running the chromedriver on OSX, but results are now [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/d09dab709d8f81d7593581b9f3734264a6bb2943/webdriver-ts/table.html).
Quite impressive. 
I'd like to add domvm 2.0 to the repo when it's available through npm.

leeoniya:
thanks 👍

PR #73, opened 2016-10-31
add domvm-v2.0.0-beta

leeoniya:
@krausest 

Since you test in Chrome, there's no need to transpile this impl down to ES5.

A simple `npm install` will also pull down `domvm`'s pre-built & minified ES5/umd bundles (in `/dist`), which are used directly by `index.html`. Rather than publishing to npm, i've set the `package.json` dependency to point directly to the git repo & commit. This should make updates easier with less collateral noise.

The perf should be improved over https://github.com/krausest/js-framework-benchmark/pull/70

Thanks!

### DIFF
--- a/domvm-v2.0.0-beta/index.html
+++ b/domvm-v2.0.0-beta/index.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>domvm v2.0.0-beta</title>
+  <link href="../css/bootstrap/dist/css/bootstrap.min.css" rel="stylesheet"/>
+  <link href="../css/main.css" rel="stylesheet"/>
+</head>
+<body>
+  <script src='node_modules/domvm/dist/nano/domvm.nano.min.js'></script>
+  <script src='src/store.js'></script>
+  <script src='src/main.js'></script>
+</body>
+</html>
--- a/domvm-v2.0.0-beta/package.json
+++ b/domvm-v2.0.0-beta/package.json
@@ -3,6 +3,6 @@
   "version": "2.0.0-beta",
   "description": "Benchmark for Domvm framework",
   "dependencies": {
-    "domvm": "git://github.com/leeoniya/domvm.git#58fd6d1afd2cadbbd5e098a2e3c655a8b9bf9b51"
+    "domvm": "git://github.com/leeoniya/domvm.git#d8bf362a9f693c321286a07f4d6704a4971f4284"
   }
 }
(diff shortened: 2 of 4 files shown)


krausest:
@leeoniya: I hope you don't mind that I added webpack and es5 transpilation again. I'm not really sure if comparability is better then but unless you're completely unhappy I'd like to keep it. 
I'll post results soon...

leeoniya:
@krausest thanks.

ES5 is probably faster anyhow (unless it does some expensive `function.bind` on each loop iteration to maintain `=>`'s lexical `this`). If you can, plz test to make sure the transpile doesnt introduce perf regressions. As long as updates remain as simple as updating the commit hash, i don't care. My workflow is `rollup -> buble -> gcc`

peterwmwong:
@leeoniya not sure you need to be worried about function.bind as much nowadays: https://twitter.com/trevnorris/status/794028307322785792

leeoniya:
>not sure you need to be worried about function.bind as much nowadays

These improvements reflect v8 5.4, which has been stable for all of 2.5 weeks [1]. Chrome's market penetration is about 50%. The amount of nodejs installs running v7 in production is still tiny as it's just a week old. Unfortunately I still have to write fast code for a world that isnt always running a 2 week old version of the fastest js engine :/

[1] https://googlechromereleases.blogspot.com/2016/10/stable-channel-update-for-desktop.html

krausest:
@leeoniya 
Results for the webpacked domvm 2 are [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/4abc882b98aab67d0c7326cc9a7835488a847991/webdriver-ts/table.html) - impressive!

I'll check if webpacking or transpiling has an impact the next few days.

krausest:
@leeoniya 
I'm glad to report that the impact of webpack and es6 seems negligible.
See domvm2-es6-no-wp column [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/309e3d7cd84cb89613e3ea8ef897d6b77145880a/webdriver-ts/table.html) running on chrome 54.

leeoniya:
@krausest thanks!

Probably no need to keep `2.0.0-dev` and the `2.0.0-es6` versions in the table or repo. `1.2.10` and `2.0.0-beta` should be enough. less clutter is good 👍 

It's interesting that the mem usage went up from `dev` -> `beta`. There were no significant lib changes, but the new bench implementation disabled the full-depth bottom-up `removeChild()` calls. I wonder if the GC's disposal of those deeper dom nodes was negatively affected. I'll trade the small mem increase for the significant speed boost though. I'll need to see if maybe there's something else I missed.

krausest:
@leeoniya 
Concerning the memory update I ran the 2.0-dev version again on current chrome and got the same performance (1.19 on average), but got a memory usage of 3.13 for "ready" and 7.34 for "run". No idea why the former run reported lower numbers (both were on chrome 54.X)

PR #74, opened 2016-11-01
Refactor the Ember 2.6.1 app

mixonic:
Howdy!

This is some cleanup of the Ember 2.6.1 application.

* Drop Ember-Data, it isn't being used
* Simplify how the data is represented via a central service. Just have the component access the service directly instead of passing data from it around.
* Drop welcome-page addon
* Change app.js imports to resolve deprecation warnings.

If this seems reasonable I'll make a new directory for Ember 2.10-beta as well.


krausest:
Results are [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/4abc882b98aab67d0c7326cc9a7835488a847991/webdriver-ts/table.html)
This ember version is much faster than my old one - thanks!

PR #78, opened 2016-11-06
Add Simulacra.js v1.5.5

sapeien:
Adding my own library. Thanks for this benchmark, it helped me discover some things that I could have optimized.

There was a case that I couldn't optimize that was actually due to how Chrome recalculates styles, and I didn't change that: appending 1000 rows after 10,000 rows. It seems to be specific to Chrome and Firefox doesn't have this issue, and the performance characteristics in browsers other than Chrome seem to be quite different across all frameworks and libraries. There were even cases where Simulacra was faster than vanilla in Firefox.


krausest:
Thank - I hope you don't mind I replaced browserify with webpack (mostly due to browserify not creating the dist directory...)

sapeien:
no problem, I couldn't figure out the webpack config which was why I tried browserify.

krausest:
@sapeien 
The last row didn't update for replace all rows benchmark (clicking run twice). I fixed it in 3c1b2fa371566a8d76e6fca3b82c008290d88667.

PR #81, opened 2016-11-19
update domvm-v2.0.0-beta (736c206)

leeoniya:


krausest:
I've updated domvm and rerun the tests. The results are very close to last run, seems like append to 10,000 rows got faster. Is that what you'd expect?

leeoniya:
I didn't expect any improvement, just want to make sure that the perf is representative of the current lib and that nothing has regressed after a round of bugfixes. I don't expect any further perf improvements in domvm's view layer, maybe ever. 10% overhead is an acceptable level for me. I will probably make a final pull request when 2.0.0 is finalized (mostly router work left, which shouldn't impact view perf). There may be some minor incidental improvements, who knows.

I plan to get back to actually building apps with domvm :)

PR #86, opened 2016-11-29
update Vue to 2.1.3 and use build-time template compilation

yyx990803:
In-browser compilation results in less-than-ideal render functions. In production `vue-loader` can apply some build-time optimizations.


anthonyettinger:
curious how vue 2.x performs now.

krausest:
npm run build-prod results in an error on my machine. Any idea how to fix it?
[Here's the npm log](https://github.com/krausest/js-framework-benchmark/files/624729/npm_log.txt)

yyx990803:
Somehow `-c webpack.config.js` in the script is making Webpack bundle itself instead of the app... this might be a bug in webpack's CLI. Fixed.

krausest:
Building works now, but I get an error when I try to remove a row
```main.js:1 Uncaught TypeError: Cannot add/remove sealed array elements(…)```
It works when the sync method doesn't freeze the rows:
```
        sync() {
            this.rows = store.data;
            this.selected = Object.freeze(store.selected);
        }
```
Is it ok to commit it that way?

yyx990803:
Oops, fixed :)

krausest:
Thanks! It's merged now. According to #89 I'll change the requirements which means I'll update vue to use a key in v-for. Feel free to comment regarding the tracking/non-tracking issue.

PR #87, opened 2016-11-30
add Svelte example

Rich-Harris:
Hi – this PR adds version 1.0.1 of [Svelte](https://svelte.technology), which *should* be one of the fastest frameworks here according to my tests locally (though I confess I couldn't get them all to run).

There isn't yet a Webpack loader for Svelte components, so it's using Rollup instead – hope you don't mind me deviating from convention slightly.

Let me know if there's anything else I'd need to do before you could add this. Thanks!

### DIFF
--- a/svelte-v1.0.1/index.html
+++ b/svelte-v1.0.1/index.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Svelte</title>
+    <link href="../css/bootstrap/dist/css/bootstrap.min.css" rel="stylesheet"/>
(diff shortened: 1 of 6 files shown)


solkimicreb:
Interesting project (: The home page is awesome.

krausest:
Thanks! The svelte implementation is currently non-tracking (see issue #89). I didn't find anything about it when skimming through the docs. Do you have any pointers?

Rich-Harris:
@krausest thanks – we don't currently support keyed updates but we soon will: https://github.com/sveltejs/svelte/issues/81

Issue #45, opened 2016-09-09
Will you be interested in my in-depth comparisons between mobx & redux?


lyonlai:
Hi, @krausest . I've done these in branch [benchmark/mobx-vs-redux](https://github.com/lyonlai/js-framework-benchmark/tree/benchmark/mobx-vs-redux) my fork. 8 more columns have been added to the result table. Don't know if you would like a PR for this, as it might goes too much detail into inferno & react, which's a little bit off the intention of this repo.

New comparisons.

* inferno v0.7.26-mobX-v2.4.4
* inferno v0.7.26-redux-v3.5.2
* inferno v0.7.26-redux-v3.5.2-immutable-js-v3.8.1
* inferno v0.7.26-redux-v3.5.2-immutable-js-v3.8.1-pure-render-mixin-15.3.0

* react v15.3.0-mobX-v2.4.2
* react v15.3.0-redux-v3.5.2
* react v15.3.0-redux-v3.5.2-immutable-js-v3.8.1
* react v15.3.0-redux-v3.5.2-immutable-js-v3.8.1-pure-render-mixin

krausest:
What's your conclusion? Seems like neither inferno nor react benefit from mobX/redux for this benchmark.

lyonlai:
You are right. mobX/redux are actually extra layer on top of inferno and react so they won't do any good for the benchmarks. But they are the libraries people will use along with inferno/redux. This is more to see which flux implementation is more efficient under this benchmark. 

The result for mobX vs redux is different when they are used under inferno and react. I've attached it with this comment.

When used under react, mobX is slightly faster than redux (with immutable + pure-render-mixin case) in the following cases:

* partial update
* select row
* swap rows
* remove rows
* append rows to large table

When it comes to inferno, mobX seems to be slightly slower than redux (with immutable + pure-render-mixin case).

![screencapture-localhost-8080-webdriver-ts-table-html-1473809254910](https://cloud.githubusercontent.com/assets/1623058/18495147/d4e0baec-7a5e-11e6-983d-d807f3cf39c1.png)

krausest:
Thought about it for a while ;-) and decided to add redux + immutable.js + pure render mixin version to the list of benchmarks, since redux is quite popular, and this version is the fastest redux version and it's interesting to see that immutable.js doesn't cost much. Results are [here](https://rawgit.com/krausest/js-framework-benchmark/master/webdriver-ts/table.html).

PR #88, opened 2016-11-30
Add Inferno + MobX benchmark

tkrotoff:
Ping @mweststrate @trueadm

### DIFF
--- a/index.html
+++ b/index.html
@@ -28,6 +28,7 @@
                 <li><a href="ember-v2.6.1/dist/"><span class="glyphicon glyphicon-arrow-right" aria-hidden="true"></span> Ember v2.6.1</a></li>
                 <li><a href="ember-v2.10.0-beta.2/dist/"><span class="glyphicon glyphicon-arrow-right" aria-hidden="true"></span> Ember v2.10.0-beta.2</a></li>
                 <li><a href="inferno-v0.7.26/index.html"><span class="glyphicon glyphicon-arrow-right" aria-hidden="true"></span> Inferno v0.7.26</a></li>
+                <li><a href="inferno-v1.0.0-beta9-mobX-v2.6.3/index.html"><span class="glyphicon glyphicon-arrow-right" aria-hidden="true"></span> Inferno v0.7.26 MobX v2.6.3</a></li>
                 <li><a href="kivi-v1.0.0-rc0/index.html"><span class="glyphicon glyphicon-arrow-right" aria-hidden="true"></span> kivi v1.0.0-rc0</a></li>
                 <li><a href="knockout-v3.4.0/index.html"><span class="glyphicon glyphicon-arrow-right" aria-hidden="true"></span> Knockout 3.4.0</a></li>
                 <li><a href="mithril-v0.2.5/index.html"><span class="glyphicon glyphicon-arrow-right" aria-hidden="true"></span> Mithril v0.2.5</a></li>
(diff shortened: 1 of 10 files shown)


trueadm:
@tkrotoff could you possibly use Inferno 1.0.0-beta21? 0.7 is very old now and will soon be ancient :) ping me if you want some help using it.

tkrotoff:
@trueadm I've copy-pasted package.json from the Inferno benchmark => identical versions

trueadm:
@tkrotoff ah sorry, I completely misread the version – just ignore me :) what results did you get out of curiosity?

tkrotoff:
@trueadm results (runned inside a Ubuntu VM while working):

screen shot 2016-11-30 at 16 27 52

I guess the day MobX uses [Proxy object](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Proxy) (see [#652](https://github.com/mobxjs/mobx/issues/652)) it will be faster/use less memory and provide a better API.

tkrotoff:
Any news on this?

nightwolfz:
@tkrotoff With the release of inferno v1, this benchmark should be updated again :)

leeoniya:
@nightwolfz can you guys also submit a keyed implementation?

tkrotoff:
> this benchmark should be updated again

Won't update it if it's not going to be merged...

krausest:
I'm hesitating to merge it (similar to #45, although impact of mobx was smaller then). There are already a very large number of frameworks. What unique benefit do you see by adding inferno + mobx?

tkrotoff:
> What unique benefit do you see by adding inferno + mobx?

I wanted to see if MobX could benefit from Inferno good performance claims (like Inferno+Redux benchmark).

Since MobX is very popular ([221,096 downloads in the last month](https://www.npmjs.com/package/mobx), [first release in 2015](https://github.com/mobxjs/mobx/tags?after=0.1.3)), I guess I'm not the only one interested by this benchmark.
Its NPM popularity is equivalent to Vue.js ([236,387 downloads](https://www.npmjs.com/package/vue), [~2013](https://github.com/vuejs/vue/tags?after=v0.7.6)) and Ember.js ([232,811 downloads](https://www.npmjs.com/package/ember-cli), [~2011](https://github.com/emberjs/ember.js/tags?after=v0.9.2)).

=> so yes, IMHO MobX deserves to get benchmarks with the 2 most popular VDOM implementations compatible with it.


PS:
MobX is more popular than Aurelia ([18,432 downloads](https://www.npmjs.com/package/aurelia-framework)), Inferno ([18,061 downloads](https://www.npmjs.com/package/inferno)), Preact ([16,661 downloads](https://www.npmjs.com/package/preact)), Riot ([13,681 downloads](https://www.npmjs.com/package/riot)), Ractive ([11,317 downloads](https://www.npmjs.com/package/ractive)), Mithril ([5,257 downloads](https://www.npmjs.com/package/mithril)), Bobril ([1,502 downloads](https://www.npmjs.com/package/bobril)), domvm ([261 downloads](https://www.npmjs.com/package/domvm)), Kivi ([152 downloads](https://www.npmjs.com/package/kivi)) and all other experimental VDOM implementations included in this benchmark (that most probably no one will ever use except their authors - no offense here).
If you are afraid there is too many frameworks in your benchmark, you should start here...

Redux is about [1,303,720 downloads last month](https://www.npmjs.com/package/redux).

leeoniya:
i'll put in my two cents. (i know, no one asked)

this benchmark is not a state management lib benchmark. nor is it a popularity contest. the rules of this bench allow you to bring your own datastore implementation, but it only matters insofar as the *improve* performance, not incur overhead (as most of them do).

there are an innumerable quantity of permutations of ui libs + state management libs since many can be used with each other. i would instead argue for also removing the redux variants *unless* they improve perf of the base UI lib...it's likely that none actually will unless their immutability guarantees come super cheap.

there's certainly room for a separate and useful benchmark that shows perf of popular combos of state libs and ui libs, but they don't belong in the same table as the rest, only adding clutter, IMHO.

tkrotoff:
@leeoniya
To sum up: as a VDOM dev you're interested in VDOM implementations benchmarks.

Most of the time end users use a state management lib above a VDOM. So you want your VDOM implementation to perform well with popular state management libs.

leeoniya:
@tkrotoff 
certainly, i'm not unbiased. however, if you were to use MobX with all of the existing libs (includijg vanilla) you would end up with exactly the same table. my point is that it's a fixed overhead which can easily be comminucated with a single statement for all libs: "add X% perf overhead and Y% mem overhead for MobX v2.6.3".

Adding a state management lib adds yet another factor that, while useful to know, is unnecessary makes a straightforward benchmark more cluttered and harder to interpret. I'm not saying it shouldn't exist at all, just not in the same table.

localvoid:
@leeoniya 

>if you were to use MobX with all of the existing libs (includijg vanilla) you would end up with exactly the same table. my point is that it's a fixed overhead which can easily be comminucated with a single statement for all libs: "add X% perf overhead and Y% mem overhead for MobX v2.6.3".

It is actually slightly more complicated. When mobx or redux is added, it is way much harder to optimize code paths that specific to benchmarks and get impressive results. With some libraries, adding mobx can reveal something that their authors are trying to hide when they trying to convince everyone that they've implemented the fastest web ui library ;)

leeoniya:
@localvoid 

> With some libraries, adding mobx can reveal something that their authors are trying to hide

can't tell if you actually have some specific impl in mind or just trolling, with yoir winkie faces ;) this bench is in much better shape now that it has been since the beginning.

krausest:
I'm closing this PR. I don't want to turn this more into a app state benchmark (it's already enough for react).

PR #90, opened 2016-12-01
switch domvm impl to keyed

leeoniya:
feel free to add this into another "keyed" folder if you plan on keeping the non-keyed versions around


krausest:
Thanks - it's merged. I've created the branch "non-keyed" to keep all implementations so far, but will remove all non-keyed implementations from master (kleening up for christmas...)

Issue #89, opened 2016-11-30
Solution to tracking or non-tracking (keyed vs. non-keyed)


krausest:
I'd like to ask for feedback to fix the tracking or non-tracking (keyed vs. non-keyed) issue. Please keep all general discussion out of this issue and put it into issue #22 instead.

**Problem**: Non tracking / non-keyed frameworks are faster than those which track / use keys. Regarding this benchmark non tracking is a correct solution. The "big" frameworks like react and angular can also perform much faster if track by if is disabled as can be seen [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/232d09a20dfb9e5a5c2ef8dcf7da58db4642e4c4/webdriver-ts/table_trackby.html). 
The real problem is that for real life applications there's a big chance that non-tracking will cause wrong results (e.g. for transitions or dom manipulation outside of the framework).

**Goal**: Require all frameworks to use a tracking implementation.

**Proposal**: Add a simple functional requirement: The button to remove a row shall show a zoom transition when holding the mouse button. After releasing the button the transition must be stopped immediately, the next row must not show a zoom effect. 
This is a very simple change that just needs an update of the css (74e7436f4dfc649ef09e079f0a56cc30b6d97ef9) and might make the results a bit less biased and shouldn't influence the measurement via webdriver.

The following frameworks currently appear to be non-tracking :
*  aurelia
* cycle.js
* dio
* domvm
* inferno
* riot
* TSERS
* vanillajs
* vue

(please note that kivi is not included in that list)

What do you think?

solkimicreb:
Hi, I am the author of [NX](https://github.com/RisingStack/nx-framework/). I am new to this community, but here are my 2 cents.

I think the frameworks should adapt to the use case and not the other way around. In this example it means supporting tracked and non-tracked mode too. Frameworks supporting non-tracked mode should naturally use it to perform better (I am surprised they did not do it so far.)

On the other hand changing things now and then would be a pretty good thing. This repo helped me to find and fix performance issues in my framework and I guess others had the same experience too (based on the huge score improvements). I am sure that this change would reveal some other issues and would help me to fix them. Occasional changes would also reveal which frameworks are speedy in general and who puts effort into fixing the possibly occurring performance issues.

TLDR: I think the current benchmark is not biased, but changes are always welcomed to shake up the community a bit.

solkimicreb:
[...]

Issue #35, opened 2016-07-27
50 shades of red


leeoniya:
all red buckets are not created equal, though most green and yellow are. looking at [1], "select row" 13.38ms is the same color as 73.94ms.

it would be useful to have the color scales be tuned to be statistically meaningful within each test. 4300ms and 500ms are very different levels of unacceptable when the baseline is 207.55ms :)

[1] https://cdn.rawgit.com/krausest/js-framework-benchmark/4e47158a10e52122bf4244724cf99b6de4ef7ac1/webdriver-java/table.html

krausest:
Would you like [that](https://cdn.rawgit.com/krausest/js-framework-benchmark/01d859b202efc5f77a0b999781f44adb3f99cfe8/webdriver-java/table.html) better?
Now factors between 1 and 1.5 are absolutely fine and thus colored green,
anything below factor 2.5 is acceptable and thus yellow,
anything else bad which is red.
Yes and why not some shading for in between ;-)
(BTW all values within 1 frame i.e. below 16 msecs are treated equal since I didn't want to penalise frameworks using RAF for repaint.)

krausest:
Or maybe some continuous coloring (factor 2 ist yellow, factor 4 and above is red) like [here](https://cdn.rawgit.com/krausest/js-framework-benchmark/8c2626bfe58de165daa55d8afc5d3381aa7f61c8/webdriver-java/table.html)?

leeoniya:
last version looks great 👍

leeoniya:
So here's a quickie that can be tweaked to adjust the scale. You can run it in console against the generated html table to recolor it. It uses a straight factor without a 16ms threshold, which helps highlight the really fast libs but maybe exaggerates the importance of large factors in sub-frame timings.

```js
// generated via http://www.strangeplanet.fr/work/gradient-generator/?c=16:F44336:FFEB3B:4CAF50
var scale = {
	"#4CAF50": 1.1,
	"#65B74D": 1.2,
	"#7FC04A": 1.3,
	"#98C847": 1.4,
	"#B2D144": 1.5,
	"#CBD941": 1.6,
	"#E5E23E": 1.7,
	"#FFEB3B": 1.8,
	"#FDD63A": 1.9,
	"#FCC139": 2.0,
	"#FAAC39": 2.2,
	"#F99738": 2.4,
	"#F88237": 2.8,
	"#F66D37": 3.2,
	"#F55836": 3.6,
	"#F44336": 4.0,
[...]

PR #96, opened 2016-12-06
Take care of performance even when the duration is less than 16ms

Atry:
This is a benchmark per se.
Even if the current benchmark runs very fast, the real world non-trivial web pages may be many times slower than our benchmarks. Also, the number is still meaningful for slow computers many times slower than your computer.
 
We should create more complicated tests, not simply ignore the result.


leeoniya:
related https://github.com/krausest/js-framework-benchmark/issues/35

krausest:
@Atry 
I'm not going to measure sub 16 msec operations. The only message the benchmark has for those operations is "fast enough". As long as some frameworks perform significantly slower that's just fine for me.

I also don't think measurement accuracy is good enough for safely measuring such fast operations (it would require far more runs than best 6 of 10).

If you have suggestions for more complex tests please add them to #89 -  I'm interested.

PR #99, opened 2016-12-08
update to new ractive edge and use delegated events

evs-chris:
Ractive edge releases are now published to npm with the `edge` tag. This updates the package manifest accordingly. Are fresh modules installed for every benchmark run?

This also swaps from per-row events to delegated events at the table level, which removes the tiny extra browser overhead for calling `addEventListener` and `removeEventListener` 1000 to 10000 times.

### DIFF
--- a/ractive-edge/package.json
+++ b/ractive-edge/package.json
@@ -23,6 +23,6 @@
     "webpack": "1.13.1"
   },
   "dependencies": {
-    "ractive": "ractivejs/ractive#edge"
+    "ractive": "edge"
   }
 }
--- a/ractive-edge/src/main.es6.js
+++ b/ractive-edge/src/main.es6.js
@@ -132,9 +132,18 @@ var ractive = window.r = new Ractive({
         this.set("selected", id);
         stopMeasure();
     },
+    clicked(ev) {
+        var el = ev.original.target;
+        var info = this.getNodeInfo(el);
+        if (el.classList.contains('glyphicon-remove')) {
+            this.remove(info.get('@index'));
+        } else {
+            this.select(info.get('.id'));
+        }
+    },
     el: "#main",
-    template:
-    `<div class="jumbotron">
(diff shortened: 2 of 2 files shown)


evs-chris:
I forgot to mention that 0.9 (edge) will be starting its alpha/beta cycle soon to get ready for release.

krausest:
Sorry for the delay. Here are the results ("ractive-edge2" is the new version):
bildschirmfoto 2016-12-20 um 21 02 01

evs-chris:
I noticed that the preliminary results are now split into keyed and non-keyed tables and wanted to note that Ractive can do either depending on what you're after. If you only use `set`, you'll get the non-keyed behavior where existing DOM is reused and anything left over is thrown away. If you use the Ractive array methods `splice`, `shift`, `push`, `merge`, etc functions, then you'll get the keyed behavior where DOM inserted or removed around the given indicies. The distinction is only _really_ relevant to Ractive when there are transitions involved, as adding and removing items with the array methods will make the nodes transition in and out appropriately.

I don't know if that has any bearing on the results or if this is the best place say it, but I thought it was worth mentioning somewhere :smile:.

leeoniya:
> The distinction is only really relevant to Ractive when there are transitions involved, as adding and removing items with the array methods will make the nodes transition in and out appropriately.

right. it matters in all frameworks when there is some transient in-dom state which must be respected but is not (or cannot be) explicitly handled. css transitions is one example that falls into this category :)

both keyed and non-keyed impls may be submitted now.

krausest:
@evs-chris Thanks I didn't know that. I split the implementation into a keyed and non-keyed implementation since I want to avoid the impression that ractive supports only non-keyed.  I inlined the store into the ractive instance, because I found it easier to use the ractive array methods this way. If you find the time to review the code I'd love to get your feedback (or a PR).

PR #113, opened 2017-01-14
update domvm to v2.0.1

leeoniya:
Hey @krausest

This...

- Removes the deprecated domvm-v1.2.10
- Updates v2.0.0-beta to v2.0.1
- Adds back a non-keyed implementation alongside the keyed
- Switches from `webpack` to `rollup-buble-uglify`, which reduces install/build time considerably - reducing `node_modules` from ~5,100 files to ~692 files and generates a bundle that's 16.5k (down from 17.5k)

thanks!


leeoniya:
@krausest thanks, feel free to kill off `domvm v2.0.0-beta` and `domvm v1.2.10` from the tables. dunno why they're sticky, but this seems to happen frequently - where the table is out of sync with the tests in the repo.

krausest:
:-) I thought it might be useful to have the comparison after the merge so intentionally I kept them for the last time. They will be removed the next time the table is rebuilt.

leeoniya:
ah, makes sense.

didn't know if this was a manual process that can get accidentally overlooked and stick around forever.

PR #116, opened 2017-01-21
Update java version to >= 8

defcc:
Update java version because of lambda expressions used in webdriver-java


krausest:
Thanks - you're right!

PR #117, opened 2017-01-21
updated names frameworks in the file index

renanol:
Updated names files of page JS-Framkeworks Benchmark;
Formated HTML code;


krausest:
Thanks for reporting it. 
Actually I always forgot to update index.html. Thus I created webdriver-ts/createIndex.ts to create index.html automatically. Now I always forget to run createIndex ;-)
(I've updated it now thus I'm not merging your PR but thanks for the heads-up!)

renanol:
No problem =). I think about it too =).

PR #119, opened 2017-01-22
Update NX to beta 1.1.0

solkimicreb:
I updated NX to beta 1.1, it's okay if it can't make it into the next article round. I will try to update more frequently in the future and thanks for updating to alpha 6.1 🙂


krausest:
Thanks. Just in time - it'll be included. BTW does nx support a "keyed"-mode?

solkimicreb:
Thanks! Yes it does, I am testing it and I will send a new PR with it in a few minutes.

PR #124, opened 2017-02-05
Make Binding.scala benchmark keyed (fix #123)

### LINKED ISSUE #123: non-keyed description is not true for Binding.scala
In the section Non keyed results on page http://www.stefankrause.net/js-frameworks-benchmark5/webdriver-ts/table.html:
> In consequence inserting or deleting an element in the data array might append after or delete the last table row and update the contents of all elements after the inserting or deletion index.

It's not true for Binding.scala. Thanks to its [precise data binding](https://github.com/ThoughtWorksInc/Binding.scala#precise-data-binding) mechanism, the DOM managed by Binding.scala always precisely follows the same changes on upstream data.

Atry:


krausest:
@Atry 
Thanks for that version. The swap rows implementation is still non keyed. (Which doesn't matter much in terms of performance.) How could you actually swap the DOM rows in swapRows()?

Atry:
@krausest No problem.

I created #125

Atry:
Hello, @krausest , would you mind updating the web page at http://www.stefankrause.net/js-frameworks-benchmark5/webdriver-ts/table.html ?

krausest:
@Atry 
Preliminary results are always available on this URL: [https://rawgit.com/krausest/js-framework-benchmark/master/webdriver-ts/table.html](https://rawgit.com/krausest/js-framework-benchmark/master/webdriver-ts/table.html)
The page you reported belongs to my blog for round 5 and won't be updated (because the numbers are referenced in the blog), but I plan to publish round 6 the next month or so.

PR #129, opened 2017-02-18
Added Stemjs framework

ericpts:
Everything should work fine.


krausest:
@ericpts Thanks. There's a bug in swap rows Delete the first 5 rows and then click swap rows and you'll see an error in the console (please compare e.g. with the react impl). Can you fix it please?

ericpts:
@krausest Thanks for the heads up!
Everything should be fixed by  	1231767

krausest:
Thanks - it's now merged and results have been [updated](https://rawgit.com/krausest/js-framework-benchmark/master/webdriver-ts/table.html).
Please note that I had to change the classification to non-keyed due to webdriver-ts/src/nonKeyed.ts if you run it with `node dist/nonKeyed.js`it reports:
...
stem-v0.2.60 is keyed for 'run benchmark' and keyed for 'remove row benchmark' non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
ERROR: Framework stem-v0.2.60 is not correctly categorized in commons.ts

Can you change swap rows to really swap the dom table rows?

PR #130, opened 2017-02-19
Added choojs framework

rahulcs:


tunnckoCore:
Cool. I thought that last few days. :)

krausest:
Thanks. I had to change the routing a bit such that it works for "http://localhost:8080/choo-v4.1.0/index.html" - is that okay (thought it may be easier to match any choo.js version).
I'm sorry to say but the [results](https://rawgit.com/krausest/js-framework-benchmark/master/webdriver-ts/table.html) are pretty bad - I run twice, but results are consistent. Did you expect results to be so bad?

rahulcs:
I didn't expect before writing, but while running I could see that it was visibly slow. Not sure why though. :(

tunnckoCore:
maybe @yoshuawuyts

leeoniya:
choo uses morphdom [1], which hasnt been particularly fast in any bench. i think they have improved or are improving since moving to vdom internals. it may be worth just benchmarking morphdom directly.

[1] https://github.com/yoshuawuyts/choo#does-choo-use-a-virtual-dom

tunnckoCore:
Actually, yea. It would be interesting to implement a benchmark with choo using `nanomorph`. Fork choo and edit just some bytes here and there.

But yea.

yoshuawuyts:
My guess without looking at the code would be you're not making use of list reordering and node caching. On phone so can't dig in and say for sure

rahulcs:
@yoshuawuyts not sure i understand the terms `list reordering` and `node caching`?

PR #131, opened 2017-02-20
Add Marionette v3.1.0

paulfalgout:
So this seems to work ok, I think it is working fine, but the results seem too fast in comparison..   Am I doing something wrong?


krausest:
Thanks - I fixed some errors. One error is still there: 'Update every 10th row' doesn't rerender the rows. I didn't find a quick solution for that. Can you fix that please?

scott-w:
Odd, I've just had a play and just doing

```js
events: {
  'change:label': 'render'
}
```

wasn't working. The only way I can get it to work is to do:

```js
initialize() {
  this.listenTo(this.model, 'change:label', this.render);
}
```

and it works.

paulfalgout:
Aha.  It was using an earlier compilation that hadn't failed and I didn't notice.  Anyhow just need to add modelEvents.  I will do that shortly

PR #132, opened 2017-02-20
Add Datum.

MartinRixham:
Hello I've added my data binding library and would be grateful if you would include it in the benchmark.

Thanks


krausest:
Thanks - but the table looks wrong. The rows' colours are not alternating, seems like tbody is repeated for each row. Can you fix that please?

MartinRixham:
Ah ok it might actually be a bug in the library. I'll fix it when I get the chance.

krausest:
Please reopen when it's fixed - thanks!

PR #136, opened 2017-03-05
add angular 4

martinoppitz:
Add Angular 4 to the benchmark.


krausest:
Thanks - actually I tried to update to angular 4 and ran into the same problems as you (typings + rxjs/rollup) but I gave up on rollup... 

I had to change the precompile target back to ngc, due to the following error:
> tsc
src/main.aot.ts(5,36): error TS2307: Cannot find module './app.ngfactory'.

Also I changed the name of index.aot.html back to index.html as a workaround for the benchmark driver in webdriver-ts (it currently always adds an / to the uri - and I didn't want to change that right now...)

You added "node_modules/.bin" to the binaries in package.json in angular 4. It should work fine without it, so I removed it again.

What would you recommend regarding the ngfactory and ngsummary files? Add them to .gitignore?

I'll compare the aot vs the jit version...

krausest:
Here are the [results](https://rawgit.com/krausest/js-framework-benchmark/1271106c70328354a20a0eef29551fd72928363c/webdriver-ts/table.html) for the updated versions. Not much of a difference for angular 4...

krausest:
JIT has indeed a much worse startup behaviour than AOT as can be seen [here](https://rawgit.com/krausest/js-framework-benchmark/0a47a2710193e97e7aa9f2409e7a1118d4a8d512/webdriver-ts/table.html).
The comparison helped me to fix the geometric mean calculation. It didn't include the startup duration correctly.

I ran swap rows again - that negative number looked strange ;-)

PR #139, opened 2017-03-05
update domvm to v2.0.3

leeoniya:
some lib internals have been simplified, allowing the FAST_REMOVE optim that was done manually in the impl to be handled automatically by the lib.

this update is mostly to simplify the user-space impl, i don't expect any perf improvement. hopefully no regressions, either :)

thanks!


krausest:
The non-keyed version is causing some issues (keyed works as expected). Creating or adding rows currently doesn't work for me (no rows are added). I get an error the second time I click on create or add. Can you please check if it works for you?
```
add took 12.680000000000064
bundle.js:1 Uncaught TypeError: Cannot read property '_node' of null
    at H (bundle.js:1)
    at P (bundle.js:1)
    at P (bundle.js:1)
    at P (bundle.js:1)
    at P (bundle.js:1)
    at G._redraw (bundle.js:1)
    at G.redraw (bundle.js:1)
    at bundle.js:1
    at m (bundle.js:1)
    at HTMLButtonElement.<anonymous> (bundle.js:1)
```

leeoniya:
sorry 😬, see #141 for fix (hopefully).

PR #141, opened 2017-03-06
fix domvm non-keyed impl bustage

leeoniya:
fix non-keyed bustage caused by  #139

### DIFF
--- a/domvm-v2.0.3-non-keyed/src/main.es6.js
+++ b/domvm-v2.0.3-non-keyed/src/main.es6.js
@@ -20,7 +20,8 @@ var stopMeasure = function() {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


krausest:
Much better ;-)
[Results](https://rawgit.com/krausest/js-framework-benchmark/11e3779b48463341dd04a2d2c9c5e717d6e1ef24/webdriver-ts/table.html) are as expected quite the same.

leeoniya:
yeah, the vnode preprocessor now incurs an extra check for the presence of lifecycle hooks to mark the entire vtree branch as non-fast-removable [1]. the slight perf hit across the board is likely from this, but worth a less convoluted/fragile impl, IMO.

thanks!

[1] https://github.com/leeoniya/domvm/blob/2.x-dev/src/view/preProc.js#L30

PR #143, opened 2017-03-13
Add ivi

localvoid:
This implementation doesn't have any benchmark specific optimizations, each row is a stateful component, each row is wrapped in redux-like connect component, doesn't have explicit event delegation, doesn't have any special optimization flags ;)

I hope it isn't so slow, just tried to optimize for real workloads, instead of focusing on some narrow optimizations to win benchmarks.

### DIFF
--- a/ivi-v0.5.0/.gitignore
+++ b/ivi-v0.5.0/.gitignore
@@ -1,0 +1,10 @@
+build
+dist
+
+node_modules
+
+.DS_Store
+.idea
+.vscode
+.npm-debug
+npm-debug.log
--- a/ivi-v0.5.0/gulpfile.js
+++ b/ivi-v0.5.0/gulpfile.js
@@ -1,0 +1,47 @@
+const gulp = require('gulp');
+const del = require('del');
+const rollup = require('rollup');
+const rollupReplace = require('rollup-plugin-replace');
+const rollupNodeResolve = require('rollup-plugin-node-resolve');
+const closureCompiler = require('google-closure-compiler-js').gulp();
+
+gulp.task('clean', function () {
+    return del(['build', 'dist']);
+});
+
+gulp.task('bundle', ['clean'], function (done) {
+    return rollup.rollup({
+        format: 'es6',
+        entry: 'src/main.js',
+        plugins: [
+            rollupNodeResolve(),
+            rollupReplace({
+                values: {
+                    __IVI_DEV__: false,
+                    __IVI_BROWSER__: true
+                }
+            })
+        ]
+    }).then(function (bundle) {
+        return bundle.write({
+            format: 'es',
+            dest: 'build/bundle.js'
+        });
+    });
+});
+
+gulp.task('build', ['bundle'], function () {
+    return gulp.src(['build/bundle.js'])
+        .pipe(closureCompiler({
+            js_output_file: 'main.js',
+            compilation_level: 'ADVANCED',
+            language_in: 'ECMASCRIPT6_STRICT',
+            language_out: 'ECMASCRIPT5_STRICT',
+            use_types_for_optimization: true,
+            assume_function_wrapper: true,
+            output_wrapper: '(function(){%output%}).call(this);',
+            warning_level: 'QUIET',
+            rewrite_polyfills: false
+        }))
+        .pipe(gulp.dest('dist'));
+});
--- a/ivi-v0.5.0/index.html
+++ b/ivi-v0.5.0/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>ivi v0.5.0</title>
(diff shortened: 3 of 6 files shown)


krausest:
Should I really [believe](https://rawgit.com/krausest/js-framework-benchmark/7f0a0b5cb5080de840cf8c5b511a3fb5c9943b02/webdriver-ts/table.html) you? ;-)

krausest:
Browser versions differ for the result table - ivi was run on chrome 57, most other on 56. But from looking at the results of vanillajs on chrome 57 this shouldn't change the fact that ivi is already very fast.

localvoid:
>Should I really believe you? ;-)

Just wanted to show that there is no need for all this "optimizations", it just slightly harder to optimize library for "slow path", instead of focusing only on "fast path" :) There are some libraries that has good results with "fast path", but they are actually worse than average vdom library, because they are terrible at "slow path".

And ivi also has normalization, positional indexing (more than half vdom libraries doesn't support it), etc.

leeoniya:
@localvoid aren't you just trading optimizations in the bench impl for automatic optimizations made possible by using an immutable store, which is fine and makes good sense.

also, by positional indexing do you mean by-identity object tracking. from the impl that's kind of what it looks like.

localvoid:
@leeoniya I've used redux-like implementation just to show that it doesn't have any significant overhead, because with some libraries it actually has a significant overhead :)

>also, by positional indexing do you mean by-identity object tracking. from the impl that's kind of what it looks like.

I am talking about use cases like this:

`[1, [a, b, c], 2] => [1, [a, b], 2]` (1,2 - non-keyed, a,b,c - keyed)

or

`[1, 2, 3] => [1, null, 3]` (typical use case with `[1, condition ? 2 : null, 3]`)

In React everything works as expected, all non-keyed nodes are using their position relative to all values at the same depth as keys, and in libraries like vue2 nodes will be recreated, internal state is lost, etc (https://jsfiddle.net/4ppdfyy8/)

localvoid:
@leeoniya I've mentioned positional indexing because `noNormalize` in Inferno disables this feature https://github.com/krausest/js-framework-benchmark/blob/7f0a0b5cb5080de840cf8c5b511a3fb5c9943b02/inferno-v1.3.0-rc3-keyed/src/controller.jsx#L41-L48 :)

But it is not something that is extremely important for vdom library, this problems can be solved just by adding explicit keys to nodes.

thysultan:
Interesting that it performs better than kivi.

localvoid:
@thysultan With 0.7 it is going to be even faster. I've managed to beat kivi at dbmon benchmark with a simple and non-optimized(no hoisting/preallocating arrays/etc) dbmon implementation :)

thysultan:
@localvoid with the features i'm adding to dio i fear it may regress a bit perf wise, hope not by much though... BTW does the `h().events(..).children(..)` pattern also help with mono/ploy-morphic types?

localvoid:
@thysultan

>BTW does the h().events(..).children(..) pattern also help with mono/ploy-morphic types?

yes, it was one of the main reasons why I've started to use method chaining API long time ago.

leeoniya:
@localvoid hmm, that noNormalize looks super-fragile if it has to be maintained by the user. unless it's compiled from a deterministic jsx. it sounds like the positional indexing is needed by libs that pool/reuse existing mutable vnodes? [1]. domvm's nodes are immutable, so i dont think it suffers from this. i've also been able to remove much of userland optimization flags from domvm's impl, making it more "obvious" with less "tricks". i think https://github.com/krausest/js-framework-benchmark/pull/145 should come in around 1.07.

the only perf work i do now in domvm is to avoid regressions as new features trickle in. i'm super happy with <= 1.10; the ROI has essentially plateued for me.

soon we'll need a new baseline that isn't data-driven. where ev handlers do both data mutation and explicit dom ops. this is getting ridiculous :D

[1] https://infernojs.org/docs/guides/optimisations : "This makes it possible to render nested arrays. Because Inferno's virtual nodes are not immutable they might get mutated during application flow."

localvoid:
@leeoniya

>it sounds like the positional indexing is needed by libs that pool/reuse existing mutable vnodes?

It is needed for libraries that remove `null` values and flattens nested arrays during children normalization. I've tried some experiments with moving all this logic into render/patch/etc functions, everything just becomes significantly more complicated and slower.

Libraries that don't support this feature, will lose internal state when nearby children are going from/to `null` value:

Preact: http://codepen.io/localvoid/pen/EZBpoz
vue2: https://jsfiddle.net/4ppdfyy8/

Here is how it works in React, and that is what users are expecting:

React: http://codepen.io/localvoid/pen/VPJBrv/

leeoniya:
@localvoid 

Thanks for the demos, they actually revealed a different domvm reconcilier bug (not with template normalization). It has been fixed and things work as expected: https://jsfiddle.net/upaycoc0/

There's no option in domvm to disable normalization, as this would lead to Very Bad Things (tm).

PR #144, opened 2017-03-15
add surplus-v0.4.0

adamhaile:
Let me know if there are any issues.  Thanks!

### DIFF
--- a/surplus-v0.4.0/index.html
+++ b/surplus-v0.4.0/index.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <title>Surplus 0.4.0</title>
+  <link href="../css/bootstrap/dist/css/bootstrap.min.css" rel="stylesheet"/>
+  <link href="../css/main.css" rel="stylesheet"/>
+</head>
+<body>
+  <div id='main'></div>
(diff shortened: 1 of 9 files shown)


krausest:
Thanks - [impressive](https://rawgit.com/krausest/js-framework-benchmark/09fceb7cf22798b87b37cee2438ba0ee29094cf3/webdriver-ts/table.html)!

adamhaile:
Thanks!  For the merge too!

Just curious, are the vanillajs times from Chrome 56 or 57?  On my box -- an MBP13 running OS 10.11.6 -- I saw an across the board slowdown of about 2-4% on Chrome 57 vs 56.  I'm asking because I usually see surplus coming in about 2-3% behind vanillajs for the keyed version and 1-2% for non-keyed, not the 5% and 7% you found.  I'm wondering if the difference is that surplus times are from Chrome 57 while the vanillajs times are from 56?  

Either way, thanks for running this benchmark!  It's really awesome for framework implementers to be able to compare across a common workload like this.

krausest:
Hi Adam, I'm not really sure where they are from, it could be that non-keyed is from chrome 56 and keyed from 57. Seems like especially the create rows benchmark suffers from chrome 57.
I reran vanillaJS this afternoon and got ~ 132 msecs for both (#145)

I'll see if I find the time to update all results within the next days...

PR #145, opened 2017-03-16
convert domvm-keyed impl to use subviews w/sCU

leeoniya:
this restructures the keyed impl a bit to use subviews w/sCU instead of a monolithic template with `FIXED_BODY` vnode flags. this adds a bit of overhead to the initial creates, but speeds up everything else by reducing the number of vnodes that must be recreated/diffed for each action.

### DIFF
--- a/domvm-v2.0.3-keyed/package.json
+++ b/domvm-v2.0.3-keyed/package.json
@@ -13,6 +13,6 @@
     "rollup-plugin-uglify": "*"
   },
   "dependencies": {
-    "domvm": "git://github.com/leeoniya/domvm.git#146afeb1d5897bf49ca6872887572dc530e41b36"
+    "domvm": "git://github.com/leeoniya/domvm.git#8ab7a791f43d4b0632af15a8e39ad8c199c2fd9a"
   }
 }
--- a/domvm-v2.0.3-keyed/src/main.es6.js
+++ b/domvm-v2.0.3-keyed/src/main.es6.js
@@ -2,14 +2,17 @@ import domvm from '../node_modules/domvm/dist/nano/domvm.nano.min.js';
 
 import {Store} from './store.es6';
 
-var startTime;
-var lastMeasure;
-var startMeasure = function(name) {
+let startTime;
+let lastMeasure;
+
+function startMeasure(name) {
 	startTime = performance.now();
 	lastMeasure = name;
 }
-var stopMeasure = function() {
+
+function stopMeasure() {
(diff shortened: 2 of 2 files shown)


krausest:
Is merged - got a bit [slower](https://rawgit.com/krausest/js-framework-benchmark/dc760a2f4f5ba279b3fde35d14a89f01ef29a721/webdriver-ts/table.html)...
I ran the keyed and nonkeyed version - strange that non-keyed also performed worse, maybe Chrome 57?

leeoniya:
i had to add an extra condition to the old -> new vtree reconciler to fix a bug found found by #143 convo. this affected both keyed and non--keyed though the slowdown did not appear this drastic in Chrome 56. i'll be looking into it. the subview-based approach increases mem usage significantly, so that'll be something to tackle. the updated domvm version also supports ES6 classes, which incurs some overhead in testing isClass [1] for every subview.

any chance of updating vanillajs to be a reliable 1.00 in more cases? i want to start more thoroughly testing  but without having to hunt down each 1.00 and run several benchmarks to establish a bseline.

leeoniya:
@krausest actually the non-keyed regression makes no sense, except "Chrome 57". i didnt touch it at all, and it still uses the original domvm version 😕

leeoniya:
ok, i got the bench running with Win10, Chrome 57 x64, latest domvm in each impl.

vs vanilla, kivi, ivi, i get 1.11 for keyed and 1.14 for non-keyed.

krausest:
I rebooted and ran again for ivi, domvm, react and vanilla:
bildschirmfoto 2017-03-19 um 19 51 41
bildschirmfoto 2017-03-19 um 19 51 49

leeoniya:
huh, dunno what to make of that.

krausest:
Can you confirm the results on windows or do you see different results?

leeoniya:
I ran the benches with "useMinimalCss". I wouldn't compare Windows results with Mac results. JITs/GCs behave differently in different OSs and on different CPUs/GPUs/RAM, etc. you can read a recent long thread with @trueadm where i gave up and basically came to this conclusion.

screen shot 2017-03-19 at 12 21 09-fullpage

leeoniya:
as long as you can get reproducible/reliable results, it doesnt matter what the absolute numbers are on any one system.

leeoniya:
i ran the domvm non-keyed bench vs vanillajs-non-keyed before and after the domvm version update and came up with 1.16 and 1.15 - so basically no change.

Issue #64, opened 2016-10-13
Webdriver doesn't start on Chrome 54


trueadm:
Webdriver doesn't run, instead throwing an error: 

`SessionNotCreatedError: session not created exception`

If you update `chromedriver` to `2.24.1` this issue goes away but I can't get chromedriver to open a window. :/

krausest:
Confirmed. chromedriver 2.24.1 is needed for chrome 54 (https://bugs.chromium.org/p/chromedriver/issues/detail?id=1451)

Bad thing is that I can't even get the slightly modified example from the homepage working:
```
var chromedriver = require('chromedriver');

var webdriver = require('selenium-webdriver'),
    By = webdriver.By,
    until = webdriver.until;

var driver = new webdriver.Builder()
    .forBrowser('chrome')
    .build();

driver.get('http://www.google.com/ncr');
driver.findElement(By.id('lst-ib')).sendKeys('webdriver');
driver.findElement(By.name('btnG')).click();
driver.wait(until.titleIs('webdriver - Google Search'), 1000);
driver.quit();
```

It opens the browser window on my MacBook, navigates to google, but it doesn't get past driver.get with chromedriver@2.24.1 and selenium-webdriver@2.53.3 (same with selenium-webdriver@3.0.0-beta-3)

The same happens with the webdriver-ts. It opens a windows on my machine, but doesn't get any further than loading the benchmark.

Have to investigate further...

krausest:
This java version works with Chrome 54, selenium-chrome-driver 2.53.1, chromedriver 2.24.1
```
package net.stefankrause;

import org.openqa.selenium.*;
import org.openqa.selenium.chrome.*;

public class ChromeTest2 {
    public static void main(String[] args) throws Exception {
        System.setProperty("webdriver.chrome.driver", "node_modules/chromedriver/lib/chromedriver/chromedriver");

        WebDriver driver = new ChromeDriver();
        driver.get("http://www.google.com/ncr");
        WebElement searchBox = driver.findElement(By.id("lst-ib"));
        searchBox.sendKeys("ChromeDriver");
        searchBox.submit();
        Thread.sleep(5000);  
        driver.quit();
    }
}
```

Let's see if I have to reanimate the java test driver.

trueadm:
any more luck on this? It's a bit frustrating as Chrome 54 brought in huge perf improvements in V8 (mainly with memory) and I wanted to see how that reflected.

krausest:
[...]

PR #146, opened 2017-03-20
bump domvm version for non-keyed impl

leeoniya:
@krausest would be cool if the whole table was regenerated with a "reboot" [1] and Chrome 57.

It may be worth adding a warmup cycle at the beginning of the bench to ensure the laptop's CPU is thermally throttled equally for all libs rather than giving those starting with "a" an advantage. I've also found 32-bit versions of Chrome to perform faster, allocate a lot less mem and be more stable on my machine.

[1] https://github.com/krausest/js-framework-benchmark/pull/145#issuecomment-287638169

### DIFF
--- a/domvm-v2.0.3-non-keyed/package.json
+++ b/domvm-v2.0.3-non-keyed/package.json
@@ -13,6 +13,6 @@
     "rollup-plugin-uglify": "*"
   },
   "dependencies": {
-    "domvm": "git://github.com/leeoniya/domvm.git#146afeb1d5897bf49ca6872887572dc530e41b36"
+    "domvm": "git://github.com/leeoniya/domvm.git#8ab7a791f43d4b0632af15a8e39ad8c199c2fd9a"
   }
 }


krausest:
@adamhaile, @leeoniya 
I've rerun all benchmarks on Chrome 57 and updated webdriver and chromedriver.
The one important thing I'm considering to change (and I did it for this run) is not to drop the n worst runs (n was 4 for the results so far). Thus the stddev and the average are generally worse and results are not directly comparable with older ones.
 
Currently chromedriver hangs again occasionally before opening an URL (similar to #64) thus I checked if an update would help - seems like it does not.
I don't think CPU temperature is really a problem - I checked that vue.js numbers stay in a comparable range when the CPU is cold.
On OSX I fortunately don't have to consider choosing between a 32 and 64 bit version, there's only the 64 bit version.

leeoniya:
thanks :)

it would be nice to be able to reproduce the same relative results across platforms & hardware, but i'm pretty confident this will not be possible.

localvoid:
@leeoniya Also, testing on Safari is interesting. Several days ago I've tried an idea with a new data structure for virtual dom (circular linked lists), it gave a nice perf boost in benchmarks on Firefox, almost the same results in Chrome, and huge perf drop on some important cases in Safari. And because mobile Safari is too important to ignore, I reverted all this changes.

localvoid:
@leeoniya Haven't tried on desktop Safari. But on mobile safari, hot code path that goes through existing virtual tree and looking for dirty components without any memory allocations started working ~3x slower. Haven't tried to investigate it any further.

PR #152, opened 2017-03-28
Improved marko and upgraded to 4.2.0

patrick-steele-idem:


krausest:
[Looks](https://rawgit.com/krausest/js-framework-benchmark/52f8005c613e08ed98221082737e4365b035bf8e/webdriver-ts/table.html) much better.
Seems like  partial update, swap rows and remove row gained from mutating the state and a forceUpdate.
Do you know if create 1000 rows and append rows to large table improved due to the version bump or extracting each row into a component?

patrick-steele-idem:
Thanks for asking! I meant to provide more details on the changes...

The improvements probably had very little to do with the changes to mutate the state and the use of `forceUpdate()`.

Splitting out each row into a separate component provided a significant speed boost because Marko skips rerendering a component if its `input` and `state` did not change (shallow compare). Therefore, if a `row` component receives the same input as a previous a previous render (which is what will happen with the partial update or swapping of rows) then it skips rerendering and diffing that portion of the subtree. Skipping rerendering is done at the granularity of a component (for now).

On top of that, we did make some significant performance improvements as part of the `4.2.0` release as a result of your benchmark :) The backstory is that Marko has always used [morphdom](https://github.com/patrick-steele-idem/morphdom) for DOM diffing/patching and it worked really well. However, `morphdom` was a third party module used by other libraries and we couldn't optimize it as much as we would have liked without breaking other apps. With the `marko@4.1.0` release we [forked `morphdom` into `marko`](https://github.com/marko-js/marko/tree/master/morphdom) and made some significant optimizations and got rid of whatever marko didn't require. We knew there was more places we could optimize `morphdom` and after doing more testing with your benchmark we gave that work a higher priority.

Here's how the [changes in `marko@4.2.0`](https://github.com/marko-js/marko/blob/master/CHANGELOG.md#420) improved [our benchmarks](https://github.com/marko-js/isomorphic-ui-benchmarks):

|                | v4.1.3        | v4.2.0        |
|----------------|---------------|---------------|
| search-results | 288 ops/sec   | 355 ops/sec   |
| color-picker   | 5,391 ops/sec | 8,293 ops/sec |

Full results below:

```
Running "search-results"...
Running benchmark "marko"...
marko x 355 ops/sec ±1.20% (58 runs sampled)
Running benchmark "preact"...
preact x 245 ops/sec ±1.24% (57 runs sampled)
Running benchmark "react"...
react x 202 ops/sec ±0.72% (58 runs sampled)
Running benchmark "vue"...
vue x 201 ops/sec ±2.41% (56 runs sampled)
Running benchmark "inferno"...
inferno x 333 ops/sec ±1.15% (57 runs sampled)
Fastest is marko

Running "color-picker"...
Running benchmark "marko"...
marko x 8,293 ops/sec ±1.33% (61 runs sampled)
Running benchmark "preact"...
preact x 2,696 ops/sec ±0.42% (61 runs sampled)
Running benchmark "react"...
react x 2,117 ops/sec ±0.60% (60 runs sampled)
Running benchmark "vue"...
vue x 3,324 ops/sec ±0.85% (61 runs sampled)
Running benchmark "inferno"...
inferno x 5,390 ops/sec ±0.54% (62 runs sampled)
Fastest is marko
```

It's interesting that `marko` beats the other UI libraries by a significant margin in our benchmark (we really didn't try to cheat the benchmark or cater it to some best case for marko...). Your benchmark measures a very specific thing (rendering and modifying lots of rows) and that is helpful, but there is a lot of benefit to having other types of benchmarks. The more benchmarks the better :)

Thank you for maintaining your benchmarks and taking the first steps to add marko!

PR #153, opened 2017-03-29
surplus: bump dep versions; optimize keyed impl some

adamhaile:
A couple tweaks to the surplus version

1) bump dependencies to latest versions.  This was mostly to bring in a bugfix in s-array, but I bumped the others while I was there.

2) a few performance tweaks to the keyed implementation:

    - when replacing rows, do a clear first, to take advantage of fast path for removing nodes

    - use a single computation to set selected row, avoiding creating N computations for N rows (better matches cardinality of state)

With these changes, here are the timings I get for the top four keyed and unkeyed versions.  Curious to see how it comes out for you on your machine this go around.

screen shot 2017-03-28 at 11 25 44 pm
screen shot 2017-03-28 at 11 26 28 pm
 
Thanks!

### DIFF
--- a/surplus-v0.4.0/package.json
+++ b/surplus-v0.4.0/package.json
@@ -23,13 +23,13 @@
   },
   "dependencies": {
     "s-js": "^0.4.0",
-    "s-array": "^0.4.0",
+    "s-array": "^0.4.2",
     "surplus": "^0.4.0"
   },
   "devDependencies": {
     "ts-loader": "0.8.2",
     "typescript": "2.2.1",
-    "surplus-loader": "^0.4.0",
+    "surplus-loader": "^0.4.2",
     "webpack": "2.2.1"
   }
 }
--- a/surplus-v0.4.0/src/controller.ts
+++ b/surplus-v0.4.0/src/controller.ts
@@ -20,9 +20,9 @@ var stopMeasure = function() {
 
 export class App {
     constructor(public store : Store) { }
-    run() {
+    run(clear : boolean) {
         startMeasure("run");
-        this.store.run();
+        this.store.run(clear);
         stopMeasure();
     }
     add() {
@@ -45,9 +45,9 @@ export class App {
         this.store.delete(idx);
         stopMeasure();
     }
-    runLots() {
+    runLots(clear : boolean) {
         startMeasure("runLots");
-        this.store.runLots();
+        this.store.runLots(clear);
         stopMeasure();
     }
     clear() {
--- a/surplus-v0.4.0/src/store.ts
+++ b/surplus-v0.4.0/src/store.ts
@@ -38,7 +38,8 @@ export class Store {
         const idx = (this.data() as any).findIndex((d : Row) => d.id == id);
         this.data.splice(idx, 1);
     }
-    run() {
+    run(clear : boolean) {
+        if (clear) this.data([]);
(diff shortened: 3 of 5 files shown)


leeoniya:
try using a `for` instead of `forEach` ;)

adamhaile:
Heh!  Yeah, that bit definitely isn't optimal.  I left it in because the code was a little cleaner and because it doesn't end up having a measurable effect.  That particular computation only runs when the selection changes, and even a forEach can update the selection on 1k rows within a frame tick.

More generally, I went back and forth about how much optimization to do in the app.  Surplus is a "real DOM" library, so there's sort of a continuum from a naive approach all the way up to task-tuned DOM manipulations.  I.e., I could just keep optimizing and eventually converge on something close to the vanillajs version, but that doesn't seem very useful.  This version tries to follow the best practices we've come up with for performance-oriented S code without going overboard.

adamhaile:
Gave it a shot.  Here are times with a for loop.  Came in ever so slightly faster -- 1.029 -- but not enough to distinguish from the noise.
screen shot 2017-03-29 at 12 34 03 am

From running profiles, the remaining differences between surplus and vanillajs appear to come down to:

1) Object creation, mostly closures and dependency graph nodes.

2) GC, to mark-sweep those extra objects.

3) Startup time.  vanillajs is 7kb, surplus is 20kb, and it just takes longer to load and start 20kb.

I've got some ideas for 3), possibly involving switching to rollup over webpack.  Tree shaking could cut the size down considerably.  I've done some work to minimize object churn, but most of it is for the case where we're re-using objects, while these benchmarks focus mostly on the initial startup.  I've got a few places I can trim down on the number of closures needed, but some of those are critical and can't be worked around.

leeoniya:
regardless, very impressive 👍

krausest:
[Indeed](https://rawgit.com/krausest/js-framework-benchmark/c9a6e8be76318beabdebaea07ad58c2f4959a3d5/webdriver-ts/table.html) - impressive!

adamhaile:
Thanks!  S and surplus started as a side project, then I moved it into some production apps last year, and I just recently decided to "go public" with it.  Really pleased to see how it's doing here!

If you're curious, surplus is a different approach than the other frameworks (or at least the ones I've looked into).  The core is S.js, which implements a fast, general-purpose dependency graph of running code.  In S, data and computations are wrapped in closures, and these closures communicate to record who reads whom.  When data changes, S walks the read graph in reverse to figure out what needs to be updated and what execution order is most efficient.

S is built to keep values up to date, and the idea behind surplus is that the DOM is just another kind of value.  Since the native DOM API is a bit hostile, surplus lets us write JSX, which it translates into inline DOM statements: `var foo = <div></div>` becomes `var foo = document.createElement("div")`, more complicated JSX becomes an IIFE, and so on.  [JSX wasn't my first choice for a syntax, but there's so much tooling around it now that I decided to join 'em rather than fight 'em.]

Dynamic parts of the JSX -- stuff wrapped in { ... } -- are translated into S computations, so as data changes, S automatically re-runs those computations, patching the DOM accordingly.  

There's a cost to construct all these closures and maintain the dependency graph, but the advantage is that when the closures run, they have direct references to the nodes and data.  So the cost is offset by the absence of the 'diff' or 'reconcile' phases used by most vdom implementations.

PR #155, opened 2017-03-30
update domvm v2.1.3

leeoniya:
Hey @krausest 

This makes both the keyed and non-keyed impls near-identical. It should to a bit better in the perf department, too. Shaved almost 1k off the lib size by dropping rarely-used features. Also ran into a very unfun bug where Closure Compiler decided to inline a now-shorter function, which ended up causing a perf regression once it hit the JIT (oddly, only in "swap rows") ; had to add a hack to prevent GCC from inlining it and bloating the caller fn which caused a de-opt.

thanks!


krausest:
@leeoniya I'm currently on a business trip so I'll have to see when I find a good opportunity to update the results.

leeoniya:
no rush, thanks!

krausest:
Thanks for your patience. [Results](https://rawgit.com/krausest/js-framework-benchmark/ec23b988d6e132b347cb8119515eebcb445cc9f7/webdriver-ts/table.html) have been updated.

leeoniya:
cool, those align with expectations.

PR #156, opened 2017-04-03
update Aurelia

Gheoan:


krausest:
Thanks. It seemed like minor changes, but startup time has become much [faster](https://rawgit.com/krausest/js-framework-benchmark/ec23b988d6e132b347cb8119515eebcb445cc9f7/webdriver-ts/table.html). Did you expect that?

Gheoan:
Yes, 5dd6b77 removed an unnecessary library, reducing the time spent parsing scripts, while 76e34fe removed the call that configures the router plugin, reducing the time spent running scripts.

PR #157, opened 2017-04-03
Update Inferno

Havunen:
Update to stable version of Inferno


krausest:
Thanks - quite a bit faster than inferno 1.3.0. Results are [here](https://rawgit.com/krausest/js-framework-benchmark/ec23b988d6e132b347cb8119515eebcb445cc9f7/webdriver-ts/table.html)

PR #158, opened 2017-04-04
submit pico-dom-v0.18.0 keyed

hville:
Here is a fast minimal lib (2.3 kb gzip) with low memory footprint to facilitate the creation of components and lists of components. On my machine, the memory use and overall slowdown is close to svelte. The overall minimized code is also quite close to the compiled svelte code.

More a library than a framework so I will let you decide if it is relevant here:
* all hyperscript functions (definitely not as sexy as other entries)
* no intermediate structure, all direct dom operations
* top-down updates (updating a node cascades down the whole tree)

This is the *keyed* version, the *non-keyed* can be obtained by deleting the [key here](https://github.com/hville/js-framework-benchmark/blob/master/pico-dom-v0.18.0/src/table.js#L32). I can add the other version if there is any interest.

Thanks for the project. Trying the tests made me realize how much involvement it must require to keep all these tests up to date.

### DIFF
--- a/pico-dom-v0.18.0/index.html
+++ b/pico-dom-v0.18.0/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
     <meta charset="utf-8"/>
-    <title>VanillaJS-"keyed"</title>
+    <title>picoDOM-"keyed"</title>
     <link href="/css/currentStyle.css" rel="stylesheet"/>
 </head>
 <body>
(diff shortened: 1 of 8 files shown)


krausest:
Thanks, it's quite fast - results are [here](https://rawgit.com/krausest/js-framework-benchmark/ec23b988d6e132b347cb8119515eebcb445cc9f7/webdriver-ts/table.html)

PR #159, opened 2017-04-06
Add Datum.

MartinRixham:
Hi I've implemented the benchmark with the latest version of Datum and fixed the zebra striping.


krausest:
@MartinRixham 
Thanks - it's now committed. I'm afraid to say it seems like it's not really [fast](https://rawgit.com/krausest/js-framework-benchmark/0ae7bff9d754380636b1227c04c135105245148e/webdriver-ts/table.html). Did you expect that?

MartinRixham:
Thanks a lot. It is a bit slower than I was hoping. Speed isn't my only design goal, but I might try and optimise it a bit. Any tips would be appreciated.

PR #160, opened 2017-04-07
Update major version of Simulacra

daliwali:
This fixes a minor breaking change and includes a performance optimization specific for this benchmark: the preference of `appendChild` over `insertBefore`, which was the most significant slowdown.


krausest:
@daliwali 
I see two issues: Selecting a row has no effect and removing a row throws "Cannot read property 'splice' of undefined". Does it work for you?

daliwali:
@krausest sorry about that, I just fixed it and squashed the commit.

also it seems that there is still more optimization I have to do in order to make appending faster 🤔

krausest:
It's committed - though it turned out to be much harder than I expected. Nice speedup as can be seen [here](https://rawgit.com/krausest/js-framework-benchmark/8c6b10411089b2f28e174f495dc4ae9f193f2d03/webdriver-ts/table.html).

I changed two smaller things in your code:
- The remove icon is appended after the table to preload the icon, otherwise there are two repaints for the first time
- For the console logging I switched from requestAnimationFrame to window.setTimeout. The RAF is executed before the repaint and thus reports a too short duration.
bildschirmfoto 2017-04-16 um 21 58 50
window.setTimeout is executed after repainting and is a better approximation:
bildschirmfoto 2017-04-16 um 22 00 00

All of this wasn't problematic. It turned out that my test driver still had some issues which were revealed by your submission and thus helped to improve the test driver. The "create many rows" benchmark was reported as too fast (a negative duration...) and it took me quite some time nailing down that issue.
I had lots of tracing categories switched on and not all performance logging entries were returned on the driver.manage().logs() call. When I finally found that out I removed unneeded tracing categories. 
Sadly I had to find out that chromedriver 2.29.0 doesn't report any performance logs for selenium-webdriver 3.3.0 (though it seems to work with the old java testdriver) -I'll wait to see if an update to either fixes that issue....

PR #161, opened 2017-04-17
Added glimmer 0.3.9

patocallaghan:
Add [Glimmer](https://glimmerjs.com) to the benchmark


daliwali:
Just a side note: this implementation seems to call `requestAnimationFrame` on every frame and run some function that takes a few milliseconds to complete, this would be very bad when you're on battery since it's using CPU cycles for nothing. It only happens when there are rows visible, but not when it's empty. The benchmark does not penalize this behavior.

patocallaghan:
@daliwali that's interesting. I don't call `requestAnimationFrame` in my code anywhere. Looks to be internal to the glimmer application lib https://github.com/glimmerjs/glimmer-application/blob/master/src/application.ts#L160

krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/cc3c80c92932f6fd79c650cc53cb97a9c25d5c0b/webdriver-ts/table.html)
I really expected much better results...

krausest:
@chadhietala May I ping you and ask for help? Is that the fastest way for glimmer?

adamhaile:
Looks like @tomdale recently fixed the rAF looping issue, post 0.3.9: glimmerjs/glimmer-component@d011784c8767dd1368cbadd55dd3a3138ef2dbce

Running under a profiler, I'm seeing calls to functions with 'DevMode' in their name, like this: https://github.com/glimmerjs/glimmer-component/blob/master/src/tracked.ts#L251 .  I don't, however, see any way to tell the code not to do that?

[Note: In case it's not obvious, I'm not in any way a Glimmer or Ember expert.  Just curious about the new tech.]

chadhietala:
There are tons of instrumentation and runtime checks that should not be happening when you build you build for production. They are currently not stripped automatically, even so these numbers seem suspect as Ember is using the same rendering engine yet the numbers are way off.

wycats:
In general, if you're getting way worse numbers for Glimmer than Ember, something is wrong.

In this case, it's two things:

1. A bug in @glimmer/component 0.3, that could in some cases trigger infinite rerenders. That bug was fixed in https://github.com/glimmerjs/glimmer-component/commit/d011784c8767dd1368cbadd55dd3a3138ef2dbce but hasn't yet been released.
2. Production builds are currently broken, including far too much dev-mode-only content. This should also be fixed in the near future.

In the meantime, if Glimmer performs worse than Ember, it's almost assuredly something broken at the surface (since, as @chadhietala, Ember shares almost 100% of the rendering code). Please reach out if you get results like this so we can make sure we address whatever is causing the weirdness in the very near term.

PR #162, opened 2017-04-18
add rx-domh v0.0.2

xialvjun:


krausest:
Results are [here](https://rawgit.com/krausest/js-framework-benchmark/cc3c80c92932f6fd79c650cc53cb97a9c25d5c0b/webdriver-ts/table.html)
I'm afraid to say it turned out pretty slow...

xialvjun:
😂，Well， I know it's pretty slow. But I think its thoughts is intresting.
It's indeed slow. ┑(￣Д ￣)┍....

PR #164, opened 2017-04-20
Update to glimmer-v0.3.10.

rwjblue:
v0.3.9 had a few issues that were addressed:

* included development assertions (which were supposed to be stripped in production builds)
* had a gnarly bug that forced it to rerender infinitely

Both of those issues are addressed with this update, and running a partial set of the benchmarks shows things are a bit better:

screen shot 2017-04-20 at 11 52 48 am


patocallaghan:
@rwjblue this file will also need to be updated to point the glimmer link to 0.3.10 instead of 0.3.9 https://github.com/krausest/js-framework-benchmark/blob/master/index.html#L1

rwjblue:
Awesome, thank you @patocallaghan, updated!

krausest:
Thanks - looks much better as can be seen [here](https://rawgit.com/krausest/js-framework-benchmark/28eed78e6a5cf1e10559e4e8de1fd48ef57e1462/webdriver-ts/table.html).
(Ping @chadhietala @wycats as 0.3.9 was disappointing)

leeoniya:
`204.95` for `clear rows` is unusually fast.
even v0.3.9's `225.77` is unusually fast.

what kind of sorcery is this? 😉

adamhaile:
@leeoniya, heh, I was wondering the same thing.

Running on my box, the fast run doesn't repro.  Glimmer comes in at 1.09 for clear10k, relative to vanillajs at 1.00.  Maybe just an outlier?

Glimmer's using a pretty vanilla removeChild() and nextSibling() for clearing a range:
```javascriopt
function clear(bounds) {
    var parent = bounds.parentElement();
    var first = bounds.firstNode();
    var last = bounds.lastNode();
    var node = first;
    while (node) {
        var next = node.nextSibling;
        parent.removeChild(node);
        if (node === last) return next;
        node = next;
    }
    return null;
}
```
Dunno.

leeoniya:
maybe it's a case of @krausest's machine after a fresh reboot? [1]

D:

[1] https://github.com/krausest/js-framework-benchmark/pull/145#issuecomment-287638169

krausest:
Not much of a reboot, but a chrome update to 58 might explain that. Came out yesterday and only glimmer 0.3.10 ran on it. Time to update all results and see if that looks more consistent.

krausest:
Seems like that. Test driver reports 203.01  for Chrome 58 and vanillajs-keyed.

krausest:
Updated [results](https://rawgit.com/krausest/js-framework-benchmark/5d7a88e5929b02a9a4d7630de63d7d1953375d0b/webdriver-ts/table.html) (of a single run) for chrome 58.
Clear rows became faster indeed. Any suspicious values? (Datum might be worth running again due to the very high variance)

PR #167, opened 2017-04-22
domvm v2.1.4

leeoniya:
this update includes an internal bulk clearing optimization that avoids invoking the full dom reconciler; should do much better for "clear rows".

thanks!


krausest:
Results have been [updated](https://rawgit.com/krausest/js-framework-benchmark/1390cabc8d1e9b255a3311dda6b91ef16221fabc/webdriver-ts/table.html).

leeoniya:
cool, thank you.

PR #170, opened 2017-05-14
Use advanced mode in Closure Compiler for Angular

trotyl:
Closure compiler provides an `ADVANCED` mode which could perform more optimization, and Angular is compatible with that.


krausest:
Thanks - looks much [better](https://rawgit.com/krausest/js-framework-benchmark/02dc4c7a644d864cc99d1df6925c7f8d60cf696d/webdriver-ts/table.html) than the 4.0.0 results. Do you have an idea what improved results most? Closure or the version bump?

trotyl:
I think it's true that v4 being a little slower than v2 (and that's by design for the benefits of generated code size), but v4 has got better compatibility with Closure compiler, so that we could use advanced mode without hacking the source, which results in the better optimisation.

I'd like to categorize it as "Better tooling support".

PR #173, opened 2017-05-15
Adding slim.js

eavichay:


leeoniya:
this is a non-keyed implementation.

below is what i get vs vanillajs non-keyed. non-keyed React is currently at 1.40.

slimjs

krausest:
Thanks for your PR. Results are [here](https://rawgit.com/krausest/js-framework-benchmark/dee6ab45cff9e1f940865c78cb4a9a3898a931d2/webdriver-ts/table.html). 
Seems not really fast enough for a react killer.

eavichay:
It's a slow killer :)

PR #174, opened 2017-05-15
surplus: smaller package via tree-shaking, leaner S

adamhaile:
Hi Stefan --
A couple changes:

- I tweaked the S and surplus packages to also export es6 modules, which allows the use of rollup's tree-shaking.  Package went from 25k to 13.6k.

- I've been profiling S apps and found that the vast majority of nodes in the dependency graph have only a single edge, so S now has some memory and perf optimizations for this case.  Here, I get S keyed at about 1.028, non-keyed about 1.02.  Curious to see how it goes on your box.

Again, thanks!
-- Adam


leeoniya:
probability is high that using the Java version of Google Closure Compiler (CLI) would compress better than Uglify. As a bonus it may also get faster. If you're feeling really adventurous, you should probably use the `ADVANCED_OPTIMIZATIONS` flag for max squish.

adamhaile:
@leeoniya Thanks, closure's been on my list to check out. So the Java version is more effective than the js one?  Didn't know that.  Useful!

As a quick check, I threw the surplus package into their online service: https://closure-compiler.appspot.com/home .  It did indeed squeeze about 600 more bytes out.  However, when I benched it, I got inconsistent performance results.  A few runs looked really fast, but a few looked like it had gone backwards about 5%.  Huh ... have to experiment more.

leeoniya:
the js version is a pale shade. see https://github.com/Rich-Harris/butternut/issues/35

krausest:
Results are [here](https://rawgit.com/krausest/js-framework-benchmark/ef4824938272739e66578ca9376a0432d8e46110/webdriver-ts/table.html).

PR #180, opened 2017-05-17
pico-dom-v0.33.0

hville:
As commented, version 0.30.0 was not compatible with the previous v0.18.0 submission.
Here is an update.

It is still an ongoing experiment to explore the possibilities of a tiny lib (2kb) with minimal API and memory footprint.


krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/b143ef4bccc4ca6dfdb8e33155a5e1dfc1964618/webdriver-ts/table.html)
Create rows seems to be a bit slower than before - is that expected?
I tweaked menu.js to wrap each button in a col-sm-6 div and surely did it not the most efficient way - maybe you want to optimize that in a future version...

hville:
Yes. The slowdown is expected. The older version was more imperative an a
bit of a foot-gun I thought. I would remove it from the table.

The newer version creates intermediate immutable Template that are easier
to import and reconfigure.

Thanks alot for your time, feel free to drop them when your table gets too
large as they are both not production ready.

PR #181, opened 2017-05-17
add Svelte 1.20.2, keyed and non-keyed

Rich-Harris:
We have a completely new approach to keyed updates now, so it's a good time to update the Svelte implementation. Thanks!

### DIFF
--- a/svelte-v1.20.2-keyed/index.html
+++ b/svelte-v1.20.2-keyed/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Svelte v1.8.1</title>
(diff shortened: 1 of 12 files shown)


krausest:
Thanks. Isn't it funny that commit #666 is for a non-framework framework?
Results are [here](https://rawgit.com/krausest/js-framework-benchmark/67a05320ee3d485f059c3cb11fa2b872ded3c5bf/webdriver-ts/table.html)
Looks pretty good except for swap rows in the keyed implementation. Seems like all rows are updated.

Rich-Harris:
Thanks! I've opened an issue so that we can look into that — https://github.com/sveltejs/svelte/issues/588.

PR #183, opened 2017-05-19
surplus: split keyed/nonkeyed implementations

adamhaile:
Hi Stefan --
My last PR shrunk surplus' package size, but startup times got worse, which was counter-intuitive.

Looking into it, I think startup may have been getting hung up on a bit of code that decided whether to use the keyed or non-keyed variants.  I was sniffing window.location.search to detect a flag on the url, then loading either the keyed or non-keyed code depending.

So this version splits the two implementations, thereby avoiding the need to decide which to run at startup.

In my local testing, this puts surplus' startup on par with vanillajs'.  Surplus tends to do a little worse on your machine -- not surprising, since I've been tuning on mine -- so I'm curious to see where this comes out for you.

If you're deep into work for the next round, feel free to ignore until a convenient time.

Best
-- Adam


ismail-codar:
@adamhaile s.js looks cool. I am testing surplus-v0.4.0-keyed/index.html manually and observing running times in console.

Scenario 1 is normal
1-Create 10,000 rows - runLots took 1504.72
2-Append 1,000 rows - add took 383.655

But Scenario 2 is abnormal
1-Create 1,000 rows - run took 159.92
2-Create 10,000 rows - runLots took 1643.2649
3-Append 1,000 rows - add took 6169.89 !!!!!!!!!!!!!!!!!!!!!!!!!!

Why scenario 2 - step 3 took very long time.Do you have an idea.

adamhaile:
@ismail-codar Woah!  Yeah, I see the same behavior under those scenarios.  I made some changes to the s-array code at the core of this and I must have broken something.  I'll take a look tonight.  Thanks for catching it, and thanks for checking out S.js!

adamhaile:
Found the issue.  I had added a "fast path" in the array-diffing code for cases when we're going to/from an empty array.  The case for going to an empty array had a bug in it whereby the same now-empty array ended up being used for both the source and mapped arrays.  This ended up causing a scenario whereby when we went to add 1,000 items to the 10,000 items, it looked to the code like we were *replacing* 10,000 items with 11,000 entirely new ones, causing a ton more work.

Scenario 1 didn't hit it, b/c it went empty -> full -> fuller.  Scenario 2 went empty -> full -> empty (bad state) -> full (bad state, but didn't hit performance issue) -> fuller (bad state caused perf issue).

I need to cook dinner for my family but will get a fix in to s-array tonight.  Thanks!

ismail-codar:
You are so fast 🥇

krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/074dade9f9802cec412c82a47e222c9e3579a83f/webdriver-ts-results/table.html).

PR #184, opened 2017-05-19
add RE:DOM

pakastin:
https://redom.js.org


krausest:
Great - except for remove row in keyed mode it performs well. Any chance to improve that? (without adding benchmark specific optimizations ;-)
Results are [here](https://rawgit.com/krausest/js-framework-benchmark/074dade9f9802cec412c82a47e222c9e3579a83f/webdriver-ts-results/table.html).

pakastin:
Yeah I noticed too, have to check that out 😎

PR #185, opened 2017-05-26
add react-easy-state

solkimicreb:
Add [React Easy State](https://github.com/solkimicreb/react-easy-state)

Easy State is a small package that aims to improve React's own state management instead of delegating it to external libraries - like Redux or MobX. To achieve this, it does two things:

- It replaces React's state and `setState` with plain native JavaScript.
- It autobinds component methods.

On my machine the benchmark scores at around Redux (slightly better). I think that the rollup build is responsible for the faster startup time. I would gladly refactor the other React examples to use ES6 syntax and rollup if the other authors agree (:


krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/074dade9f9802cec412c82a47e222c9e3579a83f/webdriver-ts-results/table.html)

Issue #172, opened 2017-05-14
Update dio


krausest:
@thysultan 
Dio seems to have changed significantly such that I see little chances for updating it correctly myself. Are you interested in updating dio to the latest version?

thysultan:
Yes, have a working(keyed) version of the new release here [https://rawgit.com/thysultan/dio.js/master/experiment/docs/examples/benchmark.html](https://rawgit.com/thysultan/dio.js/master/experiment/docs/examples/benchmark.html) Will send a PR in the coming days when i release.

krausest:
Thanks - has been merged!

PR #186, opened 2017-05-26
Add DIO v7.0.1

thysultan:
re: [issues/172](https://github.com/krausest/js-framework-benchmark/issues/172)


leeoniya:
here's what i'm getting.

your startup time isnt doing you any favors. the generated bundle is pretty hefty. also mem usage is in-line with what domvm's has been when it was generating the full list eagerly and creating new vnodes for it. i've been working to be able to do lazy recycling of the old vnodes and as you can see it has paid off pretty well in the upcoming domvm update.

js-bench

thysultan:
I don't think there's any minification going on there the mini+gzip bundle is 6kb [https://unpkg.com/dio.js@7.0.0/dio.min.js](https://unpkg.com/dio.js@7.0.0/dio.min.js)

leeoniya:
> I don't think there's any minification going on there the mini+gzip bundle is 6kb

most of the generated bundle isn't dio itself but an included "The buffer module from nodejs, for the browser". at least on my machine, though i dont know why it would matter.

> BTW i can't see you're screenshots, github is compressing them.

if you click on the image it opens the full png. it's just scaled down in the thread here.

thysultan:
> most of the generated bundle isn't dio itself but an included "The buffer module from nodejs, for the browser". at least on my machine, though i dont know why it would matter.

re: Startup time, more code to parse < startup time, the parts it's including should only be included in a server env, will look into it.

> if you click on the image it opens the full png. it's just scaled down in the thread here. 

The larger image is what i was talking about, it's compressed, you can't make out any [text](https://cloud.githubusercontent.com/assets/43234/26507208/eca3ca6a-4214-11e7-8501-c2b0432f0bc4.png).

leeoniya:
> re: Startup time, more code to parse < startup time

yeah, i was aware ;)

> The larger image is what i was talking about, it's compressed, you can't make out any text.

your browser is prolly just fitting it to your screen.  just zoom in :)

thysultan:
> most of the generated bundle isn't dio itself but an included "The buffer module from nodejs, for the browser"

It's probably this part, [https://github.com/thysultan/dio.js/blob/master/dio.js#L2515](https://github.com/thysultan/dio.js/blob/master/dio.js#L2515)

I thought i had managed to get webpack to not bundle that when i tested it before :(

thysultan:
Updated.

thysultan:
> i've been working on implementing *lazy recycling* of the old vnodes

@leeoniya I don't understand, can you elaborate?

krausest:
Thanks - results have been [updated](https://rawgit.com/krausest/js-framework-benchmark/074dade9f9802cec412c82a47e222c9e3579a83f/webdriver-ts-results/table.html).

leeoniya:
> @leeoniya I don't understand, can you elaborate?

@thysultan, see the updated impl [1]. the original discussion is here [2].

[1] https://github.com/krausest/js-framework-benchmark/blob/master/domvm-v3.0.1-keyed/src/main.es6.js#L92-L95
[2] https://github.com/leeoniya/domvm/issues/111

PR #188, opened 2017-05-28
pico-dom-v1.0.0

hville:
Finally streamlined and froze the API, published 1.0.0.
A little faster also it seams.


krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/074dade9f9802cec412c82a47e222c9e3579a83f/webdriver-ts-results/table.html)

PR #194, opened 2017-05-31
Improve row selection in Elm 0.18 implementation

evancz:
## Summary

This change simplifies the `update` logic for selecting items.

## Details

In the previous implementation, the `Select` logic included:

  1. A `List.map` that deselected all rows.
  2. A conversion from `List` to `Array`
  3. An update of that `Array`
  4. A conversion back from `Array` to `List`

This is quite a bizarre choice because everything *can* be done in that initial `List.map`. This PR changes the code slightly to skip steps 2, 3, and 4.

The benefits of this change include:

  - The `elm-community/array-extra` dependency can be removed.
  - Less garbage will be produced by skipping these `Array` conversions.
  - Less computation is necessary to select an item.
  - The resulting code is simpler.

## Expectations

This should improve speed and allocation, but I expect the improvement to be rather small.
 
Again, thank you for your consideration!

### DIFF
--- a/elm-v0.18.0/elm-package.json
+++ b/elm-v0.18.0/elm-package.json
@@ -8,7 +8,6 @@
     ],
     "exposed-modules": [],
     "dependencies": {
-        "elm-community/array-extra": "1.0.2 <= v < 2.0.0",
         "elm-lang/core": "5.0.0 <= v < 6.0.0",
         "elm-lang/html": "2.0.0 <= v < 3.0.0",
         "mgold/elm-random-pcg": "4.0.2 <= v < 5.0.0"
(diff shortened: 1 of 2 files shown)


krausest:
Results have been [updated](https://rawgit.com/krausest/js-framework-benchmark/50dc1cc8d3a6b0ab2385cba013828a0e511936c4/webdriver-ts-results/table.html). Unfortunately the numbers don't show that improvement. Old values are [here](https://rawgit.com/krausest/js-framework-benchmark/f37bbd413c435d3fff3f2fcd5207e1cba1523507/webdriver-ts-results/table.html)  for reference. I'm afraid - sometimes variance battles small improvements.

evancz:
That's alright. The code is nicer at least!

PR #195, opened 2017-06-05
Forgo positional indexing in Elm

manlon:
This is a small simplification that follows up on 794f3ebd2835dcb742212aea0024925099b7b15d and takes better advantage of that optimization. It yields a modest speedup in the "remove" benchmark.

### DIFF
--- a/elm-v0.18.0/src/Main.elm
+++ b/elm-v0.18.0/src/Main.elm
@@ -115,13 +115,13 @@ btnPrimaryBlock ( buttonId, labelText, msg ) =
         ]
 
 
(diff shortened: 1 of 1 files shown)


krausest:
Sorry for the delay + thanks. Usually I'm publishing here updated results for that change. I'm not doing it here because I assume the chrome update with the big v8 change would render that useless. I see if I can update all results the next days.

evancz:
Nice job @manlon! I was wondering how to get rid of that `take ++ drop` code when I looked. Glad to see you found a way that should also make things more reliable! :)

Issue #196, opened 2017-06-18
A review for compare with framework (t-test) would be good


krausest:
The [new](https://rawgit.com/krausest/js-framework-benchmark/master/webdriver-ts-results/table.html) result table has a compare frameworks feature. 
Would be nice if someone reviewed the [code](https://github.com/krausest/js-framework-benchmark/blob/2abf9e409f15f6f1fff980a98beb9912e608abd0/webdriver-ts-results/src/Common.tsx#!L216-L240) for the statistical test.

Charles6666:
Yes this simple need to be off look like this☡

brodybits:
> Would be nice if someone reviewed the [code](https://github.com/krausest/js-framework-benchmark/blob/2abf9e409f15f6f1fff980a98beb9912e608abd0/webdriver-ts-results/src/Common.tsx#!L216-L240) for the statistical test.

From a quick look I wonder if it would be practical to apply some judicious refactoring?

I wonder if there may be existing packages that can be used to remove a bunch of code from needed maintenance in general?

I wonder if it may be possible to identify some pieces of code that could be turned into reusable packages for the benefit of others?

As I said in #547 I think proper performance benchmarking is not so easy in general (hard to do right). Kudos to @krausest and others for the dedication and hard work on this project!

PR #197, opened 2017-06-21
Update RE:DOM to v3.0.2

pakastin:
"Remove row" test should now be ~3x faster!


krausest:
Sorry - didn't have to time to merge your pull request yet. I'll try to find some time the next few days.

pakastin:
No worries, I just resolved that new conflict for you to make it easier to merge 😉

pakastin:
Should be ready to go without any hassle. At least I ran the tests and it worked allright – much faster than the previous version! 😎

krausest:
Finally...
Results are [here](https://rawgit.com/krausest/js-framework-benchmark/964ae451d222ccc4de774fedeef7025c10227558/webdriver-ts-results/table.html)
Remove rows is much, much faster!
What's interesting is that the compare view shows that startup time improved significantly, but it seems like clear rows is a bit - but still significantly - slower (or I really need a review for my t-test #196  :smirk:) 
bildschirmfoto 2017-06-30 um 21 39 45

pakastin:
Yeah, it's a small tradeoff. It's now doing the remove with lookup table instead of looping the DOM. It looks bad in your table (red), but if you compare to other frameworks it's not *that* bad (TOP 10!) 😀👍

Now: https://github.com/pakastin/redom/blob/master/src/list.js#L60-L68

Before: https://github.com/pakastin/redom/blob/master/src/setchildren.js#L29-L35

pakastin:
Oh, and TOP 5 in memory allocation! I'm really happy with the results! 👌

Instead of micro-optimizations and clever tricks, I've wanted to keep the source clean and readable + let the browser do most of the work natively. It's proven that the DOM isn't slow after all, and that you can build a fast UI library with only 2 KB 😊

For example here's all the code for reordering child elements: https://github.com/pakastin/redom/blob/master/src/setchildren.js#L4-L36

Thanks for the merge, I've waited this anxiously! ✌️😎

krausest:
Sorry I think you got me wrong - I guess I didn't explain that screenshot good enough. 
RE:DOM's performance is definitely really nice (and quite "green")
bildschirmfoto 2017-07-01 um 09 33 35
The screenshot I pasted is just a statistical comparison. Green values mean that a t-test shows that the "green framework" is statistically significant faster, and red means it's slower. The color displays the confidence level. A light green or red means the p-value is 10% and becomes darker when the p-value is smaller. But the color is absolutely no indications how much slower or faster the other is. 
Still one can e.g. say (as long as i did the t-test right) that re:dom is significantly faster than react for create rows (which sounds very sophisticated).
bildschirmfoto 2017-07-01 um 09 41 31

pakastin:
I did understand. I just meant that the optimization to "remove one row" affected the "clear rows" slightly. It's always balancing between optimizations 🙂

Btw. it would be nice if "slowdown geometric mean" would be the default order instead of alphabetical.

Issue #182, opened 2017-05-17
Glimmer build fails (only with yarn)


krausest:
Glimmer doesn't building with the following message when packages are installed with `yarn install`, installing with `npm install` works. I changed package.json to have fixed version numbers to facilitate reproducible builds.
@rwjblue Any help would be very appreciated how to fix it.

Stefans-MBP:glimmer-v0.3.10 stefan$ npm run build-prod

> glimmer-app@0.0.0 build-prod /Users/stefan/Source/Javascript/js-framework-benchmark/glimmer-v0.3.10
> ember build --environment=production

Could not start watchman
Visit https://ember-cli.com/user-guide/#watchman for more info.
⠋ BuildingClass constructor AbstractBuild cannot be invoked without 'new'
TypeError: Class constructor AbstractBuild cannot be invoked without 'new'
    at new GlimmerApp (/Users/stefan/Source/Javascript/js-framework-benchmark/glimmer-v0.3.10/node_modules/@glimmer/application-pipeline/dist/lib/broccoli/glimmer-app.js:111:24)
    at module.exports (/Users/stefan/Source/Javascript/js-framework-benchmark/glimmer-v0.3.10/ember-cli-build.js:6:13)
    at Builder.setupBroccoliBuilder (/Users/stefan/Source/Javascript/js-framework-benchmark/glimmer-v0.3.10/node_modules/ember-cli/lib/models/builder.js:56:19)
    at new Builder (/Users/stefan/Source/Javascript/js-framework-benchmark/glimmer-v0.3.10/node_modules/ember-cli/lib/models/builder.js:30:10)
    at BuildTask.run (/Users/stefan/Source/Javascript/js-framework-benchmark/glimmer-v0.3.10/node_modules/ember-cli/lib/tasks/build.js:15:19)
    at Promise.resolve.then (/Users/stefan/Source/Javascript/js-framework-benchmark/glimmer-v0.3.10/node_modules/ember-cli/lib/models/command.js:243:46)
    at tryCatch (/Users/stefan/Source/Javascript/js-framework-benchmark/glimmer-v0.3.10/node_modules/rsvp/dist/rsvp.js:539:12)
    at invokeCallback (/Users/stefan/Source/Javascript/js-framework-benchmark/glimmer-v0.3.10/node_modules/rsvp/dist/rsvp.js:554:13)
[...]

PR #201, opened 2017-07-02
Use yarn and node 7 in travis

alexteves:


alexteves:
[Successfully installs and builds](https://travis-ci.org/alexteves/js-framework-benchmark/builds/249224917).
What does ["yarn destroys aurelia"](https://github.com/krausest/js-framework-benchmark/commit/d47f1f02c69d802e6ba0c4dcde286c378a370c0a) mean?

krausest:
Building a single js project can already be demanding - doing so for 40+ frameworks can drives me insane at times. I've run into several occasions where building with npm worked fine, but failed for yarn. I can't remember exactly what it was for aurelia, but at least for glimmer #182 was documented properly.

Automating the build with travis-ci sounds very nice. Last time I tried I failed, npm took too long and the build was cancelled. I'll try if I (too) can manage to build in in travis.

krausest:
Great! First I got a "“fatal: reference is not a tree” error" since (at least as I understood) travis tried to checkout a pretty old build that failed some months ago when I tried travis. I increased the git-depth and finally it worked - that's really great. I'm excited if I can catch build errors that way earlier.

PR #202, opened 2017-07-02
Also run travis on OS X

alexteves:
Have another branch adding per-job status badge. Will open PR soon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+os:
+  - linux
+  - osx
 language : node_js
 node_js  : 7
 cache    : yarn


krausest:
Thanks a lot!

Maybe you can help me understanding travis better. When I tried travis-ci again it failed a few times with the following message.
```
$ cd krausest/js-framework-benchmark
$ git checkout -qf ea9238612f65079c515c107ae4a0a6ae1205534b
fatal: reference is not a tree: ea9238612f65079c515c107ae4a0a6ae1205534b
The command "git checkout -qf ea9238612f65079c515c107ae4a0a6ae1205534b" failed and exited with 128 during .
Your build has been stopped.
```
I resolved it with the git depth param.  Was there a better way to fix that? Why did travis try to checkout a very old build that failed instead of the master?

alexteves:
No idea. Never had to mess around with --depth. Maybe there's some leftover setting from the first time you tried travis? These are my settings; I believe they're all default.

image

It seems that "git push -f" can also trigger this error.
https://github.com/travis-ci/travis-ci/issues/617

Issue #166, opened 2017-04-22
Faster clear by running `.textContent = ""` after a rAF


adamhaile:
Discovered this one from vidom's fast "clear rows" result and thought I'd pass it on.

If you run .textContent = "" after a requestAnimantionFrame, it saves 35-40ms in the "clear rows" benchmark.

So tweaked vanillajs-keyed Main.clear() looks like:
```javascript
    clear() {
        startMeasure("clear");
        this.store.clear();
        this.rows = [];
        this.data = [];
        var that = this;
        requestAnimationFrame(function () {
            that.removeAllRows();
            this.unselect();
        });
        stopMeasure();
    }
```

Same trick saves about 5ms in the "replace all rows" test.

krausest:
Thanks - very interesting. I implemented it only for clear all rows where it improved the benchmark significantly.
As for replace all rows I decided not to implement it, since I've seen the 5 mecs improvement you described but about half of the time it performed worse (maybe a case of bad timing when it almost had to wait almost 17 msecs for the next frame).

krausest:
Updates results [here](https://rawgit.com/krausest/js-framework-benchmark/ee0a6108d1632580bbb161179796487c4991d8da/webdriver-ts/table.html)

PR #203, opened 2017-07-03
domvm-v3.0.1

leeoniya:
hey @krausest,

this is both a lib & implementation update which takes advantage of a new `lazyList` feature in v3 that reuses vnodes from the old vtree when a cheap diff indicates no changes.

it may be a bit faster overall, but the major win here is a ~20% mem reduction.

thanks!


krausest:
Merged! Results are [here](https://rawgit.com/krausest/js-framework-benchmark/2d96cef50df08a2048e89c6df750476842cd4fb0/webdriver-ts-results/table.html).
Any idea why travis couldn't checkout domvm in the pull request correctly? Let's see if it works on master...

leeoniya:
thanks!

i removed the console timers (and the forced-sync redraws which were there to facilitate them) since i no longer use them and they bloat the impl. this change leaves the impl with domvm's native async api which has the improved `clear rows` effect already observed here [1]. i just want to note this here because it's not a micro-optimization since that's just how the lib has always worked and if anyone's lib works this way out of the box, then that's perfectly fine, too. the main purpose of the update was to reduce mem use, whhich definitely stood out among libs in the same perf range :)

i dunno what Travis is up to; he sounds like a real hooligan.

[1] https://github.com/krausest/js-framework-benchmark/issues/166

PR #206, opened 2017-07-06
introduce optimized angular without zone.js

trotyl:
Optimized Angular implementation that does not use Zone.js, manually triggers all change detection like React does. Since it may not be a recommended way, made it a separate item.

Result going from 1.34 down to 1.26 in my local machine.

Should align with the official no-zone usage in https://github.com/angular/angular/pull/17672 after it lands.

### DIFF
--- a/angular-v4.1.2-no-zone-keyed/.gitignore
+++ b/angular-v4.1.2-no-zone-keyed/.gitignore
@@ -1,0 +1,6 @@
+dist/
+lib/
+node_modules/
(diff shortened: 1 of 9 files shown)


krausest:
Thanks! Funny - got [exactly](https://www.rawgit.com/krausest/js-framework-benchmark/a13a51f40c373a036ce2aca355599f36bce54cf4/webdriver-ts-results/table.html) the same number.

PR #212, opened 2017-07-14
Add Moon

KingPixil:


leeoniya:
would be cool to have `index.html` nicely formatted, since that's a showcase of Moon's API :)

KingPixil:
Yeah, I'll format the HTML and add the directives. You are also correct about the render function, that was compiled and optimized by the Moon compiler from that template. I'm going to keep that render function as it is recommended to precompile templates to avoid having Moon compile everything again.

leeoniya:
> I'm going to keep that render function as it is recommended to precompile templates to avoid having Moon compile everything again.

of course, that's fine. it just helps to have the impl source here as users would be expected to author it.

btw, is this a `keyed` or `non-keyed` impl? (this repo supports both simultaneously).

KingPixil:
Yeah, this is a `non-keyed` implementation, Moon doesn't support a keyed virtual DOM yet and there doesn't seem to be a need for it (at least for now).

krausest:
Thanks. Results are [here](https://rawgit.com/krausest/js-framework-benchmark/168984a93952260952c30aa9b02bb044a5ae4578/webdriver-ts-results/table.html).

KingPixil:
Wow, wasn't expecting to see the results already. Thanks!

PR #214, opened 2017-07-16
Add benchmark for reflex-dom-v0.4

alexteves:


krausest:
Hi Alexandre,

I'm afraid I need to make a hard decision even though I very much appreciate your travis-ci efforts: I am not going to merge reflex-dom. To create some results I had to increase timeouts beyond meaningful values and partially still had to reduce count since performance seems to decrease over runs. The results show the hopelessly bad values:
bildschirmfoto 2017-07-17 um 20 44 19
Combined with a build process that uses (and maybe also downloads) 8,5 GB and needs root access I see no chance anyone might actually use that framework. Sorry!

Yours,
Stefan

Issue #213, opened 2017-07-15
Surplus implementation doesn't have aria attributes


localvoid:
@adamhaile 

React and probably all other implementations:
 - [Main](https://github.com/krausest/js-framework-benchmark/blob/8b9f6613cf3b7e94f7bcef1a784a2db49211b368/react-v15.5.4-keyed/src/Main.jsx#L128)
 - [Row](https://github.com/krausest/js-framework-benchmark/blob/8b9f6613cf3b7e94f7bcef1a784a2db49211b368/react-v15.5.4-keyed/src/Row.jsx#L39)

Surplus:
 - [Main](https://github.com/krausest/js-framework-benchmark/blob/8b9f6613cf3b7e94f7bcef1a784a2db49211b368/surplus-v0.4.0-keyed/src/view.tsx#L43)
 - [Row](https://github.com/krausest/js-framework-benchmark/blob/8b9f6613cf3b7e94f7bcef1a784a2db49211b368/surplus-v0.4.0-keyed/src/view.tsx#L57)

leeoniya:
yep, it's required [1]. ping @adamhaile.

[1] https://github.com/krausest/js-framework-benchmark/issues/9#issuecomment-232443783

adamhaile:
Oops, yep, that's a goof.  I don't remember removing the attribute, but I can believe it.  The underlying issue is that surplus' currently-released compiler only supports a subset of JSX, specifically the part that overlaps with surplus' earlier template syntax.  Attributes that aren't available as properties, like aria-*, aren't covered.  I did the initial implementation of the benchmark one evening months ago.  It was actually the first thing I wrote using JSX.  I'll bet I ditched aria-hidden then.  I should have fixed that before submitting but honestly didn't remember.  Thanks for catching it @localvoid.

I'm currently rewriting the surplus compiler to support more of JSX (spreads, the style object, attributes, etc).  I was planning to resubmit when that was done.  I can roll the aria-hidden fix into that or push a patch, whatever @krausest  prefers.

adamhaile:
PR #215

PR #215, opened 2017-07-16
add missing aria-hidden="true" attributes to surplus

adamhaile:
@localvoid noticed (#213) that I mistakenly didn't include the aria-hidden="true" attributes in the surplus implementation.  He's right, they should be there, so this adds them.  It does it in a somewhat artless way.  Once I finish my rewrite of the surplus compiler I'll repost a cleaner fix.

### DIFF
--- a/surplus-v0.4.0-keyed/src/view.tsx
+++ b/surplus-v0.4.0-keyed/src/view.tsx
@@ -40,7 +40,7 @@ export let AppView = (app : App) =>
                 onClick = {(e : any) => e.target.matches('.delete') ? app.delete(rowId(e)) : app.select(rowId(e))}>
                 { TBody(app) }
             </table>
-            <span className="preloadicon glyphicon glyphicon-remove"></span>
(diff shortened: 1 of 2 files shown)


krausest:
[New and updated](https://rawgit.com/krausest/js-framework-benchmark/8814a21b80d753c09995bf20b568dfdbe12bed22/webdriver-ts-results/table.html) results are as fast as [before](https://rawgit.com/krausest/js-framework-benchmark/168984a93952260952c30aa9b02bb044a5ae4578/webdriver-ts-results/table.html)!

adamhaile:
Yeah, that's about what I was expecting.  I did a couple local tests, and any difference seemed to be too small to separate from the noise.  Other than the startup tests, the create 10k test for non-keyed looks like it slowed a bit (1399 -> 1454ms).  Maybe that's the cost of 10k setAttribute() calls, or maybe just random variation, dunno.  

Svelte passed me in the non-keyed rankings!  The rats.  I'm kidding of course.  We both compile to some fairly similar DOM ops, so once they clean up a couple of their bottlenecks I suspect we'll be essentially tied.

Anyway, I'm glad to have it conformant, and thanks for merging and re-running the results!

Issue #216, opened 2017-07-19
Simulacra 2.1.5 fails to remove row


krausest:
Simulacra had a ^-version in package.json and resolved to 2.1.5 instead of 2.1.1, which doesn't respond  to remove row.
Circleci failed because of that benchmark 06_remove-one-1k.

krausest:
@daliwali Can you take a look at it?

daliwali:
Thanks for the notice -- I will have a look at it. In the tests and examples removing nodes works fine, so this seems very odd.

PR #217, opened 2017-07-20
Fix brittle benchmark implementation in Simulacra.js

daliwali:
I had previously implemented the remove benchmark using a positional index in the DOM, which was evidently flaky and unreliable. This fixes that by using a CSS selector.

Closes https://github.com/krausest/js-framework-benchmark/issues/216


krausest:

PR #219, opened 2017-07-24
Add Dojo 2

agubler:
Please find an implementation for Dojo 2, the results have highlighted some areas that we'd like to work on so hopefully we'll be able to reduce the numbers over the next few beta releases.

Please let me know if there is anything else needed.


krausest:

PR #220, opened 2017-07-25
domvm v3.0.5

leeoniya:
might see some improvement in append1k and a bit of a mem reduction, hopefully no regressions otherwise.


krausest:

PR #222, opened 2017-07-28
Format index.html somewhat

alexteves:


alexteves:
Hadn't noticed index.html is auto generated; disregard.
https://github.com/krausest/js-framework-benchmark#optional-62-updating-the-indexhtml-file

PR #225, opened 2017-08-04
update angular version

trotyl:
Update to latest stable angular version.


krausest:
Angular is giving me headaches as it crashes my memory benchmarks.  It causes an error `WebDriverError: unknown error: Cannot read property 'length' of undefined` when `window.gc` is invoked. It worked for halogen and marionette in the same run. 
Any idea how angular might cause such a problem?

krausest:
When I run chrome with `--js-flags=--expose-gc` and  run the angular app and execute `window.gc()` in the console I get the error
```
bundle.js:52 Uncaught TypeError: Cannot read property 'length' of undefined
    at gc (bundle.js:52)
    at <anonymous>:1:8
```

trotyl:
I'm not quite sure how client-side script could break driver, seems to be broken from 4.2.0 in https://rawgit.com/trotyl/js-framework-benchmark/angular-versions/webdriver-ts/table.html (forgive me for not checking the memory result), I'll investigate it tomorrow.

krausest:
I believe angular is patching window.gc:
This is react:
bildschirmfoto 2017-08-06 um 21 31 55
And this is angular:
bildschirmfoto 2017-08-06 um 21 31 35
(Chrome started with `--js-flags=--expose-gc`)

Webdriver is just reporting this error.

trotyl:
I have had some debugging, this is due to the Closure compiler renaming process, it assign name `gc` to some irrelevant function and not being properly scoped. I'll provide a fix for that soon.

image

trotyl:
I have created a new [PR here](https://github.com/krausest/js-framework-benchmark/pull/229), could you please take a look if it's being properly fixed?

PR #226, opened 2017-08-05
Update marionette benchmark

paulfalgout:
In December the Marionette team, on the library's 5th year birthday [made a public commitment](https://medium.com/blog-marionettejs/heres-to-the-next-5-years-of-marionettejs-12cd45c1da08) to iteratively bring the library to a better competitive place through incremental updates that don't require rewriting an app from the ground up in the flavor-of-the-week framework.

In February we added Marionette to this framework benchmark understanding the state we were in.  At the time we added the framework with no optimizations, no customizations, and every feature possible.  This benchmark helped us get a clear understanding of places to focus.  At the time the associated blog suggested that Marionette's performance might justify a migration.

Now I'm presenting one updated benchmark for the most recent version of Marionette. In addition to the library improvements, it also contains some common optimizations that real users of the library are utilizing such as precompiling templates and moving event listeners to a parent view of a very large collection. This is probably a general representation of a common setup. 

I am also including a second benchmark that shows where the Marionette community is heading.  With 2 very simple APIs introduced in v3.3 and v3.4 users can remove jQuery for DOM interaction and utilize a better renderer with ease to take their 5 year old project and make it competitive with common libraries.  In this case I went with morphdom, but there are [many other options](https://medium.com/blog-marionettejs/going-virtual-or-incremental-with-marionette-js-1a38837cc1be).

We're definitely looking forward to an updated results table and the next blog post :-)  Excellent benchmark!  Thanks!


krausest:
The difference between 3.3.1 and 3.4.0 is impressive and brings the average slowdown significantly below factor 2 - nice! And the domapi version is even faster!

PR #227, opened 2017-08-05
Add PureScript's Halogen v2.1.0

saylu:
Hi @krausest!

I've provided a benchmark of the Halogen framework from PureScript ([repo here](https://github.com/slamdata/purescript-halogen)).

This is a PureScript framework written 100% in PureScript including the virtual DOM. I believe it is the most widely-used framework in production.


saylu:
As a note -- I'm not sure if this automatically kicks off a Circle CI build, but I ran the tests on my fork and they've passed:

circle ci

This ought to be good to go as far as that's concerned.

krausest:
To be honest - not really impressive. Does halogen support keyed updates?

The disk size is 146MB for halogen which is compared to the 177MB for ember quite ok.

Update: After running the benchmark I noticed that halogen doesn't preload the remove icon. I changed that in commit https://github.com/krausest/js-framework-benchmark/commit/be8bb0f5746f10b2904072ff898b8e2f2a6a013a . I'll update the results soon, but I don't expect that it changes a lot.

saylu:
I believe it's possible to do a keyed variant also. I've been stretched for time lately but can work on a keyed implementation to see if that has different results.

There are other PureScript frameworks that rely on React (Thermite is basically a wrapper for React) and should be fast for that reason. However, Halogen is 100% PureScript, which both has the benefit of being a joy to develop in and the downside of less investment into performance optimization.

Thanks for the merge!

krausest:
@thomashoneyman I just noticed a bug: If you click "create 1,000 rows" and then "swap rows" the labels, but not the row ids, are swapped.

Can you please take a look at that?

PR #228, opened 2017-08-06
Migrate to CircleCI 2.0

alexteves:
The new configuration format gives us more flexibility. I believe with CircleCI 2.0 we can get much faster feedback cycles by:
- having decent control over caching
- parallelising 

As far as I can tell, the build remains the same.  However, just using 2.0 seems to speedup the build - did two runs both under 1h10m. I suspect it's running with more resources. Hopefully it stays that way.


krausest:
Funny - I had just skimmed the docs and had the impression that circleci 2 always needed a docker image  and supporting multiple languages wan't exactly easy (https://discuss.circleci.com/t/recommended-way-to-use-multiple-languages-in-2-0/14174/3)

krausest:
@alexfmpe I think the build is currently failing because circleci isn't using the node version specified in the circleci config file (node v6.1.0 is reported in the log, but 8.2 is requested to allow native async/await). Do you have an easy fix for it?

alexfmpe:
Check #259.

> Funny - I had just skimmed the docs and had the impression that circleci 2 always needed a docker image and supporting multiple languages wan't exactly easy

You were right. A non-docker build won't let you specify node version - it needs to be manually installed.

PR #229, opened 2017-08-07
fix memory result for angular

trotyl:
wrap all code in an IIFE to avoid naming conflict of global variables.


krausest:
Thanks!

PR #234, opened 2017-08-15
add etch-v0.12.5

lloiser:
Add the https://github.com/atom/etch library.

### DIFF
--- a/etch-v0.12.5-keyed/.babelrc
+++ b/etch-v0.12.5-keyed/.babelrc
@@ -1,0 +1,4 @@
+{
+    presets: [ "es2015", "react"],
+    plugins: []
+}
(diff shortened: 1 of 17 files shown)


krausest:
Thanks - not bad! Results are [here](https://rawgit.com/krausest/js-framework-benchmark/2459e72e7adce2d34ec65738d8aff04c03a68cbe/webdriver-ts-results/table.html).

krausest:
Looks like react in 7kb but with explicit update calls and only props - is that too simple?
Why is it so unknown?

lloiser:
Thx for merging.
Yes that sums it up pretty good. I don't know why it's unknown... It is already in use in the atom editor and by various plugins. Checkout there blog post about it http://blog.atom.io/2017/06/22/a-new-approach-to-text-rendering.html

PR #236, opened 2017-08-20
Update marionette to v3.4.2

paulfalgout:
This is a much smaller update compared to https://github.com/krausest/js-framework-benchmark/pull/226

One bug that was worked around from 3.4.0 was fixed and there was one small suggestion we added to the domapi test that makes it a bit more performant across the board.

Our next update will probably be v4 a few months from now.  I expect the library will get much smaller and a little faster.


krausest:
Thanks for the update. Results are [here](https://rawgit.com/krausest/js-framework-benchmark/2459e72e7adce2d34ec65738d8aff04c03a68cbe/webdriver-ts-results/table.html).

PR #237, opened 2017-08-24
item id is bound one-time in Aurelia

paolocarrasco:
An optimization is added to the Aurelia JS implementation (one-time binding behavior) for the ID item rendering.


krausest:
Thanks. Updated results are [here](https://rawgit.com/krausest/js-framework-benchmark/2459e72e7adce2d34ec65738d8aff04c03a68cbe/webdriver-ts-results/table.html) - sorry I'm afraid to say the one time binding did not improve the result significantly. BTW do you know whether aurelia supports keyed updates?

paolocarrasco:
Yes. It is a micro optimization but it is the right way in Aurelia.
I will research about that kind of update in Aurelia. If I found it I'll create a pull request. 
Thanks!

PR #242, opened 2017-08-30
Add PureScript's Thermite & Pux

saylu:
Hi Stefan,

I've added benchmarks for the other 2 widely-used frameworks in PureScript to round out the complete list (Thermite, Halogen, and Pux).

[Thermite](https://github.com/paf31/purescript-thermite) is a high-level wrapper around React. It's written by the creator of PureScript. In the benchmarks, it sees roughly the same slowdown as Halogen does. [Pux](https://github.com/alexmingoia/purescript-pux) is an Elm-inspired framework that also relies on React. It's got a very nice monadic DSL for writing HTML, but that convenience has the problem of being essentially a linked list, making huge appending operations very inefficient. It sees a significant slowdown because of this, ultimately clocking in at ~12x slower than vanilla JS.

While I'd love to see all three of these frameworks included for the general knowledge of people considering PureScript, JavaScript (especially functional-first), and Elm, I understand if you have reservations about slowdowns here. I've spoken with the authors of each of these frameworks and there doesn't appear to be an easy way to dramatically speed them up.

The build is passing on Circle CI and I've manually verified I can run and update benchmarks:
https://circleci.com/gh/saylu/js-framework-benchmark/6

Cheers,
Thomas


krausest:
I‘m on holidays this and next week - I‘ll take a look at it in two weeks.

saylu:
No worries, I understand. Have a good time on holiday!

krausest:
Thanks. Sadly you have to scroll far to the right to see the [results](https://rawgit.com/krausest/js-framework-benchmark/b6228535b2b54de7bfe02483411e8aacaccf2a1c/webdriver-ts-results/table.html).
Do any of the PureScript libraries support keyed updates? Personally I'm absolutely not fond of frameworks that support only non-keyed mode...

krausest:
@thomashoneyman They have the same bug as halogen: If you click "create 1,000 rows" and then "swap rows" the labels, but not the row ids, are swapped.

Can you please take a look at that?

Issue #243, opened 2017-09-02
What browser versions should be supported?


mixonic:
Howdy! Similar to https://github.com/krausest/js-framework-benchmark/issues/230, I want to try and make sure the additions I make here are as equitable as possible.

With that in mind, what kind of browser support should we be compiling for? Just latest Chrome, or IE9+, etc?

krausest:
Good question! I’m not sure if the results can be generalized to other browsers, currently only chrome is in the focus (the other would be firefox but I haven‘t found a way to extract performance data from it). Thus I’d say we require chrome only. If we decide otherwise I‘d say a test to verify that a benchmark implementation runs on all target browsers would be needed.
Other opinions?

gheoan:
Some framework focus on Chrome performance, so it would be interesting to see results from other browsers.

trotyl:
I guess the original question is about what browser should the application support:

> With that in mind, what kind of browser support should we be compiling for? Just latest Chrome, or IE9+, etc?

Instead of what browser should be tested for benchmark. So all the apps could have a generalized compiling target (`es2017`/`es2016`/`es2015`/`es5`).

Currently some of the testing apps are using `babel-presets-env` resulting in a potentially different compilation level:

```json
"presets": [
  [
    "env",
    {
      "targets": {
        "browsers": [
          "last 2 versions",
          "not ie <= 10"
        ]
      },
      "loose": true,
      "modules": false
    }
  ]
]
```

I think it's better to require all apps compiling to valid `es5` (including `es5.1`) so the browser optimization could be at same level, this should be easy to achieve, just run an extra e2e on PhantomJS and make sure it's running successfully.

For the performance result on different browsers, it's also an interesting topic, but could be decided later.

krausest:
@trotyl I'd allow them to support the es feature level of the latest stable chrome version. If for some reason transpiling to es5 yields better results for a benchmark implementation or it's just the way the framework is supposed to be used it should also be allowed to choose an older target. Let's just not hamper using modern es versions.

trotyl:
@krausest Current output levels are:

Framework|Language Level|Transpiler
---|---|---
AngularJS|`es5`|By Babel
Angular|`es5`|By Closure
Aurelia|not fixed but possibily `es5`|By Babel
Binding.scala|`es5`|By Scala.js
Bobril|`es5`|By TypeScript
Choo|`es5`|By Browserify plugin es2020
CycleJS|`es5`|By Babel
[...]

PR #244, opened 2017-09-02
Glimmer 0.7.2

mixonic:
* Tweak Glimmer 0.3.10 to use a single helper instead of two.
* Add Glimmer 0.7.2

screen shot 2017-09-01 at 8 38 20 pm


glimmer-v0.7.2-keyed/config/targets.js
```diff
@@ -0,0 +1,12 @@
+let browsers = [
+  'last 1 Chrome versions'
```
mixonic: Re: https://github.com/krausest/js-framework-benchmark/issues/243 I pinned this to only support the most recent version of Chrome.

krausest:
Thanks. I‘m on holidays this week - I‘ll merge it when I‘m back home.

krausest:
Results are [here](https://rawgit.com/krausest/js-framework-benchmark/b6228535b2b54de7bfe02483411e8aacaccf2a1c/webdriver-ts-results/table.html). Nice improvement, but create rows still has room for improvements.

mixonic:
Thanks for the re-run @krausest :-) there is definitely continuing room for improvement! The glimmer app libraries are being updated to work with the latest version of the VM now. I'll continue to push these updates to here so we can track improvement in this benchmark.

PR #245, opened 2017-09-12
Gruu v1.6.12 keyed

MarekLabuz:
I would be pleased and thankful for having Gruu introduced into the benchmarks.
Below there are the benchmark (50 runs) results compared to vanillajs-keyed.

gruu-benchmark

### DIFF
--- a/gruu-v1.6.12-keyed/.babelrc
+++ b/gruu-v1.6.12-keyed/.babelrc
@@ -1,0 +1,4 @@
+{
+  "presets": ["es2015"],
+  "plugins": ["gruu"]
+}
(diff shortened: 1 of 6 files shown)


krausest:
Thanks. Results are [here](https://rawgit.com/krausest/js-framework-benchmark/a5871ded4315a6179b3dac5dd825fdec082d6c41/webdriver-ts-results/table.html) - good to see they are close to your results. 

krausest:
Sorry, didn't check properly when I added gru. According to the test of nonKeyed.ts gru is non keyed:
gruu-v1.6.12-keyed is non-keyed for 'run benchmark' and keyed for 'remove row benchmark' non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results

gruu appears to reuse the table for successive calls to create 1000 rows. Can you change that behaviour? Otherwise I'd have to relabel gruu as non-keyed.

MarekLabuz:
Thank you, I will change the behaviour

PR #247, opened 2017-09-20
Angular 4.4.3

mlaval:
This PR updates Angular to version 4.4.3.
To improve performances, it also activates the compilation option to remove white spaces.


krausest:
Thanks. Faster than angular-4.3 (but please not that chrome versions might differ). Results are [here](https://rawgit.com/krausest/js-framework-benchmark/a5871ded4315a6179b3dac5dd825fdec082d6c41/webdriver-ts-results/table.html). 

PR #248, opened 2017-09-24
Upgrade to Vue 2.4.4

whtsky:


krausest:
Thanks. A bit faster than vue-2.3 (but please not that chrome versions might differ) and a bit faster than react 16-rc3. Results are [here](https://rawgit.com/krausest/js-framework-benchmark/a5871ded4315a6179b3dac5dd825fdec082d6c41/webdriver-ts-results/table.html). 

Issue #231, opened 2017-08-08
more memory metrics?


leeoniya:
Hey @krausest 

I'm in the process of hunting down a couple memory leaks in domvm and in my testing of other libs have found some very interesting mem profiles [1].

It would be quite informative to measure:

1. heap/nodes after startup
2. heap/nodes after render 1k rows
3. heap/nodes after updating 1k rows (5 cycles)
4. heap/nodes after replacing 1k rows (5 cycles)
5. heap/nodes after creating/clearing 1k rows (5 cycles)

for example, there's a children fast-clear optimization in domvm that does `parentNode.textContent = null` but causes unfavorable GC behavior due to not explicitly un-referencing the entire sub-tree and removing the cyclic refs between the old vtree and removed dom. i suspect there's a good amount of useful info (and lib trade-offs) hiding in these numbers - info that would've tipped me off sooner to the leaks.

[1] https://github.com/leeoniya/domvm/issues/164#issuecomment-320808039

krausest:
Should be pretty easy to add. I'll consider it when #232 is fixed which is very threatening.

krausest:
@leeoniya
I'm failing to see the difference between 3 and 4? I'd implement them both as clicking "create 1000 rows" 5 times or do you mean "update every 10th row" with 3?
Shouldn't 5 (at least for keyed frameworks) be equal to 4?

krausest:
First run - is there anything surprising?
<img width="577" alt="bildschirmfoto 2017-10-18 um 22 08 15" src="https://user-images.githubusercontent.com/3374055/31740204-e56fd2bc-b450-11e7-9e7e-a6a447ad33d5.png">

leeoniya:
> or do you mean "update every 10th row" with 3?

yes, i believe i meant that.

the idea is basically to cover scenarios where only the vtree must be recreated and then old one efficiently GC'd, which means 0 or minimal dom mutations. and scenarios where both the vtree and the dom must be re-generated (many mutations that force dom node regen). there are many places that can cause leaks, especially in long-running persistent apps. a variety of sequential tests helps ensure that discarded vnodes and dom doesnt accumulate over time.

those numbers look pretty good, i think.

actually the best test of whether these reveal anything useful would be to run the same bench against domvm before v3.2 an see how it does.

PR #249, opened 2017-09-24
update domvm v3.2.0

leeoniya:
hey @krausest 

i expect this update to be a bit slower as i've focused on closing mem leaks for long-lived apps. there have been some extra cleanup passes added, so the clear all rows perf should see a drop. the mem stats in this bench wont be affected because they only measure initial memory and not memory accumulated over time. i've found some interesting memory usage patterns for various libs [1].

do you have any thoughts on #231?

thanks!

[1] https://github.com/leeoniya/domvm/issues/173#issuecomment-321996899


krausest:
keyed
Please note that the OS and the Browser differs from the 3.0.5 measurement.

leeoniya:
thanks.

i'm seeing this in Win10 on Version 61.0.3163.100 (Official Build) (64-bit)
domvm-v3 2 0

PR #250, opened 2017-09-27
Update RE:DOM to v3.5.0

pakastin:
https://redom.js.org


krausest:
keyed
Please note that the OS and the Browser differs from the 3.0.2 measurement.

pakastin:
Thanks, still looking good! 😎👍

When is the next round of the blog post coming? I’d like to see RE:DOM included 😉

pakastin:
I just updated the docs to support mobile even better: https://redom.js.org/documentation 😉

PR #251, opened 2017-09-29
Add Angular v5 benchmarks

trotyl:
Add benchmarks for Angular v5 since it has came to RC (currently 5.0.0-rc.0).

- Standard ones (keyed and non-keyed) are added as companion for comparison (since v5 has not got final yet);
- Experimental ones (no-zone-keyed) are updated directly.

Highlights:

- Angular team is [currently using](https://github.com/angular/angular/blob/master/integration/hello_world__closure/package.json#L13) their own fork of closure compiler (with ability to detect entry in package.json), it's likely to be merged into closure itself later.


krausest:
keyed
Please note that the OS and the Browser differs from the 4.4.3 measurement.

PR #255, opened 2017-10-09
Update Datum to version 0.9.

MartinRixham:
I have just incremented the Datum version to the latest version.


krausest:
I didn't check earlier:  According to nonKeyed.ts datum-non-keyed is actually keyed (usually it's the other way round):

datum-v0.8.0-non-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' keyed for 'swap rows benchmark' . It'll appear as keyed in the results

Can you please check if that categorisation looks correct to you? Is the datum benchmark actually keyed?

MartinRixham:
yes I would say it is keyed

krausest:
@MartinRixham 
Selecting rows doesn't work for me. Can you check that please?
(No need to resolve index.html)

MartinRixham:
Ok it should work now with version 0.9.1

krausest:
Updated results are [here](https://rawgit.com/krausest/js-framework-benchmark/f730c37f8ea26bfe3f71254819d9396ae018af88/webdriver-ts-results/table.html).
According to my tests and my understanding datum is keyed and thus I've relabed datum as keyed.

MartinRixham:
thanks!

PR #257, opened 2017-10-16
add san.js

jiangjiu:
add san.js


krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/dec946ca63fb2254ec64bf2c10b0978f9b03f7ad/webdriver-ts-results/table.html)

PR #258, opened 2017-10-18
Use push.apply instead of arrayPushAll

### LINKED ISSUE #256: ko.utils.arrayPushAll is inefficient when called with an observable array
https://github.com/krausest/js-framework-benchmark/blob/c3488d6ba2619c9fa5a47639ecba9e691a822035/knockout-v3.4.1-keyed/src/Main.js#L59

`arrayPushAll` will make 1000 `push` calls, which each update the binding separately. Although one could argue that `arrayPushAll` should work better, this code should use the recommended approach:

    self.add = function () {
        startMeasure("add");
        self.data.push.apply(self.data, buildData(1000));
        stopMeasure();
    };

mbest:
Fixes #256

### DIFF
--- a/knockout-v3.4.1-keyed/src/Main.js
+++ b/knockout-v3.4.1-keyed/src/Main.js
@@ -56,7 +56,7 @@ var HomeViewModel = function () {
 
     self.add = function () {
         startMeasure("add");
-        ko.utils.arrayPushAll(self.data, buildData(1000));
+        self.data.push.apply(self.data, buildData(1000));
         stopMeasure();
     };
 


krausest:
Thanks it's merged, but I'm afraid it doesn't show any significant difference (left old, right new version):
knockout

mbest:
> I'm afraid it doesn't show any significant difference.

That's strange. I'll look into it.

mbest:
It turns out I misread what `ko.utils.arrayPushAll` is doing. It's actually calling `push.apply`, and so the two methods will have exactly the same performance. The extra time for the operation comes from Knockout comparing the old and new arrays to find the added items. I've included a change in Knockout for the next version so that it will not have to compare the arrays for such a change (a single, known operation). See knockout/knockout#2324

PR #262, opened 2017-10-22
gruu-v1.7.3-non-keyed

MarekLabuz:
I've updated the version and changed from keyed to non-keyed according to the test of nonKeyed.ts.


krausest:
Thanks. Results are [here](https://rawgit.com/krausest/js-framework-benchmark/04025c1f47d989efda365891f5ab7500082c907c/webdriver-ts-results/table.html)
Any idea why gruu doesn't like the new partial update benchmark with 10k rows?

PR #265, opened 2017-10-28
Update Polymer 2

CaptainCodeman:
Use Polymer 2 native class syntax
Reference built version


krausest:
Thanks. Results are [here](https://rawgit.com/krausest/js-framework-benchmark/8801573f1c7f864d11398248b4be8d70bb471275/webdriver-ts-results/table.html)
A nice improvement. Is there a way to do keyed updates with polymer?

PR #266, opened 2017-10-28
Lit-HTML implementation

CaptainCodeman:
See: https://github.com/PolymerLabs/lit-html

### DIFF
--- a/lit-html-v0.7.1-non-keyed/index.html
+++ b/lit-html-v0.7.1-non-keyed/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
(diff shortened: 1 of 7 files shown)


krausest:
Quite [nice](https://rawgit.com/krausest/js-framework-benchmark/ab4d197e995bb3310181e2cb74ec3cf617164ffb/webdriver-ts-results/table.html)!
(But I'm afraid I think frameworks really should support keyed updates - any chance for lit-html?)

CaptainCodeman:
Yes, there is a repeat extension for keyed updates but it's not optimized yet - as soon as it looks solid I'll add another variation for it :)

PR #268, opened 2017-10-29
add surplus 0.5.0-beta

griebd:
I couldn't update the index, but other than that everything seems ok.


adamhaile:
Hi Adriano -- You're welcome to post up the beta if you want, but I was holding off on bugging Stefan until I had the release version together.  I'm planning to change a few things in the app code, as a) some of the stuff it was doing has now been moved into the library, and b) I was going to make it a bit more of a "naive" implementation, less optimized.  It still comes in at about the same overall speed, because the library is a little faster and because I added one perf change around clears. Here, if you're curious, is what I've been using to test locally with 0.5.  The main difference is in the tbody, which is substantially simplified. -- Adam

```javascript
import * as Surplus from 'surplus'; Surplus;
import S from 's-js';
import { mapSample } from 's-array';
import { Store } from './store';

export const 
    AppView = (store : Store) => 
        <div className="container">
            <div className="jumbotron">
                <div className="row">
                    <div className="col-md-6">
                        <h1>Surplus v0.5.0</h1>
                    </div>
                    <div className="col-md-6">
                        <div className="row">
                            <div className="col-sm-6 smallpad">
                                <button type="button" className="btn btn-primary btn-block" id="run" onClick={e => store.run()}>Create 1,000 rows</button>
                            </div>
                            <div className="col-sm-6 smallpad">
                                <button type="button" className="btn btn-primary btn-block" id="runlots" onClick={e => store.runLots()}>Create 10,000 rows</button>
                            </div>
                            <div className="col-sm-6 smallpad">
                                <button type="button" className="btn btn-primary btn-block" id="add" onClick={e => store.add()}>Append 1,000 rows</button>
                            </div>
                            <div className="col-sm-6 smallpad">
                                <button type="button" className="btn btn-primary btn-block" id="update" onClick={e => store.update()}>Update every 10th row</button>
                            </div>
                            <div className="col-sm-6 smallpad">
                                <button type="button" className="btn btn-primary btn-block" id="clear" onClick={e => store.clear()}>Clear</button>
                            </div>
                            <div className="col-sm-6 smallpad">
                                <button type="button" className="btn btn-primary btn-block" id="swaprows" onClick={e => store.swapRows()}>Swap Rows</button>
                            </div>
                        </div>
                    </div>
                </div>
            </div>
            <table className="table table-hover table-striped test-data">
                <tbody onClick={(e : any) => e.target.matches('.delete') ? store.delete(rowId(e)) : store.select(rowId(e))}>
                    {clearAfterAnimationRequest(mapSample(store.data, row => 
                        <tr className={row.id === store.selected() ? 'danger' : ''}>
                            <td className="col-md-1">{row.id}</td>
                            <td className="col-md-4">
                                <a>{row.label()}</a>
                            </td>
                            <td className="col-md-1"><a><span className="glyphicon glyphicon-remove delete" aria-hidden="true"></span></a></td>
                            <td className="col-md-6"></td>
                        </tr>
                    ))}
                </tbody>
            </table>
            <span className="preloadicon glyphicon glyphicon-remove" aria-hidden="true"></span>
        </div>,
    rowId = (e : any) => +e.target.closest('tr').firstChild.textContent,
    // For not well understood reasons, Chrome runs DOM clear operations a good bit faster when they're
    // deferred until after an animation frame.  So we buffer the row signal in a data signal and delay
    // its update when the signal would clear the contents.
    clearAfterAnimationRequest = function deferAF<T>(s : () => T[]) {
        var data = S.data(S.sample(s)),
            updater = () => data(s());
        S(() => s().length === 0 ? requestAnimationFrame(updater) : updater());
        return () => data();
    };
```

griebd:
Hi @adamhaile, I made it mainly for two reasons:

1. I don't have chrome and don't plan to install it... (if only I could escape firefox TOO... :wink:) and I wanted to see how the new version is going on this benchmark.
2. I thought to let people know the framework is going forward when @krausest publish a new round.

So, it's your choice as owner of the framework! If you prefer to have only the final version we can just close this pull request. Or if you agree I can help you maintain this benchmark implementation to help you (beta or final, no compromises).

adamhaile:
Hi Adriano --
In terms of performance, 0.5 aims to be fast in more cases than 0.4.  It tries harder to do the minimal or fast-case DOM ops than 0.4 did.  The goal is to make "naive" implementations run faster. 

Here's the results of running the same code used in the 0.4 benchmark with the latest 0.5 master branch:

screen shot 2017-11-02 at 6 50 35 pm

Like I said, I'm planning to submit a more naive implementation when 0.5 is ready, since the performance cost isn't as large.  I also added one optimization, around clearing nodes, that ends up balancing out the cost of the others, so the overall result comes in about the same or a bit faster:

screen shot 2017-11-02 at 7 19 12 pm

griebd:
so, let it for the final version. thanks for the report!!!!

PR #269, opened 2017-10-30
Increment Datum version to 0.9.2

MartinRixham:
I've tried to fix some performance issues.


krausest:
Results are [here](https://rawgit.com/krausest/js-framework-benchmark/e85496bc82b0ff0472b7626498add7f4e7cca414/webdriver-ts-results/table.html)
A lot faster than 0.9.1:
bildschirmfoto 2017-11-02 um 22 24 59

Issue #270, opened 2017-11-01
add maquette


leeoniya:
hey @johan-gorter, now that maquette v3.0 is out, is there any interest in joining the party? :)

this bench is pretty easy to implement but if you dont have time, i can try to slap something together and you can tune it as needed.

https://github.com/AFASSoftware/maquette/issues/38

johan-gorter:
Sure we want to join this party. I currently do not have time to get much work done, but it should be enough to do the tuning. I already implemented the uibench benchmark, so you can find implementation clues [here](https://github.com/AFASSoftware/maquette/blob/51a291bb7c2bb8343417a3c46107edc72a86ba02/examples/benchmarks/uibench/app.ts#L142).

From my experience with uibench, I know that maquette is neither the fastest nor the slowest framework when it comes to pure DOM manipulation. It should be one of the fastest when there are no changes. We also want to build a [feature](https://github.com/AFASSoftware/maquette/issues/118) which allows you to minimize costly reflows when measuring DOM nodes.

leeoniya:
i made a keyed impl here: https://github.com/krausest/js-framework-benchmark/pull/271.

overall perf looks quite good. "select row" is a bit of an outlier. i also noticed that swap rows appears to touch every node in between.

johan-gorter:
I can confirm that maquette does not reuse DOM nodes when swapping, but it can still [animate a swap](http://johan-gorter.github.io/maquette-demo-hero/). A long in-depth discussion about this can be found here: https://github.com/AFASSoftware/maquette/issues/13.

johan-gorter:
A non-keyed implementation is not allowed in maquette, this is maquette's [rule#2](https://maquettejs.org/docs/rules.html)

johan-gorter:
Question: maquette's `projector` uses requestAnimationFrame to delay rendering to make sure it never renders more than 60 frames per second. I see that the benchmark uses the `projector`, so how can you measure how long the rendering took exactly?

johan-gorter:
Hello,

I finally had a chance to run the benchmark locally. It looks like using requestAnimationFrame or not is not affecting the outcome significantly. This means that the benchmark does a good job at approximating real world usage!

[...]

PR #271, opened 2017-11-01
maquette-v3.0.1-keyed

leeoniya:
addresses https://github.com/krausest/js-framework-benchmark/issues/270

I did a pretty straightforward port of domvm. considering that the only optimization is event delegation, it performs pretty well. i couldn't find a way to make maquette swallow a non-keyed impl; it bitches at runtime about needing keys when it detects similar sibling nodes.

this impl may need slight tweaking, since i just took the UMD build and dumped it into `/src`. i added it to `package.json` but didnt bother working out exactly how to require/import it from there.

maquette-keyed


krausest:
Sorry - I'm getting an exception when clicking "create 1,000 rows":
Uncaught ReferenceError: getProjection is not defined
    at HTMLButtonElement.<anonymous> (bundle.js:1)
Any idea how to fix it?

leeoniya:
@johan-gorter

> By the way, I somehow had to disable uglify in build.js, because it created errors at runtime (getProjection was undefined).

https://github.com/krausest/js-framework-benchmark/issues/270#issuecomment-342976549

johan-gorter:
One more remark about this pull request, could you please include maquette from node_modules instead of the copy that is inlined. This will enable me to try out new approaches using `npm link`. @leeoniya will you update this pull request or shall I start my own?

leeoniya:
yeah, just make a new pr. i'll close this one rather than having to maintain it.

PR #272, opened 2017-11-03
Update Datum to version 0.10.0

MartinRixham:
Sorry to keep sending you pull requests, but I think there should be some more performance improvement in this version.

thanks again


krausest:
Nice - a good [speedup](https://rawgit.com/krausest/js-framework-benchmark/19f9fb672c6aa602230f52aa2f2b955cc706e24a/webdriver-ts-results/table.html).

MartinRixham:
That's great news.

I'd like to thank you alot. Your benchmark has helped me not just to get Datum running quickly, but also to root out a few annoying bugs on the way. I feel I can now really recommend it as a tool that provides an expressive language for writing user interfaces, is easy to learn and promotes good development practices like separation of layout and logic.

I'm going try and leave it alone for now since it is pretty much feature complete and I've spent WAY more time on it than originally intended. Though of course I'll fix any bugs that get reported. There is also already a 0.11 version that provides a test API that should allow all user interactions to be tested through the view model without binding any HTML.

It would be really handy if when you next publish the benchmark results you could mention the Datum homepage datumjs.com

Thanks again

PR #274, opened 2017-11-05
update Angular to 5.0.0 stable

trotyl:
Angular just released the stable version of 5.0.0, update the dependency to that.


krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/1922d6b922816b5b7f5c81fecf7a950df7bc005e/webdriver-ts-results/table.html).

PR #275, opened 2017-11-06
Upgrade slim-js to 3.0.2

eavichay:


krausest:
Thanks - much [faster](https://rawgit.com/krausest/js-framework-benchmark/1922d6b922816b5b7f5c81fecf7a950df7bc005e/webdriver-ts-results/table.html) than the previous result. Does slim.js support keyed updates?

eavichay:
I will, probably soon.
Now can I say it’s faster than react?

krausest:
That's why I created the compare with dropdown ;-)
bildschirmfoto 2017-11-08 um 22 05 18
slim.js is significantly slower than react for create rows, remove row, create many rows and append rows. On the other hand it's statistically significantly faster for clear rows and startup time (partial update is light green and thus not significant at a 5% level). 
So it's a bit dangerous to say so. (But it's really fast for clear rows and startup is brilliant ;-)

And I'm afraid if we leave out startup things turn better for react:
bildschirmfoto 2017-11-08 um 22 12 11

Issue #104, opened 2016-12-21
split into 2 tables "keyed" and "non-keyed" ?


leeoniya:
Looking at the table [which is getting way too wide IMO], the perf numbers are very difficult to compare because many impls are keyed and many are unkeyed. these are really 2 completely different perf profiles.

if you prefer to have a single huge table at the very least, it would make sense to order/group them by keyed vs non-keyed. a trivial way to see this is to assume everything < 130ms for "replace all rows" is unkeyed and > 130ms is keyed.

i think 2 tables will make sense and would provide much more meaningful comparisons (and also not span a 2 x 1080p screens :)

the vanillajs-keyed impl would need to be fixed for this as well: see https://github.com/krausest/js-framework-benchmark/issues/103

krausest:
@leeoniya 
I think that's a very good idea, I updated the results table in 9a548fc024481b7acbe4eb215a1d5ca348307bf7. I tried to add some explanation for keyed / non-keyed. Feel free to comment if you had a better one ;-)

I expected the non-keyed impl of vanilla-js to perform faster for remove than the keyed impl, but the reported duration didn't show that. It turned out that if I revert the icon resizing transition the results are as expected a bit faster for non-keyed. I haven't reverted the css in the repo yet.

There are a few implementations that behave like keyed for the remove operation that run very fast for replace all rows. I'll have to look into that the next few days...

krausest:
@leeoniya 
ractive is really interesting. For Remove it acts like the other keyed frameworks and the transition stops. If I add the following in the run method:
```
        let elem = document.querySelector("tr");
        if (elem) {
            elem.style.backgroundColor="#f00";
        }          
```
It behaves like non-keyed for repeated create 1000 row (the red row stays visible).

leeoniya:
thanks!

i'll take a look at Ractive tonight. i'll can help you debug to ensure all the impls are doing the proper dom ops. there are several keyed impls < 140.83 for "replace all rows"...which is highly suspect.

the keyed vs non-keyed explanation looks good.

leeoniya:
ok, I tested

trotyl:
@krausest 
I think the framework/library that supports both *keyed* and *no-keyed* should be list in both table, since the user could choose which implementation to use according to their needs.

leeoniya:
ok, I tested these and as expected they are all non-keyed:

- simulacra v1.5.5
- ractive-edge
- ractive v0.8.5

[...]

PR #277, opened 2017-11-07
Update dojo2 to v0.2.0

agubler:
Update Dojo 2 dependency to 0.2.0


krausest:
Thanks, but I'm afraid this version isn't keyed according to our definition here:
```
npm run nonKeyed -- --framework dojo2-v0.2.0-keyed
dojo2-v0.2.0-keyed is non-keyed for 'run benchmark' and non-keyed for 'remove row benchmark' non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
ERROR: Framework dojo2-v0.2.0-keyed is not correctly categorized in commons.ts
```
I prefer keyed frameworks (and maybe I'd even refuse to use a non-keyed framework). Any chance to get back a keyed version?
Until then I'll rename it to non-keyed.

agubler:
@krausest Interesting so the definition of keyed isn't that there is an actual `key` being used but that the elements are "removed" and "created" as part of the tests?

krausest:
Yes. I think that would be necessary for e.g. transitions to work correctly on key change.

krausest:
See also https://github.com/krausest/js-framework-benchmark/issues/104#issuecomment-272099815

agubler:
Yes that makes sense for when that's needed, have raised a keyed impl here https://github.com/krausest/js-framework-benchmark/pull/278

PR #278, opened 2017-11-08
Dojo 2 keyed implementation

agubler:
Add dojo 2 keyed impl


agubler:
Closing, will reopen with the latest release that is due tomorrow.

PR #284, opened 2017-11-10
domvm v3.2.2

leeoniya:
no significant lib changes, but it now pulls the latest published version from npm rather than the git repo at some specific commit. should make future updates slightly easier.

thanks!


krausest:
Updated results are [here](https://rawgit.com/krausest/js-framework-benchmark/10a4966b86ba6b72af31344c7c78d8752ae9c1af/webdriver-ts-results/table.html).

PR #289, opened 2017-11-11
Allow supplying custom path to chrome executable

alexfmpe:


krausest:
I just added a check to set the option only when it's defined (though it worked as it were).

alexfmpe:
Yeah, a lot of the `.setX`functions seem to do nothing when supplied `null`or `undefined`, but this isn't always documented, so that's safer.

Issue #294, opened 2017-11-12
Swap rows doesn't work correctly for thermite, halogen and pux


krausest:
Click "create 1,000 rows", the swap rows. The id of the rows 5 and 10 remain unchanged, which is wrong.
I'll disable them until the bug is fixed.

krausest:
@thomashoneyman Can you take a look at it please?

thomashoneyman:
See merged fix here:
https://github.com/krausest/js-framework-benchmark/pull/296#issuecomment-343763975

PR #296, opened 2017-11-12
Fix issue 294 - Update to "swap rows" implementation

thomashoneyman:
Hi @krausest,

This is a small fix. I had misunderstood the implementation and preserved row IDs after being swapped; this change no longer does that.

This fixes:
https://github.com/krausest/js-framework-benchmark/issues/294

I did not uncomment the `common.ts` file so the frameworks are still disabled in this PR.


krausest:
Thanks. They will be re-enabled soon (and included in the next round).

PR #300, opened 2017-11-14
update surplus to v0.5.0

adamhaile:
The keyed implementation is considerably more "naive" in this go round, as many of the things that used to be handled in the implementation are now done by the framework.  There are really only two things beyond a truly naive implementation:

1. I'm still delegating the click handlers by hand
2. I'm buffering row changes to take advantage of fast rAF clears

I'm still thinking through how/if to move some of the non-keyed features into the framework, so it's largely as before.

Thanks again, Stefan, for the benchmark and for including surplus!  This is a really great resource for framework authors.


adamhaile:
Hi Stefan -- I'm not happy with the amount of handcoding in the non-keyed version.  If you're just about to publish version 7, go ahead and merge, but if you're at least a couple days away, I'm going to revise the PR with a simpler version.  -- Adam

krausest:
The longer I think about the more I doubt that anyone could be happy with a non-keyed version ;-)
I plan to publish the next round soon - but currently webdriver is bugging me)

adamhaile:
I should update the Surplus implementation, if nothing else b/c as per our other discussions, I was planning to remove the rAF buffering.  I see you going through a bunch of updates now.  If you're planning another official round soon, I'll get one out quickly.  If not, I may wait until a couple things I'm playing with are finished.  Thanks!

krausest:
@adamhaile I plan to create official results for Chrome 68 (i.e. soon), so it would be cool if you sent a PR. I'm still a bit undecided regarding #400 whether I should change the directory naming scheme wich might cause a delay.

PR #302, opened 2017-11-16
Dojo 2 Keyed Implementation

agubler:
Adds a keyed implementation for Dojo 2 and updates the versions for the non keyed version.


leeoniya:
here's what i got vs vanilla on Win10, Chrome 62:

dojo2

agubler:
@leeoniya nothing looks particularly unexpected there. Thanks!

PR #308, opened 2017-11-20
update domvm v3.2.4

leeoniya:
this update has 2 changes. one is not relevant to this bench (a bugfix for imperative views, which aren't used here). the other is removal of a build hack that forced Closure not to inline `syncChilden()` which resulted in a significant deopt. Chrome seems to have fixed the deopt in v62. this may affect perf a bit.

additionally, this PR further simplifies maintenance of updates by removing version numbers from the title tag and impl.


krausest:
Updated [Results](https://www.rawgit.com/krausest/js-framework-benchmark/67704fafce83e268968e545be248b714655ffa31/webdriver-ts-results/table.html):
bildschirmfoto 2017-11-21 um 22 25 45

leeoniya:
thanks, no surprises here. i think it's mostly noise.

Issue #306, opened 2017-11-20
use 1/998 for swap rows rather than 4/9


leeoniya:
@krausest 

i did a bit of testing as a followup to https://github.com/krausest/js-framework-benchmark/issues/305#issuecomment-345538990

dunno if you wanna sneak this change into Round 7, but the results are rather revealing:

![swap-rows](https://user-images.githubusercontent.com/43234/33000432-a7a51d6c-cd6d-11e7-91ec-4e5b4a4d8cc8.png)

leeoniya:
the plot thickens...

apparently some libs are micro-optimized only for leading/trailing swapping rather than a more general reconcile algo. so when we switch to 1/998 instead of 0/999, we get to see prettier autumn colors:

![swap-rows2](https://user-images.githubusercontent.com/43234/33002127-b5d7e8ce-cd77-11e7-8fa5-0df5c874b497.png)

krausest:
Will run the swap benchmark and post results this evening.

krausest:
Here they [are](https://rawgit.com/krausest/js-framework-benchmark/9e80ef2cedcd9494756240456481608499e488d9/webdriver-ts/table.html).

leeoniya:
hmm, is `choo v6.5.0-keyed` misclassified?

krausest:
Thanks for spotting it. At least its naming was already correct...
Fixed in ee308dc2ad38688bc6d1ac1c0313c06ed40f21e8

Issue #305, opened 2017-11-19
Proposal for new tests(Moves)


thysultan:
The performance of these tests would give a good idea of the number of DOM operations that different framework reconciliation heuristics can guarantee.

1. move first row to last.
1. move last row to first.

The worst case for a single move. Frameworks that execute only one DOM op would perform better than ones that move more than one Node to archive the reconciled state.

3. remove first row + append new row.
4. remove last row + prepend new row.

Common in infinite list implementations and charts. Tests whether the framework appends and moves multiple Nodes to archive a reconciled state vs a single remove + append operation to archive the least DOM ops and the least DOM state disruption. Frameworks that do the former would intuitively perform better.

5. reverse list.

Tests what happens when everything moves, similar to the "replace all rows" test but for movement.

leeoniya:
i don't think `5. reverse list` is necessary since it's a nested variation of swap rows (for bi-directional reconcilers.). i would expect any libs that get swap rows right to also do well in reversing the list. but perhaps not.

one thing i do think should be considered is to increase the `swap rows` distance, eg first/last. as discovered recently, maquette does not reuse nodes while swapping [1] (which is another topic of whether it can be considered keyed). the increased distance should accentuate what each lib actually does during the swap. right now mutating 8 nodes in between the two swapped ones is relatively cheap, which would not be the case if 998 nodes had to mutate during a swap.

[1] https://github.com/krausest/js-framework-benchmark/issues/270#issuecomment-341418345

thysultan:
Yes, i think a larger distance between the swapped rows would also archive a similar effect as the reverse list test as demonstrated in #390.

leeoniya:
@thysultan can you do some testing of 1-4 for a few fast and slow libs and see if anything unexpected shows up?

thysultan:
This forks "new-tests" branch implements them for a few frameworks, [https://github.com/thysultan/js-framework-benchmark/tree/new-tests](https://github.com/thysultan/js-framework-benchmark/tree/new-tests).

leeoniya:
thanks! i'll give them a run. probably worth updating the swap to 1/998 instead of 0/999, too [1]

[1] https://github.com/krausest/js-framework-benchmark/issues/306#issuecomment-345588019

thysultan:
Current results for that and 1 & 2, Haven't yet added 3 & 4 to webdriver tests. Also seems the vanilla impl for 2 is missing something.

[...]

PR #310, opened 2017-11-21
update impls to swap rows 1/998 instead of 4/9

leeoniya:
this implements #306 for all frameworks except reflex-dom (i dunno what the hell to do there). please double check these changes. i would be lying if i said i tested them all.

i will be working off of this branch when testing the additional metrics proposed in #305


krausest:
Thanks - 🥇. What a PR!

PR #313, opened 2017-11-21
hyperHTML keyed only version

WebReflection:
This PR goal is to remove the non-keyed version of hyperHTML
and bring a more benchmark oriented version of the keyed one.


WebReflection:
dear all, there are already developers asking me about the current published table (which is wrong in many ways) ... by any chance this can be merged and a new table re-published with updated results for hyperHTML and on the right, keyed, table?

Thank you

krausest:
Table in master is [updated](https://rawgit.com/krausest/js-framework-benchmark/45bae33310ad815e3e6ebbff9d4a5ab6612d2304/webdriver-ts-results/table.html) now.

PR #318, opened 2017-11-22
hyperHTML with Majinbuu engine

WebReflection:
Since hyperHTML has swappable DOM engines, I would like to add the glorious Majinbuu which would handle the row swap in a snap.

Thanks for considering this PR.

majinbuu


hyperhtml-v2.1.3-majinbuu-keyed/rollup.config.js
```diff
@@ -0,0 +1,35 @@
...
+export default {
+  input: "src/index.js",
+  output: {
+    file: "dist/index.min.js",
+    format: "iife",
+    name: "app"
+  },
+  plugins: [
+    alias({
+      'https://unpkg.com/majinbuu@latest/esm/main.js': 'node_modules/majinbuu/esm/main.js'
```
douglasduteil: Why is this "alias" required ?

WebReflection: I can push a version that uses `rollup-plugin-cdn` but the point is that the bundler consumes hyperHTML from locally installed dependency and the external resource is only to make hyperhtml-majinbuu usable via browsers right away without needing intermediate bundlers steps.

It's a developer convenience, it doesn't affect at all this test or users of `hyperhtml-majinbuu`.

At the end, ESM is natie on browsers, and on browsers that's right now how it works.

When unpkg will be not-experimental anymore, this _could_ go away.

Since you need rollup or a package anyway, I don't see issues in the current setup.

Would you prefer `rollup-plugin-cdn` that does not need hard coded remote urls in the config?

WebReflection: actually ... let me do that for you ...

WebReflection:
... how to move forward in here?

PR #322, opened 2017-11-28
Added synchronous init to Bobril. Prettier run on app.ts.

Bobris:


krausest:
Thanks. Updated results are [here](https://rawgit.com/krausest/js-framework-benchmark/0f276675ad4f3cf940b6e5dcd1c3662fb503d5a4/webdriver-ts-results/table.html). As expected except for startup almost no change.

PR #324, opened 2017-11-29
Run all circleci node commands in the same shell

### LINKED ISSUE #323: Broken node (un)installation on circleci
Want me to take a look @krausest ?

alexfmpe:
Fixes #323.

I don't really like this solution (and avoided it at first) since it makes logs messier and it's harder to see how much time each command took, but having to account for $PATH being reset and resourced is a pain, _especially_ when it's changed by scripts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,17 +17,8 @@ jobs:
       - run: |
           source $NVM_DIR/nvm.sh
           nvm install v8.9.1
-          nvm use v8.9.1
-          nvm alias default v8.9.1
-          node -v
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -17,17 +17,8 @@ jobs:
       - run: |
           source $NVM_DIR/nvm.sh
```
alexfmpe: Tbh, this is kind of dodgy.
I'll try to find a more reliable alternative. 
Maybe [avn](https://github.com/wbyoung/avn)?

PR #325, opened 2017-12-01
Add more automation/perf-friendly chrome flags

paulirish:
In chrome-launcher we have a [larger list ](https://github.com/GoogleChrome/chrome-launcher/blob/a48c47bd305e6ca137b6c94c7fc0b33660be71fe/flags.ts) of flags that we flip on for the browser. This adds in the ones I think make most sense for this project.


leeoniya:
thank you! will try this out later today :)

Issue #316, opened 2017-11-22
Is an imperative vanillajs a reasonable baseline for a data-driven bench?


leeoniya:
@krausest 

This issue originates from https://github.com/krausest/js-framework-benchmark/issues/305#issuecomment-346257613.

There are some cases - possibly many cases - where even efficient reconcile algorithms account for 99% of the perf measurement whereas vanillajs is able to avoid all reconciliation costs (as it does in swap rows, and possibly others). This baseline is highly unrealistic if the aim is to show the fastest possible real world data-driven implementation.

Perhaps a controversial view, but i think the baseline should be determined by the fastest data-driven implementation. There is now a sufficient quantity of extremely fast libs that handily beat vanillajs. Rather than trying to adapt vanillajs to be unrealistically faster, it may be worth dropping it altogether?

If we keep adding edge cases and implementing just the dom ops in vanilla, the metrics will quickly become misleading.

/random thoughts

cc @localvoid @thysultan @hville @adamhaile @yelouafi @Bobris

Bobris:
I have peace with any Vanillajs implementation. Actually I think it is perfectly ok to "cheat" as much as possible to see overhead of our libs to ideal DOM manipulation. I hope there are no devs who would not consider using lib just because they would think they could do better without.

leeoniya:
the issue i have is that there would never be a situation in real life that explicitly dictates how things need to be reordered as in "take first and second to last and append to end", therefore a vanillajs implementation of exactly these actions is meaningless, because no one would write such a manipulation in real code. this means that the overhead introduced is also meaningless except when compared to other general reconciliation algos.

Bobris:
I can imagine there could exist observable framework (think Mobx blended with Svelte) where swapping data in model would automatically swapped rows in dom. So this particular operation could have minimal overhead in script. Of course it would be probably very difficult to implement and would be slower in other operations, but who knows...

eavichay:
I think that the purpose of this benchmark is to focus on rendering / memory usage / startup time.
The "overhead" of state management X or data management Y are not part of the game. If a framework is bundled with a state management, it's either helps it or not.

hville:
Dropping vanillaJS and using the fastest implementation reduces the maintaining efforts (code quantity, setup and benchmarks times).

[...]

PR #326, opened 2017-12-01
use lighthouse to calculate metrics from CPU benchmark

paulirish:
Thanks to your ping in https://github.com/krausest/js-framework-benchmark/issues/316#issuecomment-348297110 I spent some time on this so we could get a TTI metric calculated for you. Basically I'm just using Lighthouse to [process the perfLog](https://github.com/GoogleChrome/lighthouse/blob/master/docs/readme.md#lighthouse-as-trace-processor).

image

I wasn't sure where this belongs between the CPU/MEM/Startup benchmarks, so I stuffed it into CPU. But the nice part here is that the only instrumentation changes are flipping on `enableNetwork`/`enablePage` and recording a few additional trace categories.

For now, I'm just logging the results to the console, though I'm sure you probably want to redirect the values somewhere more useful. 

As for the metrics themselves (and how they apply to this benchmark):

* **first contentful paint** - will be affected by render blocking resources, so the size and number of them. 
   - Probably not terribly useful
* **first meaningful paint** - the paint following the most significant layout during load. 
   - If the initial painting of the DOM is key, then it's useful. But I can imagine it might catch some of the DOM changes that the benchmark triggers.. in which case I don't think it'd be a good signal.
* **onload** - it's normal onload. 
   - seems like its basically[ your durationJS](https://github.com/krausest/js-framework-benchmark/issues/316#issuecomment-347670958)
* **first interactive** - an optimistic TTI - the first moment the main thread stays idle for a little bit.
* **consistently interactive** - a pessimistic TTI - when the CPU and network are both definitely very idle. (no more CPU tasks over 50ms)
* **script bootup time** - the total ms required to parse/compile/evaluate all the page's scripts.
* **main thread work cost** - total amount of time spent doing work on the main thread. includes style/layout/etc. 
   - seems useful for this application, for sure.
* **estimated input latency** - kind of represents how slammed the main thread was. higher numbers are bad. 
   - it's tuned for normal pageloads, so i'm not sure how good a signal it is for here.
* **total byte weight** - network transfer cost (post-compression) of all the resources loaded into the page.

hope this is helpful!!

### DIFF
--- a/webdriver-ts/package.json
+++ b/webdriver-ts/package.json
@@ -28,8 +28,10 @@
   },
   "dependencies": {
     "chromedriver": "2.33.2",
+    "columnify": "^1.5.4",
     "dot": "1.1.2",
     "jstat": "1.7.1",
+    "lighthouse": "^2.6.0",
     "lodash": "4.17.4",
     "ramda": "^0.25.0",
     "selenium-webdriver": "3.6.0",
--- a/webdriver-ts/src/benchmarkRunner.ts
+++ b/webdriver-ts/src/benchmarkRunner.ts
@@ -3,6 +3,10 @@ import {Builder, WebDriver, promise, logging} from 'selenium-webdriver'
 import {BenchmarkType, Benchmark, benchmarks, fileName} from './benchmarks'
 import {setUseShadowRoot} from './webdriverAccess'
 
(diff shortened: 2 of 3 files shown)


krausest:
Great stuff! I'll investigate how to integrate the results best into the result table.

PR #329, opened 2017-12-05
hyperHTML v2.4.0 + removal of hyperhtml-majinbuu

WebReflection:
Thanks to this benchmark, hyperHTML changed direction,
back to simplicity, and zero runtime, third parts,
swappable engines.

This PR would like to remove the currently pointless
hyperhtml-majinbuu benchmark, and introduce
at the same time the latest hyperHTML based on domdiff [1]

[1] https://github.com/WebReflection/domdiff


krausest:
Thanks. Results are [here](https://rawgit.com/krausest/js-framework-benchmark/e912ef51a61abbf22af7239eea5de01f69c9e649/webdriver-ts-results/table.html).

WebReflection:
Thanks a lot !!! As side note, I still see majinbuu version in there ... although, it's good to see it **after** the latest 2.4 :1st_place_medal:

krausest:
Yes, that's my special service that I try to publish both the old and the new results in the table.html for the PR. It'll be gone in the next results table.

PR #335, opened 2017-12-12
clamp metrics with delta of < half a frame (~8.3ms)

leeoniya:
@krausest 

in general i'm not a huge fan of clamping, but...

the current clamping behavior is heavily biased towards lightweight metrics/workloads (those that finish within 1 frame) rather than those that *differ* by 1 frame or less.

if you look at "select row", almost everyone is at 1.0 because they're < 16ms and the range is 7.5 - 15.0. now look at "partial update". the minimum is 72.0 but everything that can also finish within the same frame (up to 72 + 16.667 = 88.667) is marked with a factor larger than 1.0. a more ridiculius example is "swap rows" with a baseline of 17.0 which is already at least 2 frames, but everything else that will finish within these 2 frames (from 18.0 up to 33.334) is penalized.

what this PR does is make the clamping behavior more logical. the change will clamp any differences smaller than half a frame to 1.0, regardless of the actual workload or absolute timings. it will have the effect of making "swap rows" 1.0 from the baseline of 17.0 to ~25.3. this entire range will certainly be completely transparent to the user since everything here will take 2 frames. as i understand it, clamping is supposed to mask imperceptible differences, and this method does a better and more uniform job of it across all metrics, i think.

### DIFF
--- a/webdriver-ts-results/src/Common.tsx
+++ b/webdriver-ts-results/src/Common.tsx
@@ -236,7 +236,7 @@ export class ResultTableData {
             if (result === null) return null;
             else {
                 let mean = this.useMedian ? result.median : result.mean;
-                let factor = clamp ? Math.max(16, mean) / Math.max(16, min) : mean/min;
+                let factor = clamp && (mean - min < 16.667/2) ? 1 : mean/min;
                 let standardDeviation = result.standardDeviation;
 
                 let statisticalResult = undefined;


krausest:
I'm basically fine with the PR, but one case should be reconsidered:
bildschirmfoto 2017-12-12 um 21 24 03
In that case I think the old result was a bit better:
bildschirmfoto 2017-12-12 um 21 26 15
Maybe compute it like that?
```
                let factor;
                if (clamp) {
                    factor = (mean - min < 16.667/2) ? 1 : Math.max(16.667, mean) / Math.max(16.667, min)
                } else {
                    factor = mean/min; 
                }

```
Though it looks a bit more boring in this special case:
bildschirmfoto 2017-12-12 um 21 41 37

leeoniya:
you can always give more leeway and use a full frame rather than half frame for clamping. you can also do much more complex stuff like use a log scale for the factors based on min duration, but this is likely excessive.

there's probably more to be done here in general. for instance, even if the colors imply no perceptible difference, it would be good to show the actual, real factor in the text. it might also be cool to add a slider that adjusts the clamping threshold so that authors can get a more granular/colorful table while users get the default "real world" colors.

leeoniya:
> Maybe compute it like that?

that looks reasonably sane, though my dream of having the threshold tweakable would now need 2 settings rather than 1.

regardless, i'd like to see the real factor shown, since now there are 1.0s everywhere and it's tricky to spot where vanillajs is not *actually* 1.0, which is necessary to optimize vanilla and also spot problematic or cheaty implementations.

i'd still prefer to have a single clamping threshold and just bump it till things look as expected without being dependent at all on any metric's workload (they're all artificially selected anyhow).

krausest:
Of course we could output the real value (e.g. just make the clamp flag a checkbox), but I wonder if that's reasonable. 
1. It might render RAF-based frameworks slower than immediate rendering frameworks. 
2. And I'm concerned that it would result in comparisons that are statistically void. Just remove clamping and compare against a fast framework. There are very little significant comparison values in there and thus labelling one framework faster or slower isn't really sound:
bildschirmfoto 2017-12-12 um 22 23 54


I think it's only safe to say that except for a few frameworks all are fast enough for select row.

leeoniya:
i really think simply introducing a clamping level slider that's preset to a reasonable level and maybe with 1/2 frame steps would be ideal. in that case, it would not be necessary to show the "real" factors either and we can leave everything else as-is, allowing the clamping level to dictate the shown factor.

a slider range of 100ms may be reasonable. this woks out to exactly 6 frames (or 12 half-frame or 24 quarter-frame steps). we may need to get down to 1/4 frame to dial in the default, which looks like it may be at 3/4 frame.

https://www.pubnub.com/blog/2015-02-09-how-fast-is-realtime-human-perception-and-technology/

leeoniya:
would you be interested in an impl of a 6-frame/24-step clamping slider?

krausest:
I guess yes (if I'm convinced it makes sense).
What would the computation for the slider look like if dt was the slider's effective duration?
```
                let factor;
                if (clamp) {
                    factor = (mean - min < dt) ? 1 : Math.max(16.667, mean) / Math.max(16.667, min)
                } else {
                    factor = mean/min; 
                }
```
I'm under the impressing you don't like the max thing with 16.667, right? Would you choose `Math.max(dt, ...)`? But that makes little sense for anything but 1 frame to me.
Did you answer my objection that RAF-based frameworks would be discriminated if clamping was below 1 frame?

leeoniya:
i was thinking of doing exactly what this PR does, except with an adjustable `dt`, preset to something that makes sense, which maybe is 3/4 frame. as for rAF-based libs [of which domvm is one], i dont have a clear understanding yet of where they'll land. TBH, i don't properly understand what effect rAF currently has on metrics in this bench aside from the odd benefit it has for clearing rows.

here's another attempt to describe the `max` situation. on the surface, it's reasonable to assert that no op can take less than 1 frame to show a visual update. but treating 1 frame as an upper threshold will only cause clamping of super cheap ops. instead, you should be clamping at *all* frame boundaries, not just frame 0 boundary. in effect, if an op takes 117.5ms, then it's 117.5/16.6667 = 7.05 frames. the next step up would be 8 frames = 133.34ms. everything from 116.67 to 133.34ms should then be treated as equivalent. this way we're basically saying these ops will complete in the same # of frames.

the factors will then be simply computed as (# of whole frames required for mean)/(# of whole frames required for min). this may be the most sensible and uniform clamping behavior. we can then add a simple checkbox instead of a slider to enable or disable frame clamping.

leeoniya:
here's what this looks like. using median is a bit more odd since picodom somehow manages to come in under 1 frame.

### mean:

mean

### median:

median

adamhaile:
Forgive my possibly dumb question.  You write:

> everything else that will finish within these 2 frames (from 18.0 up to 33.334) 

That would only be true if click events always execute at the front edge of a monitor frame, right?  Is that true?  I naively assumed that click events fire at the time of the click, which may be any point between frames.  So the total time would be time to the end of the paint event plus 0-1 frame, depending on where the end of the paint event fell in the monitor's refresh cycle.  So why have any clamping?

leeoniya:
honestly the whole situation is weird. i would rather completely ditch the clamping and simply increase all the cheap workloads to avoid this situation (similar to the recent move from 1k to 10k for updates). instead of select 1 row, select 200. instead of swapping 2 rows, swap 20.

leeoniya:
> instead of select 1 row, select 200

okay, this may not be viable because it'd burden the implementations with significant changes as they can no longer have a single `selected` to check against.

how about just taking the sum of all runs rather than the average (and possibly dropping slowest 1 or 2). this way the aggregate provides a form of normalization without clamping.

i also don't see any evidence that rAF-based libs are being affected by the measurement methodology. therefore i don't think that there will be any sort of error accumulation of "unused" frame time when summing it all.

leeoniya:
i'm going to do some testing and try to elimnate the need for clamping. herecs what i have in mind:

- move every metric to 5k, drop 1k and 10k
- use the sum of runs rather than the mean
- maybe drop slowest one or 2 (though this may mask GC jank)
- maybe do 15 rather than 10 rounds

leeoniya:
@krausest was there a specific reason for choosing 10k rather than, say 5k?

my current goal is to get every metric into the 32ms+ range and kill off clamping (or make it irrelevant). moving everything to 5k accomplishes this quite well, with 10k being prohibitively expensive, especially for future "reverse list" and other "worst case" metrics. would you be opposed to just using 5k across the board and dropping 1k and 10k?

unfortunately, i haven't been able to get `select` slow enough. the easiest way to get this slower is to turn `selected` into a hash and do something like partial update of "select every 10th". the impl changes would be minimal and the conditions can all change from `if (id === store.selected)` to `if (id in store.selected)` or `if (store.selected[id] == null)`.

krausest:
@leeoniya I guess it was just that 10 is such a nice number in comparison to 5. Maybe I was just unfair 😄 
5K would be fine too.
Would select every 10th return a different result as partial update? 

Maybe it would be okay to turn the result of select row into a binary result separating the results into shorter or longer than 2 frames. It‘s then just an indicator whether the framework has an issue with an update in a single row (like choo, halogen, thermite and pux indeed have but all other currently pass)

leeoniya:
> Would select every 10th return a different result as partial update?

partial update mutates textContent/nodeValue while select does setAttribute/className/classList. they fundamentally test different things.

> Maybe it would be okay to turn the result of select row into a binary result separating the results into shorter or longer than 2 frames. It‘s then just an indicator whether the framework has an issue with an update in a single row (like choo, halogen, thermite and pux indeed have but all other currently pass)

this doesnt make much sense to me if we're able to get all the metrics expensive enough to show a reasonable difference. then we can stop fucking around with clamping and per-metic tweaking and just use the raw factors.

it may also be a good opportunity to switch over to `useMinimalCss.css`.

leeoniya:
with the 16x cpu slowdown, all metrics are > 16ms and the clamping code should probably be dropped altogether.

PR #338, opened 2017-12-15
 Use non-default port in circleci

alexfmpe:
#333 didn't replace all hardcoded mentions to port 8080, which caused benchmark `30_startup` and the `nonKeyed` script to fail when using other ports.

Circleci now tests custom port functionality by using port 8081 rather than the default 8080.

### DIFF
--- a/webdriver-ts/src/benchmarks.ts
+++ b/webdriver-ts/src/benchmarks.ts
@@ -24,8 +24,8 @@ export abstract class Benchmark {
         this.label = benchmarkInfo.label;
         this.description = benchmarkInfo.description;
     }
-    abstract init(driver: WebDriver, framework: FrameworkData): Promise<any>;
-    abstract run(driver: WebDriver, framework: FrameworkData): Promise<any>;
+    abstract init(driver: WebDriver, framework: FrameworkData, port: number): Promise<any>;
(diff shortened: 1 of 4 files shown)


webdriver-ts/src/benchmarks.ts
```diff
@@ -380,9 +380,9 @@ class BenchStartup extends Benchmark {
             type: BenchmarkType.STARTUP,
         })
     }
-    async init(driver: WebDriver) { await driver.get(`http://localhost:8080/`); }
-    async run(driver: WebDriver, framework: FrameworkData) {
-        await driver.get(`http://localhost:8080/${framework.uri}/`);
+    async init(driver: WebDriver, framework: FrameworkData, port: number) { await driver.get(`http://localhost:` + port + `/`); }
```
alexfmpe: Had to add the `FrameworkData` parameter to fix an overload incompatibility with the base method.
```
src/benchmarks.ts(374,7): error TS2415: Class 'BenchStartup' incorrectly extends base class 'Benchmark'.
  Types of property 'init' are incompatible.
    Type '(driver: WebDriver, port: number) => Promise<void>' is not assignable to type '(driver: WebDriver, framework: FrameworkData, port: any) => Promise<any>'.
      Types of parameters 'port' and 'framework' are incompatible.
        Type 'FrameworkData' is not assignable to type 'number'.
```

PR #341, opened 2017-12-15
Add maik.h

kaste:
Add library [maik.h](https://github.com/kaste/maik.h)

`maik.h` is a sibling or direct competitor of `lit-html` and `hyperhtml`, and showed surprisingly good results on my computer. Also, it's one of the few libs which supports `Rx` out of the box. 

So I thought a PR would be considered.

### DIFF
--- a/maik-h-v2.5.2-keyed/index.html
+++ b/maik-h-v2.5.2-keyed/index.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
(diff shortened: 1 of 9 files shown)


krausest:
Sorry for the delay -results are [here](https://rawgit.com/krausest/js-framework-benchmark/0893d4d4260950c7473ffbe446063506d547fe57/webdriver-ts-results/table.html).
Please consider the consistently interactive metric from lighthouse as experimental - feel free to ignore it for now.

PR #342, opened 2017-12-16
attodom-v0.4.0-keyed

hville:
Hi Again,
I did not like using pico-dom, finding the abstraction confusing (not good since I wrote it).
`attodom` is faster and smaller still with a minimal api and memory footprint for mobile devices.

I suggest also removing pico-dom while merging attodom since the list resolution algorithms are essentially the same.

### DIFF
--- a/attodom-v0.4.0-keyed/index.html
+++ b/attodom-v0.4.0-keyed/index.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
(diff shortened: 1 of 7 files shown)


krausest:
Sorry, took a bit time...
Results are [here](https://www.rawgit.com/krausest/js-framework-benchmark/0893d4d4260950c7473ffbe446063506d547fe57/webdriver-ts-results/table.html). Quite nice!
pico-dom has been removed.
I'm not sure about the consistently interactive metric from lighthouse - you might ignore it for now.

PR #343, opened 2017-12-19
Add Reagent benchmark

Deraen:
This is the first test of a ClojureScript library and should provide good example for others to use.

This adds benchmark for Reagent, a minimalistic ClojureScript interface for React.js.

~~Build script presumes [Leiningen](https://leiningen.org/) is already installed (it should be available on Circle and Travis). If needed, I could include `lein` script here, or find some solution to download it. I included `lein` script here, so it doesn't have to be downloaded.~~


krausest:
@Deraen The build works seems to work on osx and linux only. Can you please add a build for windows?

Deraen:
@krausest I created lein npm wrapper similar to what binding.scala used for sbt. I hope it will now use `lein.bat` script for Windows.

krausest:
I'll check the windows build the next days. Results are [here](https://rawgit.com/krausest/js-framework-benchmark/37256088d6dac17cc47561502bb849a4dab7706d/webdriver-ts-results/table.html)

Issue #304, opened 2017-11-19
Are non-keyed-only frameworks an anti-pattern?


krausest:
I'm about to advise against using frameworks that support **only** non-keyed updates. 
Integrating 3rd-party libraries, raw dom manipulation or using transitions bears a chance that something doesn't behave correctly and your framework has no support for getting it right. General purpose frameworks should support keyed updates.

What's your opinion?

leeoniya:
sound advice

thysultan:
Loosing state is another major downside of non-keyed updates, mainly in form inputs.

Incorrect animation transitions could equally apply to frameworks that support keyed updates depending on the reconcile heuristics, that is unless the framework in question provides the guarantee of the least number of DOM operations needed to archive a reconciled state.

psirenny:
I personally maintain all of my UI state (animations, deletions, etc.) outside of the framework. Non-keyed is fine for my use case.

leeoniya:
i think the commenting period has officially ended ;)

@krausest close?

krausest:
Hmm - most recent PRs were non-keyed frameworks and this issue isn't good press for them. Thought I'd get more controversy.

PR #345, opened 2017-12-21
update aurelia libraries

gheoan:


krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/0893d4d4260950c7473ffbe446063506d547fe57/webdriver-ts-results/table.html).

Versioning seems a bit confusing for aureliea. Is it correct to take the aurelia-framework version number as the leading version number?

Aurelia appears to support only non-keyed updates. Why is that the case? Do you have any input for #304?

gheoan:
@krausest Good question! All aurelia libraries follow semver, so all aurelia libraries specify their dependencies version using the npm caret notation. I have noticed `aurelia-framework`'s version is bumped only when code inside `aurelia-framework` is changed. Unlike other frameworks not all aurelia libraries have the same version. Call it confusing or call it semver. I can't answer your question without knowing what do you want the leading version number to represent?

I am not familiar with the concept of non-keyed vs keyed so I can't answer that. Do you know any resource that explains the difference between these and the disadvantages mentioned in https://github.com/krausest/js-framework-benchmark/issues/304?

krausest:
A good discussion regarding keyed or non-keyed can be found in https://github.com/krausest/js-framework-benchmark/issues/104#issuecomment-272099815 and #89

PR #346, opened 2017-12-27
Add AppRun

yysun:
[AppRun](https://yysun.github.com/apprun) is a 3K library for developing applications using the elm style model-view-update architecture based on event pub and sub pattern. 

This PR is to benchmark AppRun and compare with other frameworks and libraries.


krausest:
Thanks. According to our definitions the keyed version isn't keyed for create rows (but for swap rows and remove row). Seems like it reuses the exisiting dom nodes. Can you change that behaviour? Otherwise I suggest only adding the non-keyed version.
```
Stefans-MBP:webdriver-ts stefan$ npm run nonKeyed -- --framework apprun-v1.7.0-keyed

> webdriver-ts@1.0.0 nonKeyed /Users/stefan/Source/Javascript/js-framework-benchmark/webdriver-ts
> node dist/nonKeyed.js "--framework" "apprun-v1.7.0-keyed"

Frameworks that will be checked [ 'apprun-v1.7.0-keyed' ]
apprun-v1.7.0-keyed is non-keyed for 'run benchmark' and keyed for 'remove row benchmark' keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
ERROR: Framework apprun-v1.7.0-keyed is not correctly categorized in commons.ts
```

yysun:
Thank you! 
Let's add non-keyed for now. Will send another PR later for keyed.

yysun:
After fixed the library, I think the keyed behaviour is corrected.

krausest:
Keyed is fine now. I'll merge and report the results in a few minutes.

Just small a notice - doesn't really matter: Some non-keyed frameworks handle remove row differently. They remove the last row and shift the content one row up which is faster in this particular benchmark (I guess due to the striped table rows). 
```
apprun-v1.7.0-non-keyed is non-keyed for 'run benchmark' and keyed for 'remove row benchmark' non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
```

krausest:
I've merged it but noticed a small error afterwards. Can you please take a look at it? 
If you click create 1,000 rows twice then the first row looks wrong because it's wrapped inside a separate tbody element.

krausest:
Here's a screenshot of the error:
bildschirmfoto 2018-01-01 um 20 30 25

yysun:
Thank you again for helping on this.
I have update the library from 1.7.8 to 1.7.9 that fixes the issue of tbody.

PR #347, opened 2018-01-06
angular-light-v0.14.1-keyed, rivets-v0.9.6-non-keyed

marc3lsk:
Derived from knockout implementation.

Angular Light is generally fast, but there was dramatic slowdown when I created my own directive for al-text, based on example here http://angular-light.readthedocs.io/en/latest/create_directive.html

Rivets is slow to create initial rows, but fast to recreate and update them. Sadly, there is very noticable slowdown with every added binding.


krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/859146793015d36a23f5f24056e393a081fe2b99/webdriver-ts-results/table.html) .
Does rivets support only non-keyed updates?

marc3lsk:
I could not find any documentation related to keyed / non-keyed updates.

PR #348, opened 2018-01-09
Ember updates

mwpastore:


krausest:
Thanks - and shame on me. 

I must add a README to ember otherwise I'll mess up updating the next time again. Can you please help me set it straight?
1. Update only ember-cli in package.json to latest version.
2. rm -r package-lock.json node_modules/
3. npm install
4. ember init [this updates package.json again]
5. ember-modules-codemod

Is step 5 necessary in future?

[Results](https://rawgit.com/krausest/js-framework-benchmark/859146793015d36a23f5f24056e393a081fe2b99/webdriver-ts-results/table.html) are actually slower than the last run. Is that expected?

mwpastore:
You're welcome! No shame necessary; it's impossible to keep up with the ins-and-outs of every framework. You do a remarkable job! 

Unfortunately, the instructions can change from release to release. They're always published in the release notes for a given ember-cli version. However, most of the time the upgrade process follows roughly the same steps, and you're quite close. Here are the instructions for [2.16.2](https://github.com/ember-cli/ember-cli/releases/tag/v2.16.2), for example. 

> Is step 5 necessary in future?

It may be necessary to run other codemods in the future if there are big changes that have to be applied to non-blueprint files, but it's not an every-time thing. This is the first and only time it's been necessary so far. 

> Results are actually slower than the last run. Is that expected?

I saw the same in my local benchmarking as well. It's not expected, but it is what it is. I've been playing with the `rows` service and `my-table` component, trying to make them more Ember-ish and Glimmer-friendly, but I haven't been able to close the performance gap yet.

krausest:
Thanks - lets see if 1d0db20d63bb4219e4c7d6b36fc30fb7abd17378 helps :smile:

PR #349, opened 2018-01-14
Added Inferno v4 non-keyed & keyed variations.

Havunen:
I Added inferno v4 implementation and tried to explain how JSX flags should be used in v4 forwards. Its no longer needed to have `noNormalize` everywhere. Also for the sake of code examples I created other benchmark using rollup (keyed) and other one using webpack (non keyed).


krausest:
Thanks - and quite [impressive](https://rawgit.com/krausest/js-framework-benchmark/859146793015d36a23f5f24056e393a081fe2b99/webdriver-ts-results/table.html).
I'd remove 3.10 when 4.0.0 is official?

thysultan:
Would be great to have a inferno-compat version for comparison... and maybe also a preact-compat version, with a 1-to-1 alias of the react implementation.

Issue #303, opened 2017-11-18
Chromedriver crashes


krausest:
Tried at least 4 times, but test doesn't complete for all frameworks. Errors are like that:
```
(node:25652) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 exit listeners added. Use emitter.setMaxListeners() to increase limit
Benchmark failed { TimeoutError: Waiting testTextContains //tbody/tr[1000]/td[1] 2000
Wait timed out after 60050ms
    at /Users/stefan/Source/Javascript/js-framework-benchmark/webdriver-ts/node_modules/selenium-webdriver/lib/promise.js:2201:17
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7) name: 'TimeoutError', remoteStacktrace: '' }
Writing screenshot error-halogen-v2.1.0-non-keyed-25_run-clear-memory.png
Benchmark failed { TimeoutError: Waiting testElementLocatedById add
Wait timed out after 20091ms
    at /Users/stefan/Source/Javascript/js-framework-benchmark/webdriver-ts/node_modules/selenium-webdriver/lib/promise.js:2201:17
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7) name: 'TimeoutError', remoteStacktrace: '' }
Writing screenshot error-ivi-v0.9.1-keyed-24_run5-memory.png
Benchmark failed { WebDriverError: unknown error: unable to discover open pages
  (Driver info: chromedriver=2.33.506106 (8a06c39c4582fbfbab6966dbb1c38a9173bfb1a2),platform=Mac OS X 10.13.1 x86_64)
    at Object.checkLegacyResponse (/Users/stefan/Source/Javascript/js-framework-benchmark/webdriver-ts/node_modules/selenium-webdriver/lib/error.js:546:15)
    at parseHttpResponse (/Users/stefan/Source/Javascript/js-framework-benchmark/webdriver-ts/node_modules/selenium-webdriver/lib/http.js:509:13)
    at doSend.then.response (/Users/stefan/Source/Javascript/js-framework-benchmark/webdriver-ts/node_modules/selenium-webdriver/lib/http.js:441:30)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7) name: 'WebDriverError', remoteStacktrace: '' }
[...]

PR #351, opened 2018-01-18
update nerv

yuche:
Update Nerv to 1.2.5 since 1.2.4 has a bug that patch is possibly go wrong when lastChildren is empty.

And sadly, I haven't run the benchmark yet since the chromeDriver issue #303 , It crash every time in when I run `npm run selenium -- --count 3 --framework super-vanillajs-keyed`:

```
Benchmark failed Error: Server terminated early with status 1
    at earlyTermination.catch.e (/Users/yuche/Developer/js-framework-benchmark/webdriver-ts/node_modules/selenium-webdriver/remote/index.js:252:52)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
ERROR: Error: Server terminated early with status 1
    at earlyTermination.catch.e (/Users/yuche/Developer/js-framework-benchmark/webdriver-ts/node_modules/selenium-webdriver/remote/index.js:252:52)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
QUIT
(node:60377) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Server terminated early with status 1
(node:60377) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:60377) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): Error: Server terminated early with status 1
```

This benchmark can go though in a version of 300 commit behind the current master, hope it can work in your machine.


nervjs-v1.2.5-keyed/src/Store.es6.js
```diff
@@ -0,0 +1,59 @@
...
+        this.data = this.buildData(10000);
+        this.selected = undefined;
+    }
+    clear() {
+        this.data = [];
+        this.selected = undefined;
+    }
+    swapRows() {
+    	if(this.data.length > 10) {
+    		var a = this.data[4];
```
thysultan: These should be `1` and `998` like in [v1.2.4/utils.es6.js](https://github.com/krausest/js-framework-benchmark/blob/a49ca102504abc53ce3c9454479513cc381e5b53/nervjs-v1.2.4-keyed/src/utils.es6.js#L37-L45)

yuche: Thanks for point this out. Is it ok now?

thysultan: Looks ok.

thysultan:
Is there a reason the implementation is moving away from being identical to the React implementation?

yuche:
It's a little bit fast, and [Inferno](https://github.com/krausest/js-framework-benchmark/blob/master/inferno-v3.10.1-keyed/src/controller.jsx) do that too.

thysultan:
That is true, i am also guilty of this with previous implementations of DIO, but ideally it would be nice for React-like libs to share the same implementation as React(or submit a version that does) so as to not have to factor the specifics of a benchmark implementation when comparing the different React-like libs to React and each other.

Maybe @Havunen can weight in on whether we can get Inferno to also get behind this convention.

Havunen:
Yeah IMO the code should be same, I just noticed surplus and ivi implemented their own also. And React16 was also different to Inferno v3.

krausest:
New version looks not good to me.
Open page, click create 1,000 rows, click swap rows and the select the first row. This is what happens:
bildschirmfoto 2018-01-23 um 20 56 45
Can you check that please?

krausest:
nerv 1.2.7 seems to work. I'll update to 1.2.7 then the next few days.

yuche:
Just update the PR to 1.2.8. I guess it should work now.

krausest:
That works - results are [there](https://www.rawgit.com/krausest/js-framework-benchmark/2856850a81723c12d6523b931aba1b8178621a80/webdriver-ts-results/table.html)

PR #353, opened 2018-01-23
Update RE:DOM to v3.10.1

pakastin:
https://redom.js.org


krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/1fdeb6c2786be863ea8c347a3d1ab17ed1c4e23a/webdriver-ts-results/table.html).

pakastin:
Noooooooo..that swap time 😬

Well, thanks anyway – I'll continue tinkering here 😎

PR #358, opened 2018-02-16
Ember: Upgrade to 3.0.0

mwpastore:


krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/943a2ae98e1efae749a042f5bc73f908f25a4f39/webdriver-ts-results/table.html). Seemingly not really an improvement. Did you expect to see better results?
I also updated vanillajs-keyed results to see if the new chrome version caused a spectre related performance regression, but couldn't confirm that.

mwpastore:
> Did you expect to see better results?

No, but this is the latest version of Ember, and it feels disingenuous to cherry-pick an older, faster version to represent Ember in this benchmark.

PR #359, opened 2018-02-16
Update react-easy-state to v5.0.0

solkimicreb:
Title tells it all 🙂


krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/943a2ae98e1efae749a042f5bc73f908f25a4f39/webdriver-ts-results/table.html). Seems a bit faster than 4.0.1.

PR #360, opened 2018-02-19
Update Glimmer to 0.9.1

chadhietala:
I couldn't run the webdriver to get the results but it does render when you run the benchmark by it's self.


krausest:
Thanks - results are [here](https://rawgit.com/krausest/js-framework-benchmark/943a2ae98e1efae749a042f5bc73f908f25a4f39/webdriver-ts-results/table.html). 0.9.1 seems to be a lot slower than 0.8.0 (and ember 3.0.0). Did you expect that?
I also updated vanillajs-keyed results to see if the new chrome version caused a spectre related performance regression, but couldn't confirm that.

krausest:
What was the problem with webdriver on your machine?

chadhietala:
Nope. The applications we have shipped with 0.9.1 have all been faster based on Real User Metrics.

The webdriver issue was that chrome could not connect to the renderer.

chadhietala:
It's also a bit suspicious because I ran these when I was testing things out and things seemed to improve quite a bit.
screen shot 2018-02-19 at 4 18 14 pm

krausest:
The numbers on the console are just an approximation. The actual measurement is taken from the timeline. I checked some results from the timeline to see if the test driver results are believable, e.g. update every 10th row for glimmer:
bildschirmfoto 2018-02-22 um 21 37 06
and for preact:
bildschirmfoto 2018-02-22 um 21 37 45

Similar for create 1000 rows. For glimmer:
bildschirmfoto 2018-02-22 um 21 40 36
and for preact:
bildschirmfoto 2018-02-22 um 21 40 52

Here are two screenshots from some interactions with the approximated duration from the console. For glimmer:
bildschirmfoto 2018-02-22 um 21 44 35

For preact:
bildschirmfoto 2018-02-22 um 21 44 46

Seems like performance characteristics on my machine is pretty different from yours, but I don't see any evidence that the numbers in the result table are incorrect.

chadhietala:
I think we have identified a regression. I'll update again once we get a fix out.

NullVoxPopuli:
@chadhietala do you find out what the issue was?

PR #362, opened 2018-03-01
Update lit-html to 0.9.0 and add keyed version

CaptainCodeman:


krausest:
Thanks, results are [here](https://rawgit.com/krausest/js-framework-benchmark/d38c5b386fb3825ed9c7f019a2c1c5e4421936ba/webdriver-ts-results/table.html).
Unfortunately it seems like 0.9.0 is significantly slower than 0.7.1.

CaptainCodeman:
Thanks, yes - I'll check I haven't messed anything up but otherwise feed that back and it'll maybe help focus some optimization efforts. I know the keyed version still needs some work and isn't considered 'ready' but surprised to see the non-keyed slow down.

PR #365, opened 2018-03-09
Update dojo versions

agubler:
Update to the latest versions of dojo 2 packages. Should see an improvement in the swap benchmark with an improved children diffing algorithm.

interactive_results


krausest:
Thanks - results are [here](https://www.rawgit.com/krausest/js-framework-benchmark/802a2ccf15857ef89ea0f5f7be6b88afab72b8fa/webdriver-ts-results/table.html)
As you expected swap rows is much better now.

agubler:
Thanks @krausest!

PR #367, opened 2018-03-19
update Svelte to 1.58

Rich-Harris:
hey — this updates the Svelte benchmarks to version 1.58, which fixes our abysmal row swapping performance. Thanks!


krausest:
Thanks - but I'm afraid the keyed version looks horribly wrong:
https://www.rawgit.com/krausest/js-framework-benchmark/802a2ccf15857ef89ea0f5f7be6b88afab72b8fa/webdriver-ts-results/table.html
bildschirmfoto 2018-03-19 um 22 00 41

I've checked in the chrome timeline and can confirm that the measurement e.g. for select row seems correct:
bildschirmfoto 2018-03-19 um 22 01 13
The non-keyed version is also a bit slower, but much closer:
bildschirmfoto 2018-03-19 um 22 03 58
Can you please check what's wrong with it?

Rich-Harris:
yikes! not sure what happened there. This is what it looked like shortly before I cut 1.58: http://windy-able.surge.sh/

screen shot 2018-03-19 at 5 16 30 pm

I made some further changes, but only to tidy things up slightly (or so I thought) — I most have broken something badly. Will investigate. Thanks for the heads up!

krausest:
I can confirm that if I change the version to 1.57.4 I'm seeing good performance. This is select row for 1.57.4:

bildschirmfoto 2018-03-19 um 22 41 44

And when I switch back to 1.58.0 I get:
bildschirmfoto 2018-03-19 um 22 44 02

PR #369, opened 2018-03-26
update svelte to 1.58.5

Rich-Harris:
This fixes the recent performance regression (#367) with a [new list diffing algorithm](https://github.com/sveltejs/svelte/pull/1274). Should be much quicker now! Thanks


krausest:
That looks much better!
bildschirmfoto 2018-03-30 um 18 55 22
The red benchmarks are significantly slower than 1.42. Is that expected (might also be due to Chrome and OS updates)?
bildschirmfoto 2018-03-30 um 18 57 09

krausest:
Complete results are available [here](https://www.rawgit.com/krausest/js-framework-benchmark/50c288df115b3fb0cb576a1cc3c04e77a5f87150/webdriver-ts-results/table.html).

Rich-Harris:
Nice, thanks! Yes, our row creation needs to get faster. Not sure off the top of my head what explains the regression, but getting that down is very much on our TODO list.

In the meantime, I'll report this back to our chatroom so we can luxuriate in our number one spot for TTI, bootup time, memory usage etc 😀

PR #373, opened 2018-04-05
Add an optimized Angular benchmark

trotyl:
Using a custom differ in Angular (which inspired by inferno).

Other:

+ Fixed static dep version;
+ Removed the extra change detection;


krausest:
Thanks. The optimized version looks impressive.
bildschirmfoto 2018-04-09 um 20 48 44

PR #379, opened 2018-04-13
update san to 3.5.5

errorrik:
this updates the San benchmarks to version 3.5.5, which we see the big performance improvement. Thanks!


krausest:
Thanks. I messed up my results folder so I can only publish a few screenshots:

bildschirmfoto 2018-04-16 um 20 40 00
bildschirmfoto 2018-04-16 um 20 40 11

errorrik:
:D

PR #380, opened 2018-04-13
ractive version bumps and small test adjustments

evs-chris:
~~When trying to profile a bit for a new ractive release, I found that the chrome driver on windows will not completely close chrome unless you `close` the instance before `quit`ting. That results in pretty terrible performance after running a few of the benchmarks because zombie chromium instances start stacking up pretty fast. Linux doesn't seem to have that issue, and I'd assume OS X doesn't either. I hope it's ok to add that in here.~~

This bumps the Ractive 0.9 version to 0.9.13 from 0.9.9, adjusts the partial update test to be a bit more efficient, and switches to pre-parsed templates to drop the startup costs a bit.

This also bumps Ractive edge to 0.10.1, adjust the partial update test to be a bit more efficient, and switches to pre-parsed templates, again, to drop the startup costs a bit.

Completely unrelated side question: has anyone running Win 10 or Linux noticed wildly varying benchmark times on an otherwise idle system? On Windows I get it, because the anti-malware gubbins kick in at random and decide to eat CPU like mad, but the Linux variances are a little more surprising. Does OS X have the same issue?


leeoniya:
it's been a while since i ran the benches (on my Win10), but i did not notice anything unusual at the time.

> I hope it's ok to add that in here.

probably best to split this PR into two

evs-chris:
> probably best to split this PR into two

Done. Thanks!

krausest:
@evs-chris (I'll merge soon - promised)
I'm currently working on the windows version. Last night I ran all benchmarks except startup and got the following unredacted results: https://gistpreview.github.io/?a0d5f2c4ac4a7e772776f9b7897958b5
I haven't checked each and every number, but std deviation seems to be rather similar to OSX. On OSX I sometimes have varying results e.g. when spotlight starts indexing. 
What results do you get when you have "wildly varying" results? Can you post an example (duration and standard deviation)?

evs-chris:
I don't have any hard numbers available right now, but running a full set of benchmarks with count 10 across 4 or 5 frameworks in one go will result in up to 100% variation in means in either direction for some of the chewier benchmarks. Even running a specific benchmark by itself seems to vary a good bit from run to run. For instance, for the vanilla 10k keyed run, my windows machine typically runs out in about 1300ms, but sometimes it will hit up to 2800ms even across a 10 count sample. My Linux machine is a passively cooled laptop, so I chalk varied longer runs up to thermal throttling. I still get fairly varied numbers for a single run on any of the chewier benchmarks.

Totally not suggesting that there's anything wrong with the tests. I just found it odd that trying to nail down a baseline to see if framework changes make any difference is as tricky as it is. There's a lot of gut feeling involved in determining whether a run is typical, happened to be really good, or happened to be really bad for whatever reason. It may be that I just don't have a machine that's capable of truly consistent sustained performance. My system with a decent os lacks a beefy processor, and my system with a beefy processor lacks a decent os 😁

If you want, I can see if I can get some numbers on Monday. It sounds like os x with active cooling is a good deal more reliable, so as far as I'm concerned, that question is answered.

krausest:
Thanks. I messed up my results folder so I can only publish a few screenshots:

bildschirmfoto 2018-04-16 um 20 39 30
bildschirmfoto 2018-04-16 um 20 39 38

bildschirmfoto 2018-04-16 um 20 40 00
bildschirmfoto 2018-04-16 um 20 40 11

Issue #378, opened 2018-04-12
CPU usage rises for test runner on windows


krausest:
The CPU consumption of the testrunner rises to about 100% after a short time on windows (does not happen on OSX).
Removing the --no-sandbox option in benchmarkRunner.ts removes that effect.
Needs to be further analyzed.

Alexander-Taran:
reverting my remark on adding .close(). I missed one spot. It does work for me now.

krausest:
As an additional measure I'm now forking the chromedriver process.

PR #381, opened 2018-04-13
close before quit to avoid zombie instances on windows

evs-chris:
See #378. It seems that chrome driver doesn't kill off all sub-processes on windows if you call `quit` when there are still open windows. The result being that after a few benchmarks are run, the system starts bogging down terribly. This seems to work on my machine, but there is a comment in #378 suggesting that it may not cover all of the corners 100%.


Alexander-Taran:
yep.. that's exactly what I tried locally with no luck..
but I have chrome dev channel 32bit and chromedriver 2.37
so that could affect the outcome

evs-chris:
I'm on 64-bit stable channel (65) with chromedriver 2.36. I can't imagine the arch making much difference, but it wouldn't surprise me if dev channel made things go sideways.

Alexander-Taran:
Sorry for misinformation. I missed one .quit. 
added a .close before it and now it is fine for me.

krausest:
Thanks!

PR #387, opened 2018-04-24
Add svelte-2.1.1 example

funkybob:


krausest:
Thanks for your contribution, but I'm afraid can't merge it in the current state:
* doesn't build with npm run build-prod
* generated html doesn't match that of the other frameworks (e.g. name and remove links must be links)
* row selection doesn't work

PR #390, opened 2018-05-07
Update ivi 0.10.0

localvoid:


krausest:
Thanks! Results are [here](https://rawgit.com/krausest/js-framework-benchmark/ada0ff20491b8541bd2bd428221f3c6ff5148a7f/webdriver-ts-results/table.html) and here's the relevant screenshot. I'm afraid there's not much room for improvement :smile:

localvoid:
I am actually almost completely changed internal data structures for a virtual dom in this update, instead of storing children nodes in arrays, they are now linked in a circular linked list. I've tried this idea like a year ago, but it was slower because of huge complexity, and now I just removed some of the "must have" optimizations and everything became so simple, so that even without such optimizations it performs incredibly well, and also minified bundle should be ~5kb smaller :)

krausest:
You're right - I just looked at the naked runtime performance...
bildschirmfoto 2018-05-10 um 11 16 12

PR #391, opened 2018-05-07
Update lit-html to 0.10.0

CaptainCodeman:
Latest 0.10.0 version of Lit-Html


krausest:
Thanks! Results are [here](https://rawgit.com/krausest/js-framework-benchmark/ada0ff20491b8541bd2bd428221f3c6ff5148a7f/webdriver-ts-results/table.html) and here's the relevant screenshot. Partial update and append rows got significantly faster:
And for non-keyed it's here:
bildschirmfoto 2018-05-10 um 10 09 50

CaptainCodeman:
Great, thanks!

PR #392, opened 2018-05-08
Add ko-jsx

ryansolid:
[https://github.com/ryansolid/ko-jsx](https://github.com/ryansolid/ko-jsx)

An alternate renderer for KnockoutJS that uses precompiled JSX to real DOM nodes with performance in mind. I hope will consider adding it to the benchmark suite.

Thanks.


krausest:
Very nice! Results are [here](https://rawgit.com/krausest/js-framework-benchmark/ada0ff20491b8541bd2bd428221f3c6ff5148a7f/webdriver-ts-results/table.html) and here's the relevant screenshot. A lot faster than the default renderer!

PR #393, opened 2018-05-11
add redux-combiner

rmdm:
[redux-combiner](https://github.com/rmdm/redux-combiner) proposes convenient way to describe possible application state changes. It should also be quite performant.


krausest:
Thanks - full results are [here](https://rawgit.com/krausest/js-framework-benchmark/a08b22aa94a1a0bd8df833c06ed1cf20dca0977e/webdriver-ts-results/table.html)
bildschirmfoto 2018-05-13 um 20 40 23
It came out a bit faster than redux, but not statistically significant (even not for create rows):
bildschirmfoto 2018-05-13 um 20 58 01

PR #396, opened 2018-05-19
domvm v3.3.2

leeoniya:
hey @krausest,

sorry for the rapid update, but this will be the last one for a while.

i simplified the benchmark implementation, removing a couple micro-optimizations (event delegation, jumbotron redraw skipping). the new impl is also less monolithic and more functional/componentized. v3.3.2 also reworks some event listener internals to use addEventListener rather than binding directly to the dom nodes.

in my tests, the perf was within +/-2% and +0.2M in the memory metrics. well worth the clearer code, imo.

thanks!


krausest:
Thanks. Perf results look like that:
bildschirmfoto 2018-05-21 um 21 22 05

bildschirmfoto 2018-05-21 um 21 21 53

leeoniya:
these numbers make sense.

the impl is much clearer, so i'm happy about that. in the future i'll see if there's any room to get some of the perf back from internal improvements.

thanks!

PR #397, opened 2018-05-20
Move frameworks into subfolder

alexfmpe:
This is the first of a series of changes intended to allow a tree-like structure for frameworks, rather than a flat top-level one.


build.js
```diff
@@ -3,15 +3,15 @@ var exec = require('child_process').execSync;
...
 
-var excludedDirectories = ['css', 'images', 'dist','node_modules','webdriver-java'];
-
 // set the following variable to resume building with a framework and skip all
 // other frameworks that would be buils before
 var restartWithFramework = '';
 var build = !restartWithFramework ? true : false;
 
-_.each(fs.readdirSync('.'), function(name) {
-	if(!name.startsWith(".") && fs.statSync(name).isDirectory() && excludedDirectories.indexOf(name)==-1 && fs.existsSync(path.join(name, "package.json"))) {
```
alexfmpe: [`readdirSync`](https://nodejs.org/docs/latest-v8.x/api/fs.html#fs_fs_readdirsync_path_options) excludes the `.` entry

krausest:
@alexfmpe I definitely need a bit more information about _why_ I should restructure the project. What's the goal of these PR?

I'm also thinking about simplifying maintenance, but I currently consider dropping the version suffix from all frameworks. If more than one version should be included the one without a suffix should be the stable version and the others should carry a self explaining version suffix.

PR #399, opened 2018-05-29
Update ivi to 0.13.0

localvoid:
- google closure compiler replaced with simple rollup+uglify build
- async rAF update is changed to sync


krausest:
[Nice](https://www.rawgit.com/krausest/js-framework-benchmark/cc9e81c51174eb957b4d213661caa1945a4fb8a7/webdriver-ts-results/table.html) - a bit faster overall despite taking a hit with the rAF change.

PR #402, opened 2018-05-29
Remove webcomponents polyfill (lit-html)

### LINKED ISSUE #401: Building lit-html fails
@CaptainCodeman Installing lit-html fails (dependency git://github.com/webcomponents/webcomponentsjs.git#no-html-imports) - can you please take a look at it?
```
pm WARN deprecated babel-preset-es2016@6.24.1: We're super 😸  excited that you're trying to use ES2016 syntax, but instead of making more yearly presets 😭 , Babel now has a better preset that we recommend you use instead: npm install babel-preset-env --save-dev. preset-env without options will compile ES2015+ down to ES5 just like using all the presets together and thus is more future proof. It also allows you to target specific browsers so that Babel can do less work and you can ship native ES2015+ to user 😎 ! We are also in the process of releasing v7, so please give http://babeljs.io/blog/2017/09/12/planning-for-7.0 a read and help test it out in beta! Thanks so much for using Babel 🙏, please give us a follow on Twitter @babeljs for news on Babel, join slack.babeljs.io for discussion/development and help support the project at opencollective.com/babel
npm ERR! code 1
npm ERR! Command failed: /usr/local/bin/git checkout no-html-imports
npm ERR! error: pathspec 'no-html-imports' did not match any file(s) known to git.
npm ERR!
```

CaptainCodeman:
Closes #401


krausest:
Thanks for the quick PR.

Issue #403, opened 2018-05-30
make vanillajs great again


leeoniya:
1.00 (besides rAF clearing)

ryansolid:
I agree. Not only having a clear target is good. It helps library writers and test implementers see at the core level the techniques to best deal with the DOM. Admittedly some of the stuff is Chrome specific (and I'm not just talking about rAF), as each browser performs differently, but realistically all you can do is optimize for the platforms the tests run on. 

One thing I will mention is that I develop on a lower power machine than the benchmarks run on and VanillaJS is consistently better than any of the frameworks. Even more noticeable when the cpu is being throttled in power saving modes. The more powerful the computer I've found the closer to Vanilla the libraries get. But I guess I wonder if the frameworks are getting so optimized it's the test themselves that need looking at again.

Looking purely at the numbers I'm guessing partialUpdate, replaceRows and swapRows are the areas Vanilla can improve the most.  Surplus has partialUpdate pretty convincingly, ivi and inferno seem to have an advantage on replaceRows and attodom has swapRows dialed. I don't have anything to help swapRows but perhaps there are a few optimizations I can suggest to help the others.

I did come across a few things when working on Solid that I either JSPerf or found other libraries doing that seemed to help that VanillaJS doesn't do. Keep in mind I tested these independently and didn't do enough comparison tests in the context of the benchmark to conclusively say in all cases these are better.

* buildData is faster with a pre-allocated array (ie. new Array(count)) and setting index rather than push items on. 
* using textContent instead of innerText generally is less expensive since it doesn't need to look at the layout to determine how to render.
* When updating existing text nodes you get better performance out of setting the data property (ie. element.firstChild.data = ...) of it than resetting the textContent of the parent

I hope that helps. (But maybe these haven't been implemented already for a reason).

leeoniya:
> But I guess I wonder if the frameworks are getting so optimized it's the test themselves that need looking at again.

i've wanted to toy with the idea of measuring JS execution time only and throwing away layout and paint times. this has the effect of showing a much wider, more realistic gulf than the current numbers show (where most of the time is dominated by layout).

> The more powerful the computer I've found the closer to Vanilla the libraries get.

i've noticed this too, working on a slower i5 lenovo thinkpad with HD4000 and Windows 10 x64.

at the end of the day, my goal is to have a single target to build, bench and compare against instead of spending a bunch of time and resources assembling the fastest metrics.

> I hope that helps

[...]

PR #404, opened 2018-06-01
Vanillajs Great Again

ryansolid:
Honestly not sure, how much of a difference this will make but was prompted by [this thread](https://github.com/krausest/js-framework-benchmark/issues/403).  In my search of the best optimization techniques most of the top libraries did these things and I verified them in isolation to the best of my ability. Perhaps this will help VanillaJS perform a tiny bit better.

### DIFF
--- a/frameworks/vanillajs-keyed/src/Main.js
+++ b/frameworks/vanillajs-keyed/src/Main.js
@@ -33,9 +33,9 @@ class Store {
         var adjectives = ["pretty", "large", "big", "small", "tall", "short", "long", "handsome", "plain", "quaint", "clean", "elegant", "easy", "angry", "crazy", "helpful", "mushy", "odd", "unsightly", "adorable", "important", "inexpensive", "cheap", "expensive", "fancy"];
         var colours = ["red", "yellow", "blue", "green", "pink", "brown", "purple", "brown", "white", "black", "orange"];
         var nouns = ["table", "chair", "house", "bbq", "desk", "car", "pony", "cookie", "sandwich", "burger", "pizza", "mouse", "keyboard"];
-        var data = [];
+        var data = new Array(count);
         for (var i = 0; i < count; i++)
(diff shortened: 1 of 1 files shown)


kurtextrem:
Are you sure the Array change will improve perf?
arrays1
arrays2

krausest:
First of all - thanks for the PR.

I benchmarked your PR. Here are the results:
bildschirmfoto 2018-06-03 um 22 20 49

* vanillajs-keyed is the old version
* vanillajs-keyed-new is your PR
* vanillajs-keyed-array is vanillajs-keyed with the preallocated array from your PR
* vanillajs-keyed-textcontent is all dom modifications from your PT, but not the preallocated array 

The array and the dom modifications influence create rows, replace all rows and partial update. But it seems like the noise in the measurements is bigger than the improvement.
So I'm afraid currently I'd rather not merge this PR. Do you have further ideas?

ryansolid:
@kurtextrem That's exactly it. It's clearly faster to pre-allocate for create when benchmarked separately but it's hard to make any conclusion when put all together. We're getting to the point small edges aren't even really edges.

@krausest Yeah, nothing significant here. I'm not that surprised, which is why I was a bit hesitant to make the PR. TextContent in theory should be a pretty good change in the general case, and something I'd do as a library writer. Setting the element data for existing nodes on update instead of the innerText/textContent is also along this vein, which should only have noticeable effect on Partial Update really I imagine.

I only really stumbled across the Array thing when trying to find any little thing that could make the tests run better. And it seemed to on my lower power device, but it wasn't nearly as significant as other changes I made, and after those other changes I'm not sure it's actually significant at all.

Really the last thing I'm looking at is rAF. And I don't mean in the specific clear row scenario. I think it also benefits replace rows, and has minimal negative impact on on create rows. It's possible there is a more general rule for where this is beneficial. But for now I have nothing else. I think the top libraries practically generate the exact same executable JS when it comes down to the DOM manipulation. It's hard to make any definitive improvements on that side.

Issue #405, opened 2018-06-03
OSX: Chrome 67 + webdriver render textless page


krausest:
Seems like chromedriver 2.38.3 and chrome 67 don't like one another (once again).
All frameworks show empty text:
<img width="1146" alt="bildschirmfoto 2018-06-03 um 21 04 28" src="https://user-images.githubusercontent.com/3374055/40890161-c7848318-6771-11e8-9263-1d87cb4c4c3e.png">
Downgrading to chromium 66 helps.

krausest:
Created an issue for chromedriver: https://bugs.chromium.org/p/chromedriver/issues/detail?id=2462

krausest:
Since chrome 67 I must add LANG="en_US.UTF-8" or it won't render correctly. Was added in package.json.

PR #406, opened 2018-06-09
Add WebComponents flavoured VanillaJS implementation

localvoid:
This implementation should show performance overhead of WebComponents. It is implemented in imperative VanillaJS style.

Works in Chrome >= 67 ([Customized built-in elements](https://www.chromestatus.com/feature/4670146924773376))

### DIFF
--- a/frameworks/vanillajs-wc-keyed/index.html
+++ b/frameworks/vanillajs-wc-keyed/index.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8" />
+  <title>VanillaJS-WebComponents-"keyed"</title>
+  <link href="/css/currentStyle.css" rel="stylesheet" />
+</head>
+
+<body>
+  <script type="module" src="src/main.js"></script>
+  <div is="benchmark-app" />
+</body>
+
+</html>
--- a/frameworks/vanillajs-wc-keyed/package.json
+++ b/frameworks/vanillajs-wc-keyed/package.json
@@ -1,0 +1,11 @@
+{
+  "private": true,
+  "name": "js-framework-benchmark-vanillajs-wc",
+  "version": "1.0.0",
+  "description": "Vanilla.JS WebComponent demo",
+  "scripts": {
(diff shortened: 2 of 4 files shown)


krausest:
Thanks. But I'm afraid I've no results today due to #405
Currently only Chrome <= 66 works on my macbook, chrome 67 with chromedriver renders but shows no text.

krausest:
Thanks. Now that I found that I just have to set the LANG variable I was able to run a few frameworks  for chrome 67:
bildschirmfoto 2018-06-13 um 20 38 01
@ryansolid @leeoniya @adamhaile 
Seems like chrome 67 is much faster for clear rows and using RAF isn't giving an advantage anymore (hurray for ending this discussion :sweat_smile:).
I'll try to re-run all frameworks the next days.

localvoid:
Thanks, this rAF hack is bizarre, I think that it is still can give an advantage, so it is better if everyone just removes it. Here is a run on my machine (Linux, Chrome 67) with some modifications to benchmark implementations.

benchmark

**surplus-EVENTS** doesn't have rAF hack and has a noticeable difference in clear rows.

Modifications:

- [ivi-SIMPLE](https://github.com/localvoid/js-framework-benchmark/blob/sandbox/frameworks/ivi-SIMPLE/src/main.js) - Event delegation, removed connectors, removed stateful components
- [surplus-EVENTS](https://github.com/localvoid/js-framework-benchmark/blob/sandbox/frameworks/surplus-EVENTS/src/view.tsx) - Removed event delegation, removed rAF hack
- [Inferno-SC](https://github.com/localvoid/js-framework-benchmark/blob/sandbox/frameworks/inferno-SC/src/controller.jsx) - Each row is a stateful component

https://rawgit.com/localvoid/js-framework-benchmark/sandbox/webdriver-ts-results/table.html

ryansolid:
Yeah It's better but it's still there. I think it's just that the more powerful the computer the smaller the difference gets. I am seeing about a 10ms on my machine. It was like 35ms before though.. I think on the machine that run the benchmark it accounts for about a 5ms difference.  Getting to the point it doesn't particularly matter though. But as indicated above all those libraries are so close.

PR #407, opened 2018-06-09
Update marionette to v4

paulfalgout:
Marionette the library that ["may even justify a migration"](https://www.stefankrause.net/wp/?p=431#post-431) is nearing the release of v4.

This PR updates the two implementations.  As before, one that is Marionette using what comes with backbone out of the box, jQuery for DOM manipulation and underscore for templating.  And another that uses more optimized settings utilizing a DOM renderer and morphdom for DOM patching.

With jQuery we're in the midst of Ember or glimmer and with morphdom we're ahead of most of the major libraries.  In both cases we're also one of the smaller memory footprints considering Marionette is much more than a simple view library.

Thanks for this benchmark, it's helped us fine tune our 6 year old library. 

screen shot 2018-06-09 at 10 56 50 pm
screen shot 2018-06-09 at 10 57 11 pm


krausest:
@paulfalgout I've got a problem building marionette-v4.0.0-keyed. npm run build-prod reports
```
> js-framework-benchmark-marionette@1.0.0 build-prod /Users/stefan/Source/Javascript/js-framework-benchmark/frameworks/marionette-v4.0.0-keyed
> webpack -p

Hash: daf94dca5a49df7fe0f9
Version: webpack 3.8.1
Child
    Hash: daf94dca5a49df7fe0f9
    Time: 1377ms
      Asset     Size  Chunks             Chunk Names
    main.js  88.1 kB       0  [emitted]  main
       [2] ./src/Main.js 7.17 kB {0} [built]
       [3] (webpack)/buildin/global.js 488 bytes {0} [built]
       [6] ./src/mn-morphdom-renderer.js 312 bytes {0} [built]
       [8] ./src/rowtemplate.js 947 bytes {0} [built]
       [9] ./src/mn-domapi.js 987 bytes {0} [built]
        + 5 hidden modules

    ERROR in ./node_modules/backbone/backbone.js
    Module not found: Error: Can't resolve 'jquery' in '/Users/stefan/Source/Javascript/js-framework-benchmark/frameworks/marionette-v4.0.0-keyed/node_modules/backbone'
     @ ./node_modules/backbone/backbone.js 17:4-21:6
     @ ./src/Main.js
```
I tried to add an alias to jquery-thin in wepack-config which makes the build work
```		
alias: {
            'jquery' : 'node_modules/jquery-thin/jquery.js'
}
```
but it then fails in the browser with
```
Uncaught ReferenceError: Sizzle is not defined
    at main.js:11
    at Object.<anonymous> (main.js:11)
    at e (main.js:1)
    at main.js:1
    at Object.<anonymous> (main.js:1)
    at Object.<anonymous> (main.js:1)
    at e (main.js:1)
    at Object.<anonymous> (main.js:1)
    at e (main.js:1)
    at main.js:1
```
When I replace jquery-thin with jquery it builds and works - but I suppose that's exactly not what was intended.
Can you please take a look at it?

paulfalgout:
Bah sorry!  In a rush to get this in prior to leaving for a long trip I pulled the wrong branch.  I pushed the correct updates here.  But this made me realize I had missed removing the unneeded dependencies, so the `marionette-v4.0.0-keyed` run should be better than the screenshot... particularly in the memory related tests.

krausest:
Thanks - now it just worked!
bildschirmfoto 2018-06-16 um 22 25 40
Very nice performance (for the non-jquery version)! Are there any plans to move the mn-* sources to the marionette core library?

paulfalgout:
Well Marionette is becoming more of a plugin-type lib (and kind of always has been) so rather than implementing a particular renderer or templating solution we'll have multiple plugins for multiple solutions, but they'll all be no more complicated than what's here and most have already been implemented.  We'll likely be dropping the dependency on backbone at some point which might see the default implementation change, but right now Marionette defaults to essentially what comes for free with backbone which also has a clearer upgrade path for those coming from v1+.  However I don't think that is even the majority configuration.  The nativeview implementation is the oddest `mn-` here but is used publicly.  It actually doesn't make much difference to the performance, but allows dropping jquery as a dependency which helps with the memory load.  It really just gets around the backbone dependency.  Either way I'd be surprised to see jQuery too much in the future.  Too many benefits of other templating solution.  Granted changing the rendering/templating solution will make the results vary wildly, but this shows the best (probably) and worst cases..  interestingly with marionette you can implement different solutions for different views, so people coming from v1 can update the renderer on their hottest views first without needing to update their entire app.  Or they can choose the templating solution for their favorite API and optimize beyond that implementation for a particular view.

PR #408, opened 2018-06-13
bump vue to 2.5.16

wojtask9:
Also bump webpack to 4.x


krausest:
Thanks - here's a screenshot for some results (I'm not pushing an updated result table, because chrome is now at version 67 and I'll have to re-run all frameworks):
bildschirmfoto 2018-06-13 um 20 38 01
What a pity that vue isn't better at partial update.

wojtask9:
@krausest 
thank you very much.
About `partial update` i'm looking at this. I have some idea what can be wrong. I'll check this (probably next week)

PR #409, opened 2018-06-14
fix vue partial update test

wojtask9:
This PR align what angular and react does in update partial test.


krausest:
Thanks, but I'm sorry to say it does not help much...

bildschirmfoto 2018-06-16 um 22 25 40

PR #411, opened 2018-06-19
Glimmer-1.5.1-keyed

lifeart:


krausest:
@lifeart Thanks for the PR. I'm afraid there are a few issues with the code (only the first is major):
1. Removing more than one row causes an error (Uncaught (in promise) TypeError: Cannot read property 'didDestroy' of undefined at Fs.delete (app-157316eb4ebecc61797d847c29c0ca37.js:630))
2. The css should be included as ```<link rel="stylesheet" href="/css/currentStyle.css">```
3. The app div is missing the container class

Could you please explain the differences to the glimmer 0.9.1 implementation? 
@chadhietala @mixonic Would you agree to replace the 0.9.1 version with that version (after the issues are resolved)?

lifeart:
@krausest done. 
differences to the glimmer 0.9.1 implementation - custom keys management. 
Bumped glimmer version (memory + perf improvements)

krausest:
Sorry, for the late response. Your version for glimmer is non-keyed. This would at least me we'd have to rename and tag the implementation though I'd rather prefer to keep only a keyed version (#304 ).
```
tefans-MBP:webdriver-ts stefan$ npm run compile && npm run nonKeyed -- --framework glimmer-v1.1.5-keyed

> webdriver-ts@1.0.0 compile /Users/stefan/Source/Javascript/js-framework-benchmark/webdriver-ts
> tsc


> webdriver-ts@1.0.0 nonKeyed /Users/stefan/Source/Javascript/js-framework-benchmark/webdriver-ts
> cross-env LANG="en_US.UTF-8" node dist/nonKeyed.js "--framework" "glimmer-v1.1.5-keyed"

Frameworks that will be checked [ 'glimmer-v1.1.5-keyed' ]
glimmer-v1.1.5-keyed is non-keyed for 'run benchmark' and non-keyed for 'remove row benchmark' non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
ERROR: Framework glimmer-v1.1.5-keyed is not correctly categorized in commons.ts
```

lifeart:
@krausest this is Keyed version https://github.com/krausest/js-framework-benchmark/pull/411/files#diff-5124f58dd74a656d7f382b16713d98f8R32

krausest:
Can't tell from the source, but your implementation is definitely non-keyed. The nonKeyed test says so and you can easily check yourself:
Create 1,000 rows, then change with the dev tools the background of the third row to red. Then delete the first row. In a keyed implementation the second row must now be red, not the third as in your implementation.

krausest:
Closing this pull request.

PR #412, opened 2018-06-25
upgrading slim to 3.3.0

eavichay:


krausest:
There was small bug in the implementation:
Remove two rows, then swap rows:
bildschirmfoto 2018-07-04 um 21 02 23
I reverted the condition check in swapRows back to ```if(this.data.length > 998) {```. I suppose this is ok, isn't it?

eavichay:
sure. I copied-pasted it from the React implementation. If that works, it's fine.

10x.

👍

krausest:
Thanks . [Results](https://www.rawgit.com/krausest/js-framework-benchmark/fd5e18032c37921afb2d5fdfa6b8041d6b8b0097/webdriver-ts-results/table.html) have been updated.

PR #413, opened 2018-06-26
Add Angular Ivy implementation and general maintenance

trotyl:
Added a new implementation based on the new [Ivy engine of Angular](http://ivy.angular.io) which mainly focused on app size (regarding this benchmark).

The Ivy result here should be compared to `angular-noopzone-keyed` benchmark.

Maintenance parts:

+ Upgrade Angular version to 6.1.0;
+ Upgrade AngularJS version to 1.7.2;
+ Rename `no-zone` to `noopzone` to better match the API usage;
+ Rename `angular-v1` to `angularjs-v1` to match the [branding guideline](https://blog.angularjs.org/2017/01/branding-guidelines-for-angular-and.html).

Separate commits used to make a perfect git history, but seems not be respected by "File changed" summary, suggesting review individual commit if needed.


trotyl:
CI failed on `halogen-v2.1.0-non-keyed`, not related to this PR.

trotyl:
Ready for review now.

krausest:
Thanks. I'm currently reorganising the directory structure and will update the results when I'm done the next few days.

krausest:
Thanks . [Results](https://www.rawgit.com/krausest/js-framework-benchmark/fd5e18032c37921afb2d5fdfa6b8041d6b8b0097/webdriver-ts-results/table.html) have been updated. Ivy is nice in terms of startup performance, but runtime performance isn't too exciting. Do you think that'll change?

trotyl:
@krausest That's expected, Ivy has moved more compiler works to runtime, thus reduce the restrictions of compiler, and allowing lib to publish compilation results.

While the build speed be much improved, there would be a small loss of runtime speed certainly.

PS: Angular has always been sacrificing runtime speed while evolving, but likely the gain is not too much covered by this project. 😄

PR #417, opened 2018-07-17
domvm v3.4.3

leeoniya:
hey @krausest 

the way event binding works has changed within the lib (single document-level capturing listener per event type). this increased perf significantly. additionally, the pico build is now used, which does not do any kind of rAF-based redraw debouncing, leaving everything synchronous and also requires some things to be more explicit. for fun, i also made a version that uses stateful components/views for each row and the perf was exactly the same (or even better). i'm not submitting it though since i prefer this version (it's a tad simpler and less LOC). i get the following numbers on my machine (rendered in my alternative WIP results-ui builder):


leeoniya:
- v3.4.1 updated rollup for better tree shaking
- v3.4.2 updated Closure Compiler and build script to reduce `devDependencies` bloat by 1,800 files and 20MB.

feel free to squash.

krausest:

krausest:
[Results](https://rawgit.com/krausest/js-framework-benchmark/c4c75a4d11d0cb6f81346f0d4c226b2500d86bc0/results-ui/table.html) for you are now available :bowtie:

leeoniya:
thanks! looks good, no surprises \o/

PR #419, opened 2018-07-18
Added domc v0.0.2

Freak613:
[domc](https://github.com/Freak613/domc) is very basic attempt to implement [Apple's HTML templates proposal](https://github.com/w3c/webcomponents/blob/gh-pages/proposals/Template-Instantiation.md) and bypass Surplus performance along the road)


krausest:

krausest:
Thanks . [Results](https://www.rawgit.com/krausest/js-framework-benchmark/fd5e18032c37921afb2d5fdfa6b8041d6b8b0097/webdriver-ts-results/table.html) have been updated. Nice performance - is a keyed version possible?

ryansolid:
@Freak613 Awesome results on the non-keyed submission. That really highlights the raw capability the library love to see a keyed implementation.

adamhaile:
It looks like the lib doesn't (yet?) handle repeating elements -- the row updating is implemented in user code: https://github.com/krausest/js-framework-benchmark/blob/c4c75a4d11d0cb6f81346f0d4c226b2500d86bc0/frameworks/non-keyed/domc/src/app.js#L108

Freak613:
Yes, current `render` method implementation is prototype of `for` directive that will be moved into the library.
Currently, tests are for the demonstration of performance of updating 10k items without any hidden magic, just straightforward implementation. 
Later I'll push new version of library with `for` inside, along with new performance improvements via `cloneNode` that allows to constantly dominate vanillajs version.

Talking about keyed version: I really like how simple non-keyed implementation is, and browser handles nodes updates much better than reconciling them.
So I start asking myself, do we really need keyed implementation? What real-world cases it'll cover besides "if dom state is modified externally"?
Maybe one more case is related to losing focus caused by shifting the elements...

I was playing with @adamhaile most performant reconcile implementation (maybe it should be released as standalone version):
https://github.com/adamhaile/surplus/blob/master/src/runtime/content.ts#L86
And what I could do here is just to combine this method with my code, because there is nothing better possible here.
I will think about that.

adamhaile:
@Freak613 If you're looking for a good start on reconciling children, I'd recommend @localvoid 's well-documented implementation in ivi: https://github.com/localvoid/ivi/blob/1ad5a11cc5e59e6496469cbf617a5b940232d48f/packages/ivi/src/vdom/sync.ts#L501 .  Surplus's is influenced by his.

Besides animations, non-keyed has trouble with things like selection state.  Your user clicks in an input element, but then an update arrives from the server changing the list, and now their cursor is in a different item's input box.  To be a general solution, non-keyed would need to handle this and other kinds of state that almost no framework does.

Surplus originally had a `cloneNode()`-based renderer but ditched it in favor of compiling to DOM ops.  The problem with `cloneNode()` was that after cloning, you had to navigate through the new DOM structure to find the nodes that needed updating.  With DOM ops, you get that for free, since you have a reference on each node already.  That made cloning slower than the DOM ops approach.  But I was testing a while ago, so maybe `cloneNode()` is faster now and/or your needs might be different.

Freak613:
To make cloneNode performant, you need to minify template, removing `\r\n` nodes and all dynamic parts.

And from my [README](https://github.com/Freak613/domc):
> According to morphdom there are some DOM properties that doesn't require additional computation and therefore they're very fast to work with. Instead of building DOM by .createElement API calls with a lot of memory garbage, it's more effective to clone template node and use fast props to obtain references, with as less performance overhead as possible.

You can check out generated code example to get idea what happens inside domc.
I've rolled out this yesterday and got significant performance boost.
Initially, I've also built DOM using API calls for same reason, but was researching how to avoid memory garbage that comes with it.

adamhaile:
My experience was that the .firstChild and .nextSibling calls weren't free the first time they were called, b/c the JS engine had to construct the JS wrapper object for interacting with the underlying C++ nodes.  

But the proof is in the testing, and I look forward to seeing how it works out for you!

Freak613:
@ryansolid here we go:
https://github.com/Freak613/js-framework-benchmark/tree/master/frameworks/domc-v0.0.4-keyed
Still not pushing it yet, because of inlined render method, as we discussed earlier.

screen shot 2018-07-29 at 05 24 10

ryansolid:
That looks promising. 

Looking at this implementation I wonder if the guidelines for submission should provide more details. I hit something similar when I submitted Solid, where I was doing certain things that were being done in the Vanillajs implementation that were considered unsuitable for any of the submitted libraries. We sort of arrived at an agreement between the library writers what was considered reasonable and I removed a couple of my optimizations.

I wonder if there should be considerations around application entry point. Most if not all libraries use DIV.main as the anchor point. From a TodoMVC sort of standpoint I like that. It demonstrates managing nested contexts even if as this implementation demonstrates you can hang the event handlers off the window and essentially start the implementation at the TR.  This allows much less code that bypasses the need for certain considerations that I think should be on display here. I'm not sure what the consensus is but I'm going to ask the question.

PR #420, opened 2018-07-22
attodom 0.5.2=>0.6.1

hville:
Further API simplification to only expose Nodes.
Slightly faster with smaller memory footprint still.
Plain ES5 should also be very fast on phones and older browsers


krausest:

krausest:
Thanks . [Results](https://www.rawgit.com/krausest/js-framework-benchmark/fd5e18032c37921afb2d5fdfa6b8041d6b8b0097/webdriver-ts-results/table.html) have been updated. Nice!

PR #421, opened 2018-07-23
Update Marionette v4

paulfalgout:
Using the beta with a couple of additional perf updates.  I don't expect to need to update this again after the beta is complete.


krausest:

krausest:
Thanks . [Results](https://www.rawgit.com/krausest/js-framework-benchmark/fd5e18032c37921afb2d5fdfa6b8041d6b8b0097/webdriver-ts-results/table.html) have been updated. Really not bad!

PR #422, opened 2018-07-23
Update bower to fix CI

paulfalgout:
Bower deprecated and removed their old registry at http://bower.herokuapp.com/ and so now only 1.8.4 will work


paulfalgout:
Related: https://github.com/bower/bower/issues/2519

Though ideally bower could be dropped as a dependency

krausest:
Thanks. I allowed myself to fix the bower version (though maybe bowwer is the one package that doesn't make the build unstable).

paulfalgout:
I would be extremely surprised to see another bower version.  This update was to switch to the new registry, and what broke was the old registry was removed for cost reasons.  There are no issues marked for an upcoming release newer than 2015.  Ideally bower is removed entirely.

PR #423, opened 2018-07-24
new results-ui

leeoniya:
Hi @krausest,

This is a from-scratch implementation of an interactive results UI. This work builds on top of the necessary minor tweaks in #418.

The original motivation for this was to reduce `node_modules` bloat in `webdriver-ts-results` (~100MB, ~13,000 files) and perhaps improve build times, bundle size as well as boost UI performance. I'm pleased to report that the final result has surpassed my original expectations by a wide margin.

Full feature parity with `webdriver-ts-results` was not a goal, so some things got the axe:

- **Use median instead of mean** - recent improvements in the benchmark runner have made the differences mostly irrelevant
- **Separate keyed and non-keyed** - the ability to display keyed and non-keyed frameworks in the same table is gone
- **Compare with...** - i never really understood how this feature was supposed to work, including why it has a drastically different color scale. How is it different than simply enabling only the frameworks you're interested in comparing?
- **building results from `src/results.ts`** - the results-ui now only builds from the raw `.json` files produced by `webdriver-ts` because it doesn't matter what that file contains since all its contents are from a different environment. if you don't have the original `.json` files produced on your own machine, they may as well not exist.
- **CSS** - dropped bootstrap due to huge size in favor of custom css

The important stuff is still there:

- Toggle frameworks
- Toggle benchmarks
- Sort by benchmark, name or total slowdown
- Reverse sorting **NEW**

Additional notes:

- Some bench names and descriptions no longer represented what they actually did, so I adjusted the names to be more-or-less uniform and accurate.
- The color scale is not identical but it's quite close. Rather than using a smooth scaling function, I generated a 16 color scale [1] that closely resembles the current one and selected the step thresholds [2] carefully to match what exists now. This has a good UI performance benefit since classes are much more likely to be shared during mutation and do not need to change as often as style attributes tweaked precisely to the factor.
- For better compression, I reduced the precision of all measurements to 2 decimal places, followed by encoding all benchmarks into arrays rather than objects. These are inflated at run-time with minimal overhead, but reduce the bundle size enormously.
- The implementation is based on domvm, rollup/buble and uglify-js
- There are no vdom optimizations in this code - the whole vdom is regenerated and diffed for all interactions. This is done for simplicity, but there's still some perf left on the table, if needed.

Now for some fun stats and pictures:

- **node_modules:** 97.6MB (12,927 files) -> 12.0MB (536 files)
dependencies
- **build time & `table.html` size:** 18.3s -> 2.7s, 1MB -> 197KB
build-times
- **initial render (scripting):** 440ms -> 100ms
load
- **sorting by name, each benchmark and total (scripting):** 2333ms -> 260ms
sorting-11x
- **toggling each keyed bench 2x (scripting):** 2090ms -> 450ms
toggle-benches

(The build time improvement is actually even more drastic since `webdriver-ts-results` was building from a prepared `results.ts` while `results-ui` was synchronously reading/parsing 1,656 `.json` files from `webdriver-ts/results`)

[1] https://www.strangeplanet.fr/work/gradient-generator/?c=16:63BF7C:FFEB84:F9696C
[2] https://github.com/leeoniya/js-framework-benchmark/blob/results-ui/results-ui/src/ui.js#L102-L107

### DIFF
--- a/results-ui/package.json
+++ b/results-ui/package.json
@@ -1,0 +1,24 @@
+{
+  "name": "results-ui",
+  "version": "1.0.0",
+  "description": "js-framework-benchmark results ui",
+  "main": "index.js",
+  "scripts": {
+    "build-prod": "node ./src/build.js"
+  },
+  "keywords": [],
+  "author": "Leon Sorokin",
+  "license": "MIT",
+  "homepage": "https://github.com/krausest/js-framework-benchmark",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/krausest/js-framework-benchmark.git"
+  },
+  "devDependencies": {
+    "domvm": "git://github.com/domvm/domvm.git#3.4.3",
+    "rollup": "*",
+    "rollup-plugin-buble": "*",
+    "rollup-plugin-uglify": "*"
+  },
+  "dependencies": {}
+}
--- a/results-ui/src/build.js
+++ b/results-ui/src/build.js
@@ -1,0 +1,95 @@
+const fs = require('fs');
+const rollup = require('rollup');
+const buble = require('rollup-plugin-buble');
+const uglify = require('rollup-plugin-uglify').uglify;
+const path = require("path");
+
+const start = +new Date();
+
+const RESULTS_PATH = path.resolve(__dirname + '/../../webdriver-ts/results');
+
+function encodeBench(obj) {
+	return [
+		obj.benchmark.substr(0,2),
+		+obj.min.toFixed(2),
+		+obj.max.toFixed(2),
+		+obj.mean.toFixed(2),
+		+obj.median.toFixed(2),
+		+obj.geometricMean.toFixed(2),
+		+obj.standardDeviation.toFixed(2),
+		obj.values.map(v => +v.toFixed(2)),
+	];
+}
+
+let libs = {
+	keyed: {},
+	unkeyed: {},
+};
+
+// grab result files, group by framework, bench types and encode benches into arrays
+fs.readdirSync(RESULTS_PATH).filter(file => file.endsWith('.json')).forEach(file => {
+	var r = JSON.parse(fs.readFileSync(RESULTS_PATH + "/" + file, 'utf-8'));
+	var implGroup = r.keyed ? libs.keyed : libs.unkeyed;
+	var libName = r.framework;
+
+	if (implGroup[libName] == null) {
+		implGroup[libName] = {
+			name: libName,
+			bench: {
+				cpu: [],
+				memory: [],
+				startup: [],
+			},
+		};
+	}
+
+	implGroup[libName].bench[r.type].push(encodeBench(r));
+});
+
+// convert to arrays and sort
+Object.keys(libs).forEach(implType => {
+	libs[implType] = Object.values(libs[implType]).sort((a, b) => a.name.localeCompare(b.name));
+});
+
+fs.writeFileSync(path.resolve(__dirname + '/data.js'), 'export default ' + JSON.stringify(libs), 'utf-8');
+
+async function build() {
+	const bundle = await rollup.rollup({
+		input: __dirname + "/ui.js",
+		plugins: [
+			buble(),
+			uglify(),
+		],
+	});
+
+	const { code, map } = await bundle.generate({
+		format: "iife",
+	});
+
+	const html = [
+		'<!doctype html>',
+		'<html class="no-js" lang="">',
+		'<head>',
+		'<meta charset="utf-8">',
+		'<meta http-equiv="x-ua-compatible" content="ie=edge">',
+		'<title>Interactive Results</title>',
+		'<meta name="description" content="">',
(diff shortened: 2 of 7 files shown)


krausest:
Thanks. I'll take a look at it when I rerun the results with Chrome 68.

I'm not completely sure that we're after the same goals though. I don't care too much about dependency size of the result-ui (I have 94 implementations and some with large dependencies - I'm not sure if this is the place to start sizing down) though faster build times are very welcome of course.

The compare with feature is essential to me. It's such a shame that I failed to explain what it does. Put shortly: If you compare against a framework and you don't get either a red or green box for the other framework you still might say one is faster than the other, but you can't back your claim on statistical significance.
Benchmarks neglecting statistic tests can be easily criticized and I'm working hard not to be.

I think at first I'll add your results-ui as an alternate results-renderer and gradually merge it, but I want to keep (p)react [sorry!], compare with, gradient results and mean/median. I'll also merge #418 for the new results.

Maybe we should start a new benchmark from the results renderer? By now we have a react, preact, vue and domvm implementation  :smirk:

leeoniya:
i probably have an irrational fear of huge dependencies. part of it might be due to the fact that i run many things portably (off a USB flash drive) so installing/deleting/updating 13,000 files is significantly worse than 500 files.

> Put shortly: If you compare against a framework and you don't get either a red or green box for the other framework you still might say one is faster than the other, but you can't back your claim on statistical significance.

do you have any examples from the current data where the mean and/or median are misleading, but get saved by sound statistics (besides < 20ms tests, where variance is wild).

> I'll also merge #418 for the new results.

that would be great.

thanks!

krausest:
I had a small problem with missing results and tried to fix it  - might not be the most elegant way.

leeoniya:
fix looks good, though i try to avoid truthy/falsey coercions in property checks, preferring `x.y == null` to `!x.y`. also, if you use spaces, it probably makes sense to convert the whole file to spaces. your changes makde it mixed spaces/tabs.

krausest:
@leeoniya Thanks for the review, it was badly needed. I reviewed the code and compared it to the results of the nice web site above and excel and found 3 (!) mistakes:
* std deviation was computed incorrectly: missing true parameter to `jStat(result.values).stdev(true)`
* Sum of squares was incorrect bracketed (oh no...)
* Colouring started actually with a  p-Value of 20 instead of 10
I'm now printing the p-Value below the factor such that checking the values should be easier.
bildschirmfoto 2018-08-06 um 21 37 21
The web site you found computes almost the same value:
bildschirmfoto 2018-08-06 um 21 38 33
And now the result looks like that:
bildschirmfoto 2018-08-06 um 21 52 24
Looking forward your review!

leeoniya:
@krausest 

DISCLAIMER: i'm not a statistician, so treat everything i say with skepticism.

a typical t-test is double-sided [1] in that its p value that represents the likelyhood of being wrong in both directions. (e.g.: mean is *either* greater *or* smaller).

maybe you want to take only the implied side of hypothesis, so in this case, `d ≥ 0` for inferno >=ivi. which gives you the `p` value for the mean only being greater rather than also including the chance that it was less. this halves the p value from 0.00466 to 0.00233. i'm not really sure if this is appropriate here, but since the colors imply only a specific side of the assumption, maybe it makes sense?

one thing i find confusing is why the p-value drops when you only take one side. it's not like excluding the probability of being wrong in the opposite direction makes you more certain of the other direction.

also, since we could be comparing metrics with differing variances, a Welch's t-test [2] would be more appropriate.

[1] https://stats.idre.ucla.edu/other/mult-pkg/faq/general/faq-what-are-the-differences-between-one-tailed-and-two-tailed-tests/
[2] https://en.wikipedia.org/wiki/Welch%27s_t-test

leeoniya:
also, i'm pretty sure you can't do a t-test using medians: https://github.com/krausest/js-framework-benchmark/blob/master/webdriver-ts-results/src/Common.tsx#L248

if you plan to keep using jStat, i recommend switching over to using a Welch's t-test unconditionally [1] (steal code from here [2]). still not sure if a 2-tailed or 1-tailed test is appropriate here :/

[1] http://daniellakens.blogspot.com/2015/01/always-use-welchs-t-test-instead-of.html
[2] https://www.icalcu.com/stat/two-sample-t-test-calculator.html

krausest:
Using medians makes indeeed no sense for the t-test.

I've now switched to welch's t-test. Excel and my program report the same values.
The resulting p-Values are very similar to a t-test (not sure how bad the assumption of equal variance was...)
Since I'm currently colouring p-Values below 10% (which is rather high) for a two-sided test it is  equivalent to one-sided tests at 5% (which is okay). So I see no reason to change that.

leeoniya:
nice.

> The resulting p-Values are very similar to a t-test (not sure how bad the assumption of equal variance was...)

either way, Welch's properly handles different sample sizes as well as variances...all without any downsides. at the end of the day, reducing variance should be a primary goal of this bench, so hopefully we dont have to rely on being saved by Mr. Welch :)

PR #425, opened 2018-07-28
Update solid-js to v0.0.10

ryansolid:
Significant rewrite to improve performance and memory usage.
Removed all implementation hacks (no stored tr, no rAF). 

Removing those hacks was a very significant performance de-optimization, but hopefully the net is close to neutral (although I think it's a bit slower).


krausest:
Thanks . [Results](https://www.rawgit.com/krausest/js-framework-benchmark/fd5e18032c37921afb2d5fdfa6b8041d6b8b0097/webdriver-ts-results/table.html) have been updated. Performance looks pretty solid :sunglasses: Nice!

PR #426, opened 2018-07-28
update Surplus to 0.5.3

adamhaile:
Here's one to test the new maintenance approach :)

Updates Surplus to latest, 0.5.3, along with updating dependencies.

- Removes rAF clear optimization
- Adds an optimization on setting className

For me, comes out about the same to slightly faster.  Hard to distinguish from the noise (mostly the swap test, which seems to decide the top five placings on its own :P).

### DIFF
--- a/frameworks/keyed/surplus/package.json
+++ b/frameworks/keyed/surplus/package.json
@@ -25,15 +25,15 @@
     "url": "https://github.com/adamhaile/js-framework-benchmark.git"
   },
   "dependencies": {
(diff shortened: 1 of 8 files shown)


krausest:
Thanks . [Results](https://www.rawgit.com/krausest/js-framework-benchmark/fd5e18032c37921afb2d5fdfa6b8041d6b8b0097/webdriver-ts-results/table.html) have been updated. Looks very nice!
How was your experience with the new directory layout and dynamic framework list creation?

PR #431, opened 2018-07-31
WIP: Allow skipping irrelevant builds/tests

alexfmpe:
This is still at a proof-of-concept/experimental stage, but continuing implies some invasive changes, so I'd like to get some feedback for suitability/usefulness/etc before going any further.

My original motivation was making CI faster in the extremely common case that a PR only touches a few frameworks, thought this might end up being useful for local dev.

The main idea is that if we only care about whether we broke things, we can ignore framework folders that haven't been changed (unless the `webdriver-ts` folders changed).
Caching `node_modules` can allow us to effectively skip re-builds, but not the actual benchmarks.

Here's a sample commit which only changes [two frameworks](https://github.com/alexfmpe/js-framework-benchmark/commit/6b146590cc001b12baa6206025afeb5b934a4428) with a [3 min build](https://circleci.com/gh/alexfmpe/js-framework-benchmark/226) instead of the usual ~2 hours.

Remaining issues:
- Considering untracked files - `git diff` ignores them
- Due to `--frameworks` matching by prefix, passing it the name of one framework name can also select others (e.g `angular` -> `{angular, angular-ivy,...}`). This is a useful feature which I'd like to keep, but implies that going through with this PR would require pulling the project-structure logic in `benchmarkRunner.ts/common.ts` to the top level.


krausest:
To be honest: I don't benefit much from the current circleci builds. For my taste I'm getting too many broken builds for no good reason. (But I also admit I didn't invest significant time in improving things...) Maybe this helps making the builds more reliable.
I added restartWith as a parameter to the script.

PR #432, opened 2018-08-01
MnV4 change label directly perf improvement

paulfalgout:
I was told I wasn't embracing the jquery enough for this test, and they were right.

### DIFF
--- a/frameworks/keyed/marionette-jquery/src/Main.js
+++ b/frameworks/keyed/marionette-jquery/src/Main.js
@@ -162,9 +162,9 @@ const MyCollectionView = CollectionView.extend({
 
         this.swapChildViews(view1, view2);
     },
-    onChangeLabel(model) {
+    onChangeLabel(model, label) {
         const view = this.children.findByModelCid(model.cid);
-        view.render();
+        view.$('.js-link').text(label);
     },
     onRender() {
         this.clearSelected();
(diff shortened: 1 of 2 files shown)


krausest:
Results are available [here](https://rawgit.com/krausest/js-framework-benchmark/3cf45a2e2053a4b5c6ff8af429e67c9b736830ba/webdriver-ts-results/table.html)

krausest:
I'll check this evening and re-run it.

krausest:
@paulfalgout Sorry, you're absolutely right. Shame is I created a rebuild script that should help me to avoid these situations, but I failed to invoke it for the right framework (and built keyed/marionette but not keyed/marionette-jquery). I'll update the results soon. From a first run (3 iterations) I saw a median of 85 msecs for partial update - much better! 

As for the variance in the previous [run](https://rawgit.com/krausest/js-framework-benchmark/fd5e18032c37921afb2d5fdfa6b8041d6b8b0097/webdriver-ts-results/table.html) create 1,000 rows was indeed faster and had a much lower std deviation, could be that some background process had a negative impact.

krausest:
[Here's](https://rawgit.com/krausest/js-framework-benchmark/b16730a86128d049787063a1efb21ff1f26b102a/webdriver-ts-results/table.html) another run.
Partial update is down - as you expected. Created rows now also has a much lower variance.
Do these numbers look more plausible?

paulfalgout:
Yep excellent thanks!

PR #435, opened 2018-08-02
results-ui cleanups

leeoniya:
some cleanups:

- tabs -> spaces
- cleaner templates (pre-filter enabled benches & frameworks)
- split `bootstrap-reboot.css` into own file
- concat css and use `clean-css` to minify
- bump domvm to v3.4.4 (Uglify isn't as good at dead code elim as Closure, so it left behind an unused `requestAnimationFrame`, which is now feature-gated)

the included `table.html` is generated from latest `results.ts` that i extracted into `.json` files.


leeoniya:
added a commit to use minified bundle for keyed domvm. unrelated, but saves an extra PR.

krausest:
Since I used the wrong method to compute the std deviation it reports a too low value (factor 1,054092553 for an iteration count of 10). I fixed it in webdriver-ts, but until all results have been re-run results-ui won't use that data. In webdriver-ts-result. I'm currently computing the std deviation from the raw values. Not sure if you want results-ui to follow that decision.

leeoniya:
> I'd like to add yargs as a dependency and allow passing a framework name pattern as a parameter. That would allow me to create a result table easily for all available versions of a framework.
Would you object?

yargs is pretty lightweight, so that'd be fine. i assume it'll just add a bit of code to the build script for framework filtering? what param format did you have in mind?

> In webdriver-ts-result. I'm currently computing the std deviation from the raw values. Not sure if you want results-ui to follow that decision.

eh, this greatly bloats the bundle for not good reason and adds overhead to the frontend for re-computing stats that will never change. the raw values aren't necessary for anything in the current UI. event t-tests just need mean/median, variance/stddev and sample counts.

i'm working on a compare feature for the results ui using https://www.npmjs.com/package/ttest (which also has support for Welch's [1]. the visual approach i'm taking is somewhat different than what webdriver-ts-results does, tho.

this doesnt have ttest intgrated yet, just green and red color scales for 0.5x - 2.0x factors.

wip-compare

[1] https://github.com/AndreasMadsen/ttest/blob/master/hypothesis/welch.js

leeoniya:
> In webdriver-ts-result. I'm currently computing the std deviation from the raw values. Not sure if you want results-ui to follow that decision.

after some thinking, maybe it's not such a bad idea to have all `.json` files from `webdriver-ts` exclude any stats and just do all the stat calcs in the UI. i made a small `stats()` function that i'll use for this purpose [1]. the smallest impl of Welch's t-test that i've found is here [2] and i'll try to integrate that into my results-ui impl.

[1] https://gist.github.com/leeoniya/734c2d89146687c3e262b492ab024aaf

PR #439, opened 2018-08-05
add stdweb and yew (Rust+WebAssembly benchmarks)

saturday06:
Hi,
I wrote benchmark target apps with following frameworks:

- stdweb is low level api that has Web and DOM bindings.
- yew is high level framework like Elm or Redux. non-keyed only.

Please add it.


saturday06:
Benchmark is failing on CircleCI. I'll write a fix.
```
Benchmark failed { TimeoutError: Waiting testElementLocatedByXpath //tbody/tr[1000]/td[2]/a
Wait timed out after 60009ms
    at /home/circleci/project/webdriver-ts/node_modules/selenium-webdriver/lib/webdriver.js:834:17
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7) name: 'TimeoutError', remoteStacktrace: '' }
Writing screenshot error-stdweb-01_run1k.png
```

saturday06:
That failure was fixed by https://github.com/krausest/js-framework-benchmark/pull/439/commits/9de78e75115d7032c594bfb74238289cc9cd7516

krausest:
Thanks - pretty interesting. stdweb is indeed very vanillajs alike (far from a high level framework) but having a webassembly entry here ist very welcome.
Yew is quite on the right side (of the table)...
As the dependencies and build is very heavy I decided to go the way we did for the haskell frameworks. Theres a build-prod-force task that actually builds and the build-prod task just echos that it a bit heavy. Hope you don't mind.
Results are [here](https://rawgit.com/krausest/js-framework-benchmark/e713af4db483ffc627798caa6f5ea0ac4ca1e5fd/webdriver-ts-results/table.html)

saturday06:
I see. Thanks!

krausest:
Yes it runs on my machine. But it seems like I added a no-op for the build and forgot to add the compiled JS.
The rust libraries are very heavy and take pretty long to compile so I decided not to build them unless requested (as for some haskell libs)
Can you please try if the following works:
`cd frameworks/non-keyed/yew/`
`npm run build-prod-force`
try to open http://localhost:8080/frameworks/non-keyed/yew/
Do you see the page with the buttons for yew? If so now the benchmark runner should work.
I'll create an issue to commit the compiled JS files to github.

szmtcjm:
@krausest em... I use preexist .js and .wasm in the `frameworks/non-keyed/yew/` dir and I see the page with the buttons for yew. But it also has the error.

There are rust and cargo env in my machine too, and I build myself. the error exist too.

krausest:
I checked again. You're right it's not due to missing files - and maybe I could have come to that conclusion earlier if I had taken a closer look at your log excerpt.

It seems like it passes the benchmark initialisation (which means the page works and the run button is clicked 5 times), but it times out in the run method.

Could you please update to the latest version? 
I improved the wait condition in that benchmark such that for each of the five warm up clicks it waits for the correct row to appear. Maybe that helps (or it fails earlier). After `git pull` please go to the `webdriver-ts` directory and issue
`npm run compile && npm run bench -- --framework yew --benchmark 02_`

If it still doesn't work can you please post a screenshot of the browser instance opened by chromedriver just before the timeout happens? It should display the table starting with row 5001, but obviously it doesn't.

Does it work for other frameworks?

Issue #221, opened 2017-07-27
dojo doesn't build with yarn


krausest:
Another build failure with yarn:
```
Stefans-MBP:dojo2-v2.0.0.beta2.5-keyed stefan$ npm run build-prod

> dojo2-v2.0.0-beta2.5-keyed@1.0.0 build-prod /Users/stefan/Source/Javascript/js-framework-benchmark/dojo2-v2.0.0.beta2.5-keyed
> dojo build

Commands are not available: TypeError: Failed to load module /Users/stefan/Source/Javascript/js-framework-benchmark/dojo2-v2.0.0.beta2.5-keyed/node_modules/@dojo/cli-build-webpack
Nested error: Feature "es6-array" exists and overwrite not true.
```
Building with `npm install` just works.

alexteves:
Got same error. Looks like an issue with dojo: https://github.com/dojo/shim/issues/97

krausest:
Disabled yarn. npm 5 got much better anyways.

PR #442, opened 2018-08-11
Upgrade Ember to 3.3, and simplify the benchmark

NullVoxPopuli:
I switched the commands to local versions so it's easier to know what it being used.


frameworks/keyed/ember/app/utils/benchmark-helpers.js
```diff
@@ -0,0 +1,82 @@
+const _random = (max) => {
```
NullVoxPopuli: I pulled this from the react folder

frameworks/keyed/ember/app/components/my-table.js
```diff
@@ -1,35 +1,101 @@
...
+    });
+  }
+
+  @action async swapRows() {
+    startMeasure('swapRows');
+
+    this.set('data', swapRows(this.data));
+  }
+
+  @action async remove(id) {
```
rtablada: Probably don't use `async` without `await` or promises.

NullVoxPopuli: the intent was to allow the UI to not be blocked due to the long running time of the benchmark helpers

leeoniya: startMeasure and stopMeasure can be dropped completely, btw. they are not used by this benchmark.

krausest: The real measurement happens with the test driver in webdriver-ts that uses chrome's timeline feature to measure the duration. Anything else is just an approximation (sometimes not too bad to validate the measurements).

NullVoxPopuli: oh neat. ok, yeah, I'll remove all the measuring stuff.

NullVoxPopuli: @krausest done

krausest:
Thanks all. I allowed myself to remove yarn (so far my experience wasn't good: #221 and #182). I suppose it's fine that way too, isn't it?

Sadly it seems like 3.3 with the changes is a small step backwards regarding performance:
bildschirmfoto 2018-08-26 um 20 56 17
Did you also see a performance regression?

NullVoxPopuli:
Depends, honestly. I can tons of tests, and was comparing vs react instead of vanilla. Also had to make sure I used private browsing with no extensions. And not have much running on my computer.

There were some tests that had the updates much faster than react. Some that the updates were pretty much the same. And some that were slightly slower.

Lots of variance. :/

PR #443, opened 2018-08-12
Update solid(0.0.14) and ko-jsx(0.0.16) to latest version

ryansolid:
Updates to underlying compiler. Streamlined directive syntax and reduced memory usage.

Unfortunately computer that I run the benchmarks is in a weird state where I'm not getting consistent benchmark results but not expecting much performance change.


ryansolid:
I've made a few more version updates. Mostly just package tweaks to reduce the overall package size. The biggest being to use a new alpha version of Knockout which lets us leverage tree shaking with ES6 Modules.

PR #445, opened 2018-08-14
update domvm-v3.4.5

leeoniya:
this update reverts the removal of an internal cleanup pass after some potential memleak issues were discovered related to imperative/retained views (though not used in this bench).

i expect this to slightly reduce perf of metrics with mass removals (1k replace and 10k clear).

really enjoying the new "low maintenance" update process/diff :+1:


krausest:
Here we go:
bildschirmfoto 2018-08-26 um 16 40 12
bildschirmfoto 2018-08-26 um 16 40 28

leeoniya:
thanks!

PR #447, opened 2018-08-15
Update lit-html

ruphin:
Updated the lit-html framework to the latest release


krausest:
Thanks. Here are results for both versions:
bildschirmfoto 2018-08-26 um 19 56 40
bildschirmfoto 2018-08-26 um 19 56 50

PR #448, opened 2018-08-18
Updated domc to 0.0.5

Freak613:
- Added keyed implementation for domc
- All rendering logic is inside the library, app.js has only application logic
- Main entry point is div#main, so no feeling of 'dirty hack' should be now


krausest:
Really impressive.
bildschirmfoto 2018-08-26 um 16 56 58
bildschirmfoto 2018-08-26 um 16 57 08

PR #449, opened 2018-08-21
vanilla-dom non-keyed

mgibas:
Added non-keyed benchmark of [vanilla-dom](https://github.com/mgibas/vanilla-dom).

- all tests run without errors
- fixed dependencies in `package.json`
- generated index.html


mgibas:
Yeah, name came from countless v-dom implementation though this one is just html compiled to vanilla js so there it is - vanilla-dom 😄 

Svelte ? Yeas and no. vanilla-dom is just templating library so It does not bother about components etc. which means it has no opinion on how you want to build them. I also don't care about templates in string literals, jsx and all of that _modern_ stuff with blowed out dev stacks and bazillion dependencies.
I need to say though that projects like svelte and stencil where my main inspiration. 

I think that most similar would be domc (impressive perf 👏 ) so @Freak613 - this means a war and Im barely warming up 😉

PR #450, opened 2018-08-25
Fixed issue #427: MobX test not working correctly

### LINKED ISSUE #427: React MobX error with 08_create1k-after10k
Currently the benchmark 08_create1k-after10k fails and quite a strange way after updating react and mobX (5.0.3, no matter if react 16.1 or react 16.4). It worked fine with mobX 3.3.1.
Creating 10,000 rows and then appending 1,000 causes a loss of the first 10,000 rows. If I change the number of rows created at first to 1,000 it works as expected - sounds strange.

Click create 10,000 rows:
bildschirmfoto 2018-07-29 um 09 06 49
As expected there are 10,000 rows. Then click on append rows:
bildschirmfoto 2018-07-29 um 09 07 03
Here the strange thing happens. This is the modified add action:
```
    @action add() {
        console.log("store.add length before", this.data.length);
        this.data = this.data.concat(this.buildData(1000));
        console.log("store.add length after", this.data.length);
    }
```
It shows that there are 10,000 rows in the store after the add call. But as one can see the row ids have changed and only 1,000 rows are left.
Fun fact: It works correctly for any row count < 10.000
@mweststrate Can you please take a look?

mweststrate:
Using `spliceWithArray` (a splice that don't needs spreading, which can cause issues in some JS engines if the collections is large) instead of concat fixes the issues (atm still unclear why, I could'nt reproduce the issue in a unit test). 

Also made the store more idiomatic by never replacing the `data` property, but always updating it

Fixes #427

### DIFF
--- a/frameworks/keyed/react-mobX/src/Store.es6.js
+++ b/frameworks/keyed/react-mobX/src/Store.es6.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var {observer} = require("mobx-react");
 var {observable, computed, action} = require ("mobx");
 
 function _random(max) {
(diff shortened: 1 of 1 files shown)


krausest:
Thanks for the fix. Results are here:
bildschirmfoto 2018-08-26 um 21 39 43

PR #452, opened 2018-08-28
[results-ui] replace yargs with yargs-parser

leeoniya:
https://packagephobia.now.sh/result?p=yargs
https://packagephobia.now.sh/result?p=yargs-parser

```
yargs:          1.22 MB, 290 Files, 65 Folders
yargs-parser:   53.4 KB,  10 Files,  3 Folders
```

also updates rollup & uglify-js

### DIFF
--- a/results-ui/package.json
+++ b/results-ui/package.json
@@ -17,10 +17,10 @@
   "devDependencies": {
     "clean-css": "^4.2.1",
     "domvm": "git://github.com/domvm/domvm.git#3.4.5",
-    "rollup": "^0.64.1",
+    "rollup": "^0.65.0",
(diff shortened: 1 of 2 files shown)


krausest:
Thanks. I'm on holiday right now. Will be merged in about two weeks.

leeoniya:
cool, no rush.

i might have the "compare" feature done by then, too. been working on it on-and-off trying to avoid sucking in a huge stat lib dependency. good news is that i have the entire working Welch t-test down to 35 LOC. another 35 LOC function in the build script re-computes all the needed stats for the UI from the raw values, so it's no longer necessary to store them on disk as part of the result jsons. the rest is playing with intuitive color scales and such - easy stuff.

PR #453, opened 2018-08-31
Added stage0

Freak613:
This is VanillaJS implementation using tools from [stage0](https://github.com/Freak613/stage0)

These tools are parts of domc, exported as small functions that can be applied for much more cases.
And what I like in this approach, is its flexibility, while being on VanillaJS level without using any abstractions.

Performance is close to domc, but may be a little worse, so I can't decide whether we should remove domc or leave both.

I hope there is no restrictions on amount of libraries from same author :)
Approach of Vanilla components looks very interesting. This gives another view on how vanilla can be done without usage of frameworks.

### DIFF
--- a/frameworks/keyed/stage0/app.js
+++ b/frameworks/keyed/stage0/app.js
@@ -1,0 +1,171 @@
+import {h, setupSyntheticEvent} from 'stage0'
+import {reconcile} from 'stage0/reconcile'
+
+let did = 1
+function buildData(count) {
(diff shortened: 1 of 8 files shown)


krausest:
Cool. I'm on holiday right now. Will be merged in about two weeks.

krausest:
It's so crazy fast that I had to re-run domc and vanillajs on chrome 69 to get a better picture.
Results for Chrome 69 (domc 0.0.2 was run on a previous chrome version):
bildschirmfoto 2018-09-09 um 21 41 19
bildschirmfoto 2018-09-09 um 21 41 29

Freak613:
Great!

But looking at https://rawgit.com/krausest/js-framework-benchmark/master/webdriver-ts-results/table.html
I see that there is some problem with startup metrics. I think it's definitely shouldn't be zero for stage0, and for non-keyed it's missed.

PR #456, opened 2018-09-09
Update SolidJS to 0.1.0 and ko-jsx to 0.1.2

ryansolid:
Borrowed a page from the Surplus Implementation in how to better handle removing computations for row selection, and borrowed a page from Domc, by rewriting the JSX compiler to output similar cloneNode, and element reference traversal techniques at compile time.

Result is cleaner API, and another drop in memory usage, and build size. Performance is a bit better, although not as much as I hoped locally, but I remain optimistic as on my machine(which is lower powered) nothing every comes close to VanillaJS performance.

Hope you are enjoying your holidays. Cheers.

### DIFF
--- a/frameworks/keyed/ko-jsx/package.json
+++ b/frameworks/keyed/ko-jsx/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-ko-jsx",
-  "version": "0.0.18",
+  "version": "0.1.2",
   "main": "index.js",
   "js-framework-benchmark": {
     "frameworkVersionFromPackage": "ko-jsx"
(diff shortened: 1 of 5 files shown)


ryansolid:
Not sure if it saves you any effort being a single pull request, but I took the opportunity to update ko-jsx with the same improvements. Also now that my computer isn't acting up benchmarked the difference between the versions of KO, and current production is significantly more performant, so I reverted. This is by far the best performance I've seen out of the library and I hope it reflects here.

krausest:
Thanks. Results for Chrome 69:
bildschirmfoto 2018-09-09 um 20 56 08

ryansolid:
That's great for ko-jsx-v0.1.2-keyed.  That's where I hoped it would fall. Was there something wrong with solid-v0.1.0-keyed? Or just a matter of timing of getting the results. If ko-jsx did that well I'm very optimistic for the solid release.

krausest:
@ryansolid Sorry - I've seen that I tried to run solid-js. Here are the results (please note results for older solid and ko-jsx versions are from chrome 68):

bildschirmfoto 2018-09-10 um 09 07 20

ryansolid:
Perfect thank you. That's a nice performance boost. Shows where the create/clear overhead is.

Chrome 69 fixed a lot of weird performance issues I was having with 68 on my computer so I think I will be able to more accurately benchmark again.  It may be the cause of some of these performance improvements as Chrome 68 really seemed to do some weird stuff in come situations. Look forward to see how Chrome 69 makes its way through the benchmark.

PR #457, opened 2018-09-09
cleaner and with version 0.9.0

hville:
A little simpler still: cleaner implementation with less magic.
Should be even smaller and faster


krausest:
Thanks. Results look like that for Chrome 69:
bildschirmfoto 2018-09-09 um 20 54 28

PR #459, opened 2018-09-09
Hyperapp update to 1.2.9

SkaterDad:
Updating hyperapp to the latest: v1.2.9.

There were some performance optimizations since 1.2 (the current benchmark version), so I'm very interested in seeing how it does.

### DIFF
--- a/frameworks/keyed/hyperapp/package.json
+++ b/frameworks/keyed/hyperapp/package.json
@@ -7,7 +7,7 @@
     "frameworkVersionFromPackage": "hyperapp"
   },
   "dependencies": {
-    "hyperapp": "1.2.0"
+    "hyperapp": "1.2.9"
   },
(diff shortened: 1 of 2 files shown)


krausest:
Nice speedup!
bildschirmfoto 2018-09-16 um 15 11 54

krausest:
Please keep in mind that it's also a chrome 69 vs. chrome 68 comparison.
Only vanillajs and 1.2.9 were run  on chrome 69.

jorgebucaran:
👋 
Thank you, @SkaterDad @krausest. Nice! 

> I'm sure they new algorithm for V2 handles those better.

Yes, V2 is better. For starters, it uses a faster algorithm, accurate data structures and incorporates these (SkaterDad's) speedup improvements. Add lazy rows to the mix and boom! 💥

PR #460, opened 2018-09-09
Vanillajs use cloneNode instead of document.createElement

ryansolid:
After seeing how much of difference this seems to make for some libraries thought I'd see if it was worth it.  This commit changes the createRow function into a domc-esque node clone:

```js
createRow(data) {
    const tr = rowTemplate.cloneNode(true),
        td1 = tr.firstChild,
        a2 = td1.nextSibling.firstChild;
    tr.data_id = data.id;
    td1.textContent = data.id;
    a2.textContent = data.label;
    return tr;
}
```

This didn't make much of a difference on my computer, but at minimum this can be a comparison point.

### DIFF
--- a/frameworks/keyed/vanillajs/src/Main.js
+++ b/frameworks/keyed/vanillajs/src/Main.js
@@ -22,6 +22,9 @@ function _random(max) {
     return Math.round(Math.random()*1000)%max;
 }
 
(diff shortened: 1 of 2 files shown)


krausest:
Thanks. Seems a bit faster: 126.7 instead of 130.2 for create 1,000 rows.

PR #463, opened 2018-09-28
[BUGFIX] build the ember app in production mode

runspired:


runspired:
nvm, it seemed as though the command being executed was `npm run build` but I see here it is `build-prod` https://github.com/krausest/js-framework-benchmark/blob/master/build.js#L47

PR #465, opened 2018-10-02
Update RE:DOM to 3.14

pakastin:
- Swap rows should be way faster now!
https://github.com/redom/redom/releases/tag/v3.14.0

### DIFF
--- a/frameworks/keyed/redom/package.json
+++ b/frameworks/keyed/redom/package.json
@@ -13,7 +13,7 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "redom": "3.14.0",
+    "redom": "3.14.1",
     "rollup": "0.66.2",
     "rollup-plugin-buble": "0.19.2",
     "rollup-plugin-node-resolve": "3.4.0",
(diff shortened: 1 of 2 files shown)


pakastin:
Hmm.. that test seems to fail in some other framework.. 🤔

pakastin:
CI test seems to fail in `react-easy-state` before even getting to RE:DOM.. Should work alright!

solkimicreb:
It fails the build process, which should be exactly matching with the other react implementations if I remember correctly. I will check what's the issue. (I am maintaining easy-state)

solkimicreb:
Fixed here: https://github.com/krausest/js-framework-benchmark/pull/466. Lets hope it passes.

pakastin:
Ok thanks, let’s hope so! 🤞

krausest:
Results are up now. Here's a (cross browser versions, OS versions) comparison with the older versions:
bildschirmfoto 2018-10-04 um 21 28 35
Nice speedup for swap rows!

pakastin:
Oh yeah! 🤘

PR #466, opened 2018-10-02
fix build process for react-easy-state (bump webpack)

solkimicreb:
Fixes https://github.com/krausest/js-framework-benchmark/pull/465. easy-state was stuck with a beta version of webpack 4, which had some issues. I updated the webpack version to match with the other react implementations. (I think Redux is also behind, but I didn't want to touch it).

### DIFF
--- a/frameworks/keyed/react-easy-state/package.json
+++ b/frameworks/keyed/react-easy-state/package.json
@@ -27,7 +27,7 @@
     "babel-preset-es2015": "6.24.1",
     "babel-preset-react": "6.24.1",
     "jsx-loader": "0.13.2",
-    "webpack": "^4.0.0-beta.1"
+    "webpack": "4.16.3"
   },
   "dependencies": {
     "react": "16.1.0",


solkimicreb:
It fails for surplus now ): This commit https://github.com/krausest/js-framework-benchmark/commit/e235a5fe6872ab1130de9fc651d2947bbed1e53f probably broke a few things

adamhaile:
Let me know if there's something I could do better in the surplus build.  A quick glance at the errors looks like its a typescript version mismatch: whatever version of typescript is doing the building doesn't seem to like the node type declarations.

krausest:
I‘ll remove tomorrow the carets from the version in surplus package.json and I bet it‘ll work.
If it doesn‘t fix it I‘ll get back to @adamhaile

krausest:
I was mistaken. Culprit was a target es5 in the tsconfig.json file.
Let's see where CI fails next ;-)

PR #468, opened 2018-10-05
Update to lit-html 0.12.0

CaptainCodeman:
The repeat directive of lit-html has been re-worked resulting in a significant performance increased for keyed updates!


krausest:
Thanks - has been merged:

PR #470, opened 2018-10-10
Added glasgow-0.2.7-keyed

vanviegen:
Glasgow is a JSX-based lib with a focus on simplicity. https://github.com/vanviegen/glasgow


krausest:
Thanks - has been merged:
I removed all unrelated changes  from the commit except the index.dot update (Thanks for spotting this one). What was the reason for the changes in react, angular and so on?

PR #472, opened 2018-10-14
Reworking Vanillajs

ryansolid:
Hi Stefan,

I've been working a bit the last while on trying to figure out how to improve Vanillajs numbers. It looks like your recent changes to cpu throttling swap rows and select have achieved that. It also bugged me how was it not one of if not the smallest implementation. Other implementations are pulling libraries.

I started this work earlier really trying to pick apart why Stage0 or Domc would be getting better results than Vanilla secretly hoping I'd find some DOM operation trick. I didn't. Instead I noticed something about the implementation itself a bit weird. The stock vanillajs implementations were maintaining 4 lists, instead of 2 and many cases doing the same operation on all 4.  What I mean is there are:

1. store.data, which the methods initially get called on
2. this.data which gets copied from store.data
3. this.rows which is an array of the dom nodes
4. the actual dom nodes themselves as rendered

This was expecially evident in the keyed implementation which in many ways could be a lot simpler since it doesn't need to update the existing nodes everytime something changes outside of "update". A prime example is look at swap rows for keyed. Look how many times it swaps rows.

Obviously JS execution is much cheaper than DOM ops but I tried this anyway. Replaced childNode references with firstChild/nextSibling as done in DomC and Stage0. As far as I can tell this has noticeable effect on any large table operation like insert 10k and append 1k on 10k. I removed a bunch of dead code that had been sitting in the file not doing anything, and built it with rollup instead of webpack like pretty much every other small implementation. 

Anyway, wanted to share. Whether you use it, see something you want to grab, like anything etc.. I tried to preserve it as close to the original as I could. I do feel the keyed implementation is cleaner as it was the idea I had coming in.  I got the feeling from the way it was written that it was based on the non-keyed and not vice versa. The non-keyed has the changes but it's more a reflection of what I did in keyed.

If you can bench it for comparison sake. I'm hoping the changes are noticeable, I just wasn't getting anywhere with little per op tricks like I've been submitting to date.

### DIFF
--- a/frameworks/keyed/vanillajs-1/index.html
+++ b/frameworks/keyed/vanillajs-1/index.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
+    <title>VanillaJS-"keyed"</title>
+    <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+<div id='main'>
+    <div class="container">
+        <div class="jumbotron">
+            <div class="row">
+                <div class="col-md-6">
+                    <h1>VanillaJS-"keyed"</h1>
+                </div>
+                <div class="col-md-6">
+                    <div class="row">
+                        <div class="col-sm-6 smallpad">
+                            <button type='button' class='btn btn-primary btn-block' id='run'>Create 1,000 rows</button>
+                        </div>
+                        <div class="col-sm-6 smallpad">
+                            <button type='button' class='btn btn-primary btn-block' id='runlots'>Create 10,000 rows</button>
+                        </div>
+                        <div class="col-sm-6 smallpad">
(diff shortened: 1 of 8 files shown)


leeoniya:
nice. will test later today.

probably worth removing all this jazz:

```
"@babel/core": "^7.1.2",
"@babel/preset-env": "^7.1.0",
"rollup-plugin-babel": "^4.0.3",
```

brodybits:
@krausest do you mean to keep both `vanillajs` and `vanillajs-1`?

I wouldn't mind having an extra VanillaJS framework for the extra comparison, if you want to do this deliberately.

krausest:
Thanks looks indeed significantly faster. I merged it as vanillajs-1 (great to make that comparison easier). I think I'll replace vanilla-js with that version some day. Regarding your observation regarding state handling. I remember how I hated creating the vanillaja version after working with react, angular and vue and how painful I found it getting all things straight. Very likely there's a better way to handle the state.

bildschirmfoto 2018-10-16 um 21 32 21

PR #473, opened 2018-10-14
Upgraded to Inferno v6

Havunen:
Inferno v6 JSX depends on Babel v7 so I had to update that also. I dont know what compilation level you want to use in this benchmark, it now goes with default `babel-env-preset`


krausest:
Thanks - has been merged:

PR #477, opened 2018-10-20
Fix Solid Select performance

ryansolid:
Updated Solid to 0.1.2.

Seriously this benchmark is the best. After seeing how strangely variable Solid results were on the 16x slowdown select I realized a pretty serious performance issue in my selection operator where I was re-evaluating the whole list on every select needlessly.


krausest:
Thanks for the kind words! 
Results in comparison (warning: potentially older chrome versions were used for older framework version) - pretty incredible!
bildschirmfoto 2018-10-27 um 21 43 38

PR #478, opened 2018-10-23
Update ivi to 0.16.0

localvoid:
It should significantly reduce memory consumption.


krausest:
Thanks! Results in comparisons (warning: potentially older chrome versions were used for older framework version)
Seems like it wasn't only about memory consumption:
bildschirmfoto 2018-10-27 um 21 43 38
bildschirmfoto 2018-10-27 um 21 47 20

PR #481, opened 2018-10-30
feat(angular): upgrade to 7.0

trotyl:
## Summary

+ Upgrade all Angular benchmarks to 7.0 version;
+ Remove duplicate `<body>` tags;
+ Remove unnecessary items in `.gitignore` files;
+ Refactor `rollup.config.js` format to align with current version of rollup;
+ Add module mapping for `@angular/compiler` until https://github.com/angular/angular/pull/26734 landed;

## Notes for Ivy implementation

+ Failed with closure compiler, switched to rollup for now, should have little difference in speed (but size would grow);
+ Now using real `@angular/common` library rather than custom picked `NgForOf` codes with `ngcc` tooling support;
+ Add a patch script in `postinstall` due to `ngcc` issues;


trotyl:
CI failure unrelated.

PR #486, opened 2018-11-13
add rawact

sokra:
https://github.com/sokra/rawact


krausest:
BTW here are the preliminary results for the non-keyed results:
bildschirmfoto 2018-11-14 um 22 14 28

sokra:
> Your version of replace rows is almost twice as fast as all other benchmarks: Great - or maybe not really...
> I think your keyed version is cheating, since it's not recreating all rows as it should for create rows and a keyed framework. Sorry, but I won't add it to the keyed section that way.

Yeah I wondered about that too, but thought it would be "ok" since the `isKeyed` check accepted it. I'll update the lib and this PR with a version not recycling DOM nodes.

krausest:
Sounds good. The isKeyed test wasn't good enough to identify your implementation as non-keyed. It just checked that at least one tr was added, I'll commit a better test soon.

sokra:
Updated

krausest:
Looks good now (in every way):
bildschirmfoto 2018-11-15 um 22 23 37
Startup looks very good. Did you expect those results for memory usage?
bildschirmfoto 2018-11-15 um 22 28 41

sokra:
Yeah memory usage is something I need to optimize. Rawact current creates closures for every element. That's a bit memory hungry. An idea is to only create a closure per type of element and use this everytime.

leeoniya:
@sokra you should also investigate why the mem stays high for `creating/clearing 1k rows (5 cycles)`. most libs allow the GC to reclaim it fully or partially, even if initial usage is high.

ryansolid:
Yeah when I've had high usage in both replace and creating/clearing and the numbers almost match, it's usually been because of a memory leak. Where the memory usage is directly proportional to the number of iterations. In this case 5.

PR #386, opened 2018-04-21
Fix reagent build in windows

alexfmpe:
Fixes #377 
Looks like leiningen needs an [explicit installation](https://github.com/technomancy/leiningen#windows) step on windows.
Only tested via [appveyor](https://ci.appveyor.com/project/alexteves13469/js-framework-benchmark/build/1.0.22) thus far, which unfortunately timeouts while running the benchmarks, so I haven't opened a PR for that.

PR #487, opened 2018-11-14
ci: test Windows

DanielRuf:
TravisCI has now support for Windows so we can also test there.

See https://blog.travis-ci.com/2018-10-11-windows-early-release

This PR is meant to be WIP as we have to check what we have to adjust in the CI config to get it running there too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 os:
+  - windows
   - osx
   - linux
 


alexfmpe:
Tried getting appveyor builds a while back but IIRC there was one framework that wouldn't work
See https://github.com/krausest/js-framework-benchmark/pull/386

DanielRuf:
As TravisCI is not used anymore I'll close this PR. in general TravisCI supports macOS, Linux and Windows. This would not require 2 platforms.

The Windows issue can be checked and probably solved.

alexfmpe:
Yeah there was a linux/mac Travis CI a ways back, but with a 1h timeout (whereas circleci had 5) - had to split the build into lots of jobs which increased the total time due to booting them all up. Might be more forgiving nowadays.

DanielRuf:
Here is a current overview for TravisCI environments: https://docs.travis-ci.com/user/reference/overview/

PR #488, opened 2018-11-14
ci: decrease Git clone depth

DanielRuf:
Cloning the last 700 commits seems a bit much or is there a (good) reason?

The last 5 commits should be sufficient and we get faster clone times.


krausest:
Merged - though I'm currently not using travis (and I'm neither seeing a lot of benefit from circleci).

PR #489, opened 2018-11-14
ci: use Node.js 8

DanielRuf:
Node.js 7 like every odd release branch is a short living release and has already reached its EOL.

Current LTS are 6, 8 and 10. 8 should be fine here.

See https://github.com/nodejs/Release/blob/master/README.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ env:
   - JS_FRAMEWORK=vue-v2.3.3-non-keyed
 
 language : node_js
-node_js  : 7
+node_js  : 8
 
 addons:
   chrome: beta


brodybits:
Is there any reason not to go for Node.js 10?

Node.js 10 will stay in LTS and maintenance mode longer, according to <https://github.com/nodejs/Release#release-schedule>.

DanielRuf:
Hm, seems CI is not enabled for PRs.

alexfmpe:
We changed to CircleCI a while back, it's running node 8.9.1 for now https://github.com/krausest/js-framework-benchmark/blob/46369a371ec4e8aae4a22d1b04a9f21731af1b0f/.circleci/config.yml#L17.
The travis file looks like it was forgotten behind

DanielRuf:
Ok so we should remove it then. But CircleCI also does not seem to be enabled for PRs.

What about changing it to just `8` to use the latest 8.x?

krausest:
I disabled CircleCI for PRs. It was too annoying to see it fail randomly due to network problems or other non PR related issues.
I have no problems to merge an updated travis-config if it helps someone though. Just let me know in case.

Issue #430, opened 2018-07-31
Question: Should we be setting any ground guidelines for what a Data Driven implementation is?


ryansolid:
I know this sort of thing has come up before (https://github.com/krausest/js-framework-benchmark/issues/230). But that's not my intention. When I [first submitted Solid](https://github.com/krausest/js-framework-benchmark/pull/398) I did several things unknowingly mimicking the vanillajs implementation not realizing it was generally seen as bad form. They might not be the type of considerations a VDOM library would consider, but in the range of data-driven libraries perhaps there is something here that should be said.

Roughly I'm thinking:

1.  A data driven library should provide at minimum:
    *  mechanism to update data
    *  automatically/manually trigger DOM updates
    *  provide a mechanism or abstract complexity of reconciliation of basic template control flow (conditionals, loops)

2.  An implementation should:
    *  At any at rest view state should be re-constructable from data only. (Ie. any atom of state cannot exist solely in DOM nodes at rest)
    *  Treat DIV.main as the mount/entry point. All DOM interactions within should be bound as part of the library's mechanisms.
    *  Not explicitly exploit Chrome's unclear performance boost using requestAnimationFrame to clear rows.

Is calling this out unreasonable? Is this too restrictive or specific?

ryansolid:
Similarly is it worth calling out reference implementations like VanillaJS, WC, etc.. since including them has value and they shouldn't be held to the same standards.

Havunen:
Yeah, this is a fair point. For example DOMc has html hard coded in index.html

https://github.com/krausest/js-framework-benchmark/blob/master/frameworks/keyed/domc/index.html

So it does not need to create the DOM which is major slow down in JS API. This makes comparison difficult.

leeoniya:
there are a lot of libs that are starting to get into questionable territory lately.

even for ivi, where @localvoid has historically held the philosophy of a straightforward, obvious implementation without "tricks" has drifted into using custom delegation logic and utilizing `textContent` pulled from the dom to improve ivi's numbers. :/

https://github.com/krausest/js-framework-benchmark/blob/master/frameworks/keyed/ivi/src/main.js#L86-L97

Freak613:
I'm totally fine to move reactive view into JS for the sake of consistency.

I just can't understand blaming with the words "hard coded" and "does not need to create the DOM".
@Havunen , can you provide some metrics that proves your point that inlined HTML provides any performance benefit for domc.
[...]

PR #490, opened 2018-11-17
Updates to solid, ko-jsx, mobx-jsx

ryansolid:
Pretty major change to the API of the underlying renderer where it is no longer automatically wrapping every expression. This prompts a new release for all 3 of solid, ko-jsx, mobx-jsx.  Some decent memory improvements for the first 2 but mobx-jsx doesn't seem to have really benefited from this update.

Also an even more significant update to solid to reduce API surface, and code, and increase compatibility with libraries like S.js and Surplus, at what I figure should be a noticeable performance change (it now wraps with ES6 proxies in all cases along the whole state tree which should have an overhead). As to the suggestions out of the setting a baseline discussion(#430) I'm attempting something similar to @localvoid  event delegation example API. I'm still sort of out on it, as it seems like it's good from an event bubbling perspective so the implementor doesn't need to trace a DOM tree, but I mean reading DOM events is so primary (like event.target.value to read from an input) that I'm not sure it makes sense to be a thing.

I've spent the past week using this benchmark to try to reduce memory and regain that perfomance so hopeful net isn't so bad.  This is one of those roll the dice commits in terms of performance, but I believe it's a big step forward for these libraries. I look forward to seeing the results.

### DIFF
--- a/frameworks/keyed/ko-jsx/package.json
+++ b/frameworks/keyed/ko-jsx/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-ko-jsx",
-  "version": "0.1.2",
+  "version": "0.2.0",
   "main": "index.js",
   "js-framework-benchmark": {
     "frameworkVersionFromPackage": "ko-jsx"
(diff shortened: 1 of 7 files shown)


krausest:
FYI: I'm a bit busy at the moment. I hope to merge it later this week.

Issue #428, opened 2018-07-30
Updating Datum failed


krausest:
Updating to 0.12.1 gives me a "Main.js?1221:22 Uncaught TypeError: Datum is not a constructor"
@MartinRixham Could you please take a look at it?

krausest:
@MartinRixham I'm thinking about retiring Datum. Any objections?

MartinRixham:
Hi, I would rather you didn't.

I've been meaning to fix this bug but have been quite busy, sorry.

krausest:
Thanks for you answer. I'll keep Datum. Would be nice if you could fix the issue.

PR #491, opened 2018-11-23
Update Datum to fix loading problem: issue #428

MartinRixham:


krausest:
bildschirmfoto 2018-12-05 um 20 19 45

MartinRixham:
no problem.

It would be interesting to see table of the genuine data binding libraries like Datum and Angular without all the jsx libraries.

PR #494, opened 2018-11-29
Update ivi to 0.20.0

localvoid:
- Removed all crappy "advanced" optimizations from the benchmark implementation that were discussed in [this issue](https://github.com/krausest/js-framework-benchmark/issues/430).
- Reconciliation algorithm were completely reimplemented from scratch.
- Composition model is now even more expressive than React composition model (https://codesandbox.io/s/k9m8wlqky3).
- Virtual DOM is now completely immutable.
- Fragments.
- Key namespaces to prevent collisions of keys.

This implementation should be slower in some test cases and consume more memory than the previous one mostly because it is now implemented in almost exactly the same way as react-redux implementation.

### DIFF
--- a/frameworks/keyed/ivi/package.json
+++ b/frameworks/keyed/ivi/package.json
@@ -16,12 +16,12 @@
     "build-prod": "rollup -c rollup.config.js"
   },
   "dependencies": {
-    "ivi-html": "0.16.0",
-    "ivi-state": "0.16.0",
-    "ivi": "0.16.0"
+    "ivi-html": "0.20.0",
+    "ivi-state": "0.20.0",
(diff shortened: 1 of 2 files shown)


krausest:
Thanks, especially for your first bullet point - let's hope it's a role model for the others!

Here are the results:
bildschirmfoto 2018-12-05 um 20 19 45

Have you tried the box plot view yet? I find it kind of interesting:
bildschirmfoto 2018-12-05 um 20 20 44
bildschirmfoto 2018-12-05 um 20 21 12

localvoid:
Thanks, yes, I've tried box plot view, it is awesome :+1:

PR #496, opened 2018-12-05
Update react-redux to 6.0

localvoid:
- Updated react-redux to 6.0.0, it is now using modern Context API
- Redux version is captured from `react-redux` package instead of `redux`


krausest:
Thanks. Here are the results:
bildschirmfoto 2018-12-05 um 21 28 54

PR #499, opened 2018-12-08
Cleanup Imba implementations

somebee:
Hey. Creator of Imba here. I cleaned up the Imba implementations. Since we're using immutable data here, I also added a simple shouldComponentUpdate-like check in row rendering.


krausest:
Thanks - results are here:
bildschirmfoto 2018-12-15 um 21 34 07

Issue #498, opened 2018-12-08
[Sticky] Number of paint calls is higher than expected


krausest:
Warning could be resolved for vanilla-dom-framework by making sure that the 
```
<span class="preloadicon glyphicon glyphicon-remove" aria-hidden="true"></span>
```
element is correctly included in the rendered html.

TODO: Check whether that applies for pux and reflex-dom wich are the other two frameworks with that warning

krausest:
See 2fda7e9f2d464d53e907a9d84f40741bcbedab24 for vanilla-dom-js, 01b07c23cde27cd4a5916ba7a62bd32ac0a1ad98 for pux.
Warnings are gone for these frameworks.

krausest:
@alexfmpe I think the warning for reflex-dom is also due to that missing span. Can you please add it?

krausest:
Looks good. No more warnings for those benchmarks. I'll leave the issue open for future reference.

mindplay-dk:
Why is there a cap on the number of paint calls?

It appears to cause problems with the render scheduler in Fre - see this:

https://github.com/yisar/fre/issues/271#issuecomment-907577159

The code seems to suggest this is a warning and not an error?

https://github.com/krausest/js-framework-benchmark/blob/dc4e9bbd2777431fc32d4e1876c14fa0d777c2ee/webdriver-ts/src/forkedBenchmarkRunner.ts#L206-L209

Not sure if this warning matters, or whether it's avoidable with Fre.

It's an interesting new framework with a fairly novel approach to rendering and scheduling - it would be nice if we could get this onboard with the benchmarks, but I wonder if this type of scheduling makes it inherently difficult to benchmark. 🤔

krausest:
This message mattered for older versions of chrome when more than a single paint event really meant that something didn't behave as expected. For newer chrome version this occurs fairly often, even for vanillajs like
```
# of paint events  2
duration to paint  134.088
duration to paint  137.612
```
The  benchmark uses the last paint event to compute the duration for an operation. I took a look at fre, which reports 3 paint events for create rows:
```
# of paint events  3
paint  1.79
duration to paint  153.488
duration to paint  155.631
```
This looks OK to me (and is what I see in the timeline). I think I'll remove the warning someday. Please ignore the warnings for the time being.

BTW the fre implementation isn't correctly keyed regarding create rows and swap rows:
```
npm run isKeyed -- --headless keyed/fre2
...
[...]

PR #500, opened 2018-12-10
Bump reflex-dom

alexfmpe:
Takes care of #498 for reflex-dom.
Turns out I had added that remove icon span a year ago, but it wasn't merged to my fork.

I was also getting some build errors on the previous version (probably due to either nix upgrade or mac os update). Should be fixed by this bump.

The 'Duration in milliseconds' geometric mean slowdown has gone from ~5x to over 10x between [round 8](https://stefankrause.net/js-frameworks-benchmark8/table.html) and the [current results](https://krausest.github.io/js-framework-benchmark/current.html) for some reason. I reproduced these strange results on my mac as well. Given that the reflex-dom benchmark hadn't been updated in many months, I suspect some chrome update might not get along with it. With this bump, the value seems to have gone back to ~5x slowdown.


krausest:
Thanks - results are here:
bildschirmfoto 2018-12-15 um 21 34 07

PR #501, opened 2018-12-10
Add simi wasm framework

limira:
Like other wasm frameworks, the generated files are already provided.


krausest:
Thanks - results are here:
bildschirmfoto 2018-12-15 um 21 36 42

PR #504, opened 2018-12-20
Add new framework: rax.

Lxxyx:
Repo: https://github.com/alibaba/rax


krausest:
Thanks. Here's a preview for the results:
bildschirmfoto 2018-12-20 um 20 43 59

PR #506, opened 2018-12-26
Optimize Ember app, to prevent creating unnecessary empty text nodes

simonihmig:
Seems to have a significant positive effect on render times (given the many iterations)

/cc @NullVoxPopuli

### DIFF
--- a/frameworks/keyed/ember/package.json
+++ b/frameworks/keyed/ember/package.json
@@ -44,6 +44,7 @@
     "ember-cli-update": "^0.27.0",
     "ember-decorators": "3.0.1",
     "ember-export-application-global": "^2.0.0",
+    "ember-hbs-minifier": "^0.3.0",
     "ember-load-initializers": "2.0.0",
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
@simonihmig, do you have screenshots? :(

simonihmig:
> do you have screenshots?

You mean of the results table, or of what?

For the results I just compared the files in `webdriver-ts/results` with previous runs without that change...

NullVoxPopuli:
Yeah, just something showing how much of a difference this PR makes?

simonihmig:
Some of the observed changes (though just with a test count of 5, so probably not very precise):

- run1k: 598 -> 460
- swap: 560 -> 527
- create10k: 3576 -> 2778

krausest:
Thanks. Indeed a bit faster. I noticed and removed an unnecessary style="cursor: pointer" on the first a-element in each row:
bildschirmfoto 2019-01-08 um 21 47 01

PR #507, opened 2018-12-30
Update Solid, ko-jsx, mobx-jsx

ryansolid:
Please merge this update to the latest version of these libraries.
- Reduced build size
- Centralized reconciliation for all libraries in renderer runtime, (based on Stage-0)
- Added support for JSX Fragments in List reconciliation
- Further reduced and streamlined Solid API
  - Changed to React Hooks inspired API
  - Added a compact Falcor inspired path API to Solid setState
- Removed many mostly useless micro optimizations from implementations that hurt readability/compactness
- Updated MobX implementation to use production mode.

I'm hoping none of this has a negative performance impact, but centralized reconciliation will lend to easier time of optimizing in the future.

### DIFF
--- a/frameworks/keyed/ko-jsx/package.json
+++ b/frameworks/keyed/ko-jsx/package.json
@@ -17,7 +17,7 @@
     "url": "https://github.com/krausest/js-framework-benchmark.git"
   },
   "dependencies": {
-    "babel-plugin-jsx-dom-expressions": "0.3.5",
+    "babel-plugin-jsx-dom-expressions": "0.3.6",
     "knockout": "^3.4.2",
     "ko-jsx": "0.3.0"
   },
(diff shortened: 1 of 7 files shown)


krausest:
Thanks. Looks good to me:
bildschirmfoto 2019-01-08 um 21 42 13

ryansolid:
Thank you that looks pretty level (I realize it's across different browsers but it's awesome you have the historicals for comparison). There might be something to investigate on the append 1000k. But I'm pretty stoked. Solid partial update looks to have gotten a little boost. 

I'm actually most interested to seeing if reducing the file size for the Solid implementation(~ 3kb) helped with the startup metrics (it's the one benchmark that doesn't seem to run under my Ubuntu on Windows setup since the update a few months back).

Looking forward to doing some comparisons with the results when the front page is updated :).

krausest:
Done - thanks for the reminder!

PR #510, opened 2019-01-10
Update lit-html to 1.0.0-rc.2

CaptainCodeman:
Sounds like this will be the last RC before release

Also updates build dependencies


krausest:
Looks good (though quite unchanged from rc.1):
bildschirmfoto 2019-01-17 um 21 52 23

PR #511, opened 2019-01-19
Added lighterhtml -> new Implementation of hyperHTML

mzeiher:
Hi, I added  lighterhtml to keyed frameworks


krausest:
Thanks - I hope to take a look at it the next few days.

mzeiher:
hi, thanks, I just updated lighterhtml to 0.7.0

mzeiher:
I'll take a look, lighterhtml should work like hyperhtml in keyed mode, I'll take a look at it

ryansolid:
Is there the ability to do a keyed version with lighterhtml? If so, highly recommended. Non-keyed is nice optimization in simple scenarios, but it isn't really representative of the performance of the library and is generally ignored during comparisons.

I know this is a bit of a selfish desire, but there has been a lot of push in this area(tagged template literal rendering) with lit-html reaching 1.0.0 and improving its performance quite substantially the past 6 months.  While I am not as familiar with ligherhtml I've always been a fan of Andrea Giammarchi's work, and would love to see what the latest and greatest from him can do head to head.

mzeiher:
I also really like the work of @WebReflection, the libs are easy to use and quite elegant, Maybe I could use the existing hyperhtml implementation for a keyed version, I'll take a look into it.

WebReflection:
@mzeiher give [lighterhtml](https://github.com/WebReflection/lighterhtml#lighterhtml) a try, maybe that's all you need 😉

WebReflection:
maybe I can help here, will give it a try 👋

WebReflection:
@ryansolid @mzeiher I have created what I believe would be recognized as keyed version of _lighterhtml_

https://codepen.io/WebReflection/pen/WPOJBo?editors=0010

please let me know if it makes sense to include it 👋

mzeiher:
Hi jep i'm already in favour for lighterhtml 😄, I also got confused about keyed and non-keyed, after re-reading https://www.stefankrause.net/wp/?p=342, keyed in this context means using the same node for the same data, thus reorganizing the DOM, @WebReflection the codepen you provided, behaves like a keyed implementation in the sense of the post above I would just include it in the PR as a keyed implementation for lighterhtml

WebReflection:
@mzeiher if that works, I'd like for you to wait for a more elegant implementation where all you have to change from your test is `html` to `html.for(item)` in the inline return.

This is a good improvement, in general, for the library. So thanks a lot for this PR 👍

WebReflection:
Last, but not least, and if it's not asking too much, or if it's even possible, I'd love to have _lighterhtml_ in both categories, used as keyed and unkeyed, for comparison sake ❤️

mzeiher:
@WebReflection wow that was fast, i'll add a version for keyed and non-keyed

mzeiher:
done, added keyed and non-keyed implementation of lighterhtml

mzeiher:
thanks @WebReflection for the fast keyed feature 👍

mzeiher:
@krausest the keyed/non-keyed issue is now fixed by providing an implementation for both.

WebReflection:
@mzeiher last minor aesthetic change could be adding:

```
context: 'null',
moduleContext: 'null',
```

to the rollup config so that no warning whatsoever due third parts features detection shows up.

example
```js
export default {
  input: 'esm/index.js',
  context: 'null',
  moduleContext: 'null',
  plugins: [
    resolve(),
    babel({
      runtimeHelpers: true,
      presets: ['@babel/preset-env']
    })
  ],
  output: {
    exports: 'named',
    file: 'index.js',
    format: 'iife',
    name: 'lighterhtml'
  }
};
```

krausest:
Results from a first run (last commit is not yet included):
bildschirmfoto 2019-02-05 um 19 44 03
@mzeiher @WebReflection Partial update and select row are very slow for keyed updates. Do you want to take a look at it or should it merge it anyways?
Non-keyed looks better (select rows is still a bit slow):
bildschirmfoto 2019-02-05 um 19 47 05

WebReflection:
@krausest thanks for testing it and publishing results. I'm pretty happy about the unkeyed result 'cause lighterhtml does much more than lit-html so being right there makes me happy.

It's also not so bad that a rushed keyed implementation is right beside React, but `v0.8.1` might score just slightly better, but not too faster than now.

The thing is, lighterhtml is optimized for "_don't care, I like magic_" use cases, where everything is handled in pseudo-keyed way behind the scene, so I wouldn't mind having the latest version in, even if it won't score too well compared to others lightweight solution.

I'm sure I'll manage to optimize it further, since the engine behind the scene shows already its performance on unkeyed tests.

Thank you!

WebReflection:
@mzeiher @krausest to be fair, the keyed demo could have an easy improvement to go 100% keyed just using `html.for(container)` as first entry point so that no unkeyed part of the code gets ever touched. That would make the demo purely keyed and, I believe, faster.

@mzeiher mind changing it? I've also pushed 0.8.3 but it's less relevant than having that single non keyed operation in the wild.

WebReflection:
@mzeiher the mentioned change is here https://github.com/WebReflection/lighterhtml/blob/master/test/keyed.html#L101

mzeiher:
jep, i'll add it!

WebReflection:
@mzeiher as we speak, I think that will score poorer than now. Please wait for my final confirmation of what should be changed. Apologies for the burden, I'm just too hyped to have lighterhtml in the mix.

WebReflection:
@mzeiher OK, with [that change], and version **v0.8.6**, everything should be fine. I think the benchmark now would score better than React, but if that's not the case, I stand positive in having this PR landed in this benchmark.

Thanks again everyone for your patience. If I'll score crap, it'll be on me and I'll do whatever it's possible to improve.

Best Regards

WebReflection:
Just **one more thing** ... I've noticed that competitors are running native ES2015 without any transpilation in place so I've [updated hyperHTML](https://github.com/krausest/js-framework-benchmark/pull/514) to reflect the same strategy.

The rollup config is easier, and I couldn't really see any benefit/improvement in using that `minifyHTML()` thing, but I thought it was maybe worth it mentioning that the build could be even easier than now (but I'm not asking to change anything, just saying things could be faster/simpler to build too 😉 )

WebReflection:
P.S. I've edited the lighterhtml version with the latest

mzeiher:
no worries :) can I then close this PR?

WebReflection:
@mzeiher not at all !!! this PR is about `lighterhtml`, not `hyperHTML`, these are two different libraries.

This PR should have test `lighterhtml` as dependency and use `html.for(container)` then it can be merged within this project.

mzeiher:
okok just got confused :D i'll update ASAP

mzeiher:
done, keyed seems a lot faster with `html.for(container)`

WebReflection:
> done, keyed seems a lot faster with `html.for(container)`

as quick explanation: before it was looping through all template literals holes, then I've patched in a way that was appending the whole thing once without looping through everything, then I've re-patched it so that a render will never re-append the same thing ever and only sub nodes are, eventually, modified 🎉

Thanks for the update 👍

WebReflection:
@krausest if you'd be so kind to merge this (and re-test it if you have time) that'd be awesome 🤟

krausest:
It's merged. Here are the last results (please keep in mind that most other frameworks were run on chrome 71 - I still need to update all of them). Keyed got a bit better - but from the conversation I suppose you expected even better results:
bildschirmfoto 2019-02-06 um 20 14 32
And here are the non-keyed results:
bildschirmfoto 2019-02-06 um 20 14 51

WebReflection:
thanks @krausest, it's above the line I was expecting but regardless a great reference for improvements 👍

PR #513, opened 2019-02-06
Update lit-html to 1.0.0

CaptainCodeman:
Released !


krausest:
Congrats! Here are the keyed results:
bildschirmfoto 2019-02-06 um 20 34 54

CaptainCodeman:
Sweet - thank you !

krausest:
@CaptainCodeman All results have been updated for Chrome 72. What's a bit puzzling is that Chrome 72 is significantly faster for create and replace while relative ranking for lit-html seems rather stable...

CaptainCodeman:
Thanks - I'll take a look into it and see if I can find out why.

PR #514, opened 2019-02-06
Update hyperHTML to 2.25.4

WebReflection:
Not only the current version of _hyperHTML_ is way outdated, I have also noticed that competitors are allowed to use native ES2015 syntax without any transpiled penalization.

Well, I guess it's time to compare keyed _hyperHTML_ for real 😃

### DIFF
--- a/frameworks/keyed/hyperhtml/package.json
+++ b/frameworks/keyed/hyperhtml/package.json
@@ -12,16 +12,11 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "hyperhtml": "2.13.0"
+    "hyperhtml": "2.25.4"
   },
   "devDependencies": {
-    "babel-core": "6.26.3",
-    "babel-plugin-external-helpers": "6.22.0",
-    "babel-preset-es2016": "6.24.1",
-    "rollup": "0.63.4",
-    "rollup-plugin-babel": "3.0.7",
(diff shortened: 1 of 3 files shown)


leeoniya:
> I have also noticed that competitors are allowed to use native ES2015 syntax without any transpiled penalization

out of curiosity, is this actually a thing? i've personally not seen ES5 run slower than ES2015 when transpiling with Rollup. maybe it's different with Babel?

krausest:
Thanks. Here are the results (once again - most other benchmarks were run on chrome 71):
bildschirmfoto 2019-02-06 um 20 48 16

WebReflection:
@leeoniya it is, just read this and check current hyperHTML full of unnecessary Babel bloat:
https://medium.com/@WebReflection/avoiding-babels-production-bloat-d53eea2e1cbf

`typeof`, `template literals`, all penalized by the transpilation.

WebReflection:
@krausest thanks gain. The engine is more or less the same as _lighterhtml_ so I am not expecting miracles, but I want to understand where such engine is causing bottlenecks, so an updated reality check is super good for me 👍

krausest:
@WebReflection FYI - I added a rule to the README some time (after some discussions)  ago that targeting the latest stable chrome is fine.

WebReflection:
@krausest thanks for the update, I should've followed closer debates

PR #515, opened 2019-02-07
Update lighterhtml test

WebReflection:

  * the keyed version of the test shouldn't create new objects each time otherwise the key gets lost for no reason
  * the click event prevents defaults and stops its propagation to grant nothing else is affected while benchmarking

### DIFF
--- a/frameworks/keyed/lighterhtml/src/index.js
+++ b/frameworks/keyed/lighterhtml/src/index.js
@@ -26,6 +26,8 @@ const clear = () => {
   _render();
 };
 const interact = e => {
+  e.stopPropagation();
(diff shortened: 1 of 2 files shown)


krausest:
Very nice! FYI - all other results were updated for chrome 72.
bildschirmfoto 2019-02-07 um 20 40 30

WebReflection:
Thanks! Just to clarify, you should see an improvement in the non keyed too, after the event changes I've made. However, my tests where showing max 0.01 difference again lit-html, not reflected in that screenshot.

I'll try to see if there are other improvements to make. 👋

PR #517, opened 2019-02-08
Update lighterhtml to 0.8.10

WebReflection:
First of all, thank you so much for your patience.
I have investigated a bit further possible bottlenecks when it comes to the selection, and the only thing that came up is that addressing `className` instead of dealing with a `class` `Attribute` node might place _lighterhtml_ slightly higher in the table (in terms of faster, so ... moving to the left).

With the promise that I won't change/PR anything else for a little while since there's literally nothing else I could optimize in the current _lighterhtml_ library (it's eventually a core change that would take much longer to land), please consider updating, for the last time, this library.

Thank you!

### DIFF
--- a/frameworks/keyed/lighterhtml/package.json
+++ b/frameworks/keyed/lighterhtml/package.json
@@ -24,13 +24,15 @@
   },
   "homepage": "https://github.com/krausest/js-framework-benchmark#readme",
   "dependencies": {
-    "lighterhtml": "0.8.7"
+    "lighterhtml": "0.8.10"
   },
   "devDependencies": {
-    "rollup": "1.1.2",
-    "rollup-plugin-commonjs": "^9.2.0",
-    "rollup-plugin-minify-html-literals": "^1.1.2",
+    "@babel/core": "^7.2.2",
+    "babel-plugin-remove-ungap": "^0.1.1",
+    "rollup": "^1.1.2",
+    "rollup-plugin-babel": "^4.3.2",
     "rollup-plugin-node-resolve": "^4.0.0",
-    "rollup-plugin-terser": "4.0.2"
+    "rollup-plugin-replace": "^2.1.0",
+    "rollup-plugin-terser": "^4.0.4"
   }
 }
--- a/frameworks/keyed/lighterhtml/rollup.config.js
+++ b/frameworks/keyed/lighterhtml/rollup.config.js
@@ -1,35 +1,32 @@
+import resolve from 'rollup-plugin-node-resolve';
+import replace from 'rollup-plugin-replace';
+import babel from 'rollup-plugin-babel';
 import { terser } from 'rollup-plugin-terser';
-import minifyHTML from 'rollup-plugin-minify-html-literals';
-import commonjs from 'rollup-plugin-commonjs';
-import nodeResolve from 'rollup-plugin-node-resolve';
 
 export default {
-    input: `src/index.js`,
-    context: 'null', 
-    moduleContext: 'null',
-    output: {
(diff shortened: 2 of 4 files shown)


WebReflection:
P.S. I have also optimized the bundling and the amount of needed modules

WebReflection:
P.S.2 I have also managed to do some cleanup and avoid using unnecessary boilerplate used to make the library resistant against TypeScript. If it's performance that we want compare here, I guess it makes sense to compare these for the real environment it's targeting, which is not TS.

The bumped version is only because 0.8.10 also has an `es.js` file within its bundle, but nothing actually changed from 0.8.8, just the bundling part (0.8.9 had a typo).

If you want I can put the version back to 0.8.8 but what interests me is that finally I have a pure ES6 version of the library to compare, some as competitors, so that numbers will be more meaningful.

krausest:
Absolutely no problem - I merged 0.8.10.
But I'm afraid it seems like 0.8.10 didn't score faster than 0.8.7. I ran it two times to make sure that the measurement wasn't bad but results were pretty close:
bildschirmfoto 2019-02-08 um 21 38 23
bildschirmfoto 2019-02-08 um 21 38 57
Don't hesitate to push further PRs if you think you found some optimisations.

WebReflection:
Why my benchmark isn't like that? 🤔
Anyway, that looks like a slight improvement anyway, but something ain't right with our setup

WebReflection:
@krausest I really have different results in here ...

screen shot 2019-02-08 at 22 36 38

screen shot 2019-02-08 at 22 36 51

screen shot 2019-02-08 at 22 36 59


How can we be sure this benchmark is reliable ?

krausest:
Well, just to make sure I'll check a few other frameworks tomorrow to see whether one of the security updates (Chrome updated to 72.0.3626.96 yesterday or the OSX update from today) had an impact.
What's your configuration?

WebReflection:
not sure how to give you all details about my configuration, but this is the result on my machine:
https://twitter.com/WebReflection/status/1093988455292588033

screen shot 2019-02-08 at 22 44 18

screen shot 2019-02-08 at 22 44 58

WebReflection:
P.S. I know you're not new to this but hard refreshing results gave me latest benchmarks, 'cause the server is kinda greedy with cache in here 😉

WebReflection:
P.S. have you also `npm run build-dev` before running the test? 🤔

WebReflection:
I've done another round of tests, in case this helps verifying results are somehow aligned.

## Keyed
screen shot 2019-02-09 at 11 23 19
screen shot 2019-02-09 at 11 23 33
screen shot 2019-02-09 at 11 23 42

## Non Keyed
screen shot 2019-02-09 at 11 23 51
screen shot 2019-02-09 at 11 24 01
screen shot 2019-02-09 at 11 24 09

krausest:
> P.S. have you also `npm run build-dev` before running the test? 🤔

That's not necessary.

Numbers on your machine look pretty consistent across runs - looks fine to me (Have you seen any suspicious results?). Your smaller but newer MBP seems to be indeed a bit faster than my old MBP.

WebReflection:
> Have you seen any suspicious results?

Just somehow a huge difference once I've minified the template literal HTML, but that's probably not suspicious at all.

https://github.com/krausest/js-framework-benchmark/pull/518

ryansolid:
Yeah the empty text nodes that appear when you don't remove whitespaces add a decent amount of overhead for rendering. This has been noticed in a number of implementations, most noticeably recently with Ember minifying the html template actually improved the insert 10k rows by more than 500ms.

PR #518, opened 2019-02-09
Update lighterhtml bundle + minifyHTML

WebReflection:
Turns out that minifying HTML is somehow key to score better at this benchmark.

I know I've promised I wouldn't have changed the library, and indeed I just changed the rollup config.

At this point there's really nothing else I can do, except eventually find out slow paths and improve, but for now I just want a fair comparison with the major competitor.

Thanks again for your patience.

screen shot 2019-02-09 at 12 42 44

### DIFF
--- a/frameworks/keyed/lighterhtml/index.html
+++ b/frameworks/keyed/lighterhtml/index.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <meta charset="utf-8" />
-<title>lighterhtml-"non-keyed"</title>
+<title>lighterhtml keyed</title>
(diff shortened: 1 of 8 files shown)


krausest:
Indeed - it's faster. 
Here's a comparison between this PR (labled lighterhtml-v518, a version that has just the minifyHTML (in rollup.config.js) removed  labeled lighterhtml-v518-noMinify and the old version labeled lighterhtml-v0.8.10)
bildschirmfoto 2019-02-10 um 13 00 51

I've updated the results (just in case I confused someone with the versions above: On the github results page the lighterhtml-v0.8.10 version is now the one with this PR applied.)

PR #520, opened 2019-02-11
Update lighterhtml to 0.9.0

WebReflection:
My deepest apologies but I've realized, after focusing too much on the keyed result, that non-keyed stack was broken, but it has been, fixed right now via https://github.com/WebReflection/lighterhtml/issues/15

Please don't hate me ❤️

### DIFF
--- a/frameworks/keyed/lighterhtml/package.json
+++ b/frameworks/keyed/lighterhtml/package.json
@@ -24,7 +24,7 @@
   },
   "homepage": "https://github.com/krausest/js-framework-benchmark#readme",
   "dependencies": {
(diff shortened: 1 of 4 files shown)


krausest:
> Please don't hate me ❤️

For a PR without intricate discussions? I promise I won't. Results have been updated and are very close to what we had.
bildschirmfoto 2019-02-13 um 21 20 57

bildschirmfoto 2019-02-13 um 21 19 46

WebReflection:
FWIW, 0.01 matters here, as there's only domvm-v3.4.5-non-keyed in between now 😉

PR #523, opened 2019-02-14
Simplify lit-html benchmark code (without direct DOM manipulation)

aomarks:
This is the same as PR https://github.com/krausest/js-framework-benchmark/pull/521, but without the controversial change involving directly setting a class on the table row.

A few simplifications to the lit-html benchmark implementations:

* Minor style updates (semicolons etc.).
* Update `select` and `label` properties in-place and remove `guard` directive, instead of creating new row objects every time.
* Remove some unnecessary attributes and use `closest()` to find the action and row id.
* <strike>Directly update row class on select instead of forcing an unnecessary render.</strike>

cc @justinfagnani @rictic @CaptainCodeman

### DIFF
--- a/frameworks/keyed/lit-html/src/index.js
+++ b/frameworks/keyed/lit-html/src/index.js
@@ -30,47 +30,47 @@ const clear = () => {
 const interact = e => {
   const interaction = e.target.getAttribute('data-interaction');
   const id = parseInt(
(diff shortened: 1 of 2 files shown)


krausest:
Thanks a lot - results have been updated. Here's a screenshot:
bildschirmfoto 2019-02-18 um 20 15 49

Issue #522, opened 2019-02-14
New rule: Prevent direct DOM access


krausest:
Please vote "thumbs up" for the formulation of the new rule that should be added to the benchmark or propose a better formulation:

krausest:
The implementation must use only the idiomatic style of its library (e.g. direct dom access is not permitted).

krausest:
Proposal 2:
The initial rendering must be able to render the selection state.
(Question: Is that sufficient to prevent reaching into the dom with the initial state?)

krausest:
Proposal 3:
Don't use shortcuts in your implementation. Using explicit event delegation, manipulating the dom directly or using direct request animation frame calls are considered bad smells and may lead to rejection.

WebReflection:
@krausest mind expanding on the explicit event delegation? Both lit/lighterhtml use the row to intercept once internal events, while hyperHTML, as example, places listeners right on the node that should trigger the action.
Wich implementation is correct, if not both?

WebReflection:
P.S. I like all proposals but I'm not sure I get the last one

krausest:
@WebReflection 
The idea with event delegation came from #430, but maybe it's not really well put. I'd consider it an illegal shortcut if e.g. a react implementation used userland code to add the select row event listener to the table. 

Maybe this way:

Proposal 4:
Use the idiomatic style of the library and avoid userland shortcuts in your implementation like userland dom manipulations, userland request animation frame calls or userland event delegation. These are considered bad smells and may lead to rejection.

leeoniya:
yeah, there are a bunch of libs that delegate and dom-climb:

https://github.com/krausest/js-framework-benchmark/search?q=parentNode&unscoped_q=parentNode

WebReflection:
@krausest I like proposal 4 the most 👍

leeoniya:
> (e.g. direct dom access is not permitted).

this part is gonna be the most controversial. i feel like outright disallowing it is a bit heavy-handed, but policing the details gets messy in a hurry. with the right abstraction, it could probably be acceptable as long as it's generalizable for other applications and is uniform for create *and* update. for example, if the framework had a way to notify of the granular necessary `textContent` changes but left the actual mechanics down to a subscriber/driver. maybe if there is a framework that allows you to bring your own rendering driver. like if core React lib was the submitted lib and react-dom (the "driver") was just part of the impl.

adamhaile:
This conversation has gotten so far into the weeds.

[...]

PR #524, opened 2019-02-17
Update Solid, KoJSX, and MobXJSX v0.4.0

ryansolid:
Libraries are going to take a little bit of a hit on raw rendering with this one. Lots of feature enhancements, no specific focus on performance for this benchmark. Highlights:

* Full Support for JSX Fragments/Array in Control Flow, Direct Nested Multi-Level
* Control Flow as top level node and children expression wrapping optional
* Multi-mode Reconciler Add-on to do data diffing to support large data snapshots (DBMon, UIBench)
* Non-JSX option: HyperScript syntax run-time wrapper (although limited usefulness as skips compiler which does most of the heavy lifting)
* Implicit Event Delegation(I think, I don't even know anymore), with semi-synthetic events, support for composed Shadow DOM events and Custom Events. Follows Inferno convention of camelCase -> delegate and lowercase -> direct native bound.

I'm hoping this will be enough to address #522 as far as my libraries are concerned. Admittedly I'm a lot happier how unobtrusive this event delegation is. I guess see how this goes and try more specific optimization after.

### DIFF
--- a/frameworks/keyed/ko-jsx/package.json
+++ b/frameworks/keyed/ko-jsx/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-ko-jsx",
-  "version": "0.3.0",
+  "version": "0.4.0",
   "main": "index.js",
   "js-framework-benchmark": {
     "frameworkVersionFromPackage": "ko-jsx"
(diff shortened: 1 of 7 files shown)


krausest:
bildschirmfoto 2019-02-18 um 20 13 13

ryansolid:
Yeah exactly about what I expecting.. Bit more costly on the teardown, but the raw render wasn't as hit. Thank you.

PR #516, opened 2019-02-07
Update sifrr to 0.0.2

aadityataparia:

PR #526, opened 2019-02-19
Fix rule violation in sifrr

aadityataparia:
From https://github.com/krausest/js-framework-benchmark/pull/516#issuecomment-464866673

### DIFF
--- a/frameworks/non-keyed/sifrr/elements/main/element.html
+++ b/frameworks/non-keyed/sifrr/elements/main/element.html
@@ -54,7 +54,8 @@
     for (let i = 0; i < count; i++)
       data.push({
         id: i + from,
-        label: adjectives[_random(adjectives.length)] + ' ' + colours[_random(colours.length)] + ' ' + nouns[_random(nouns.length)]
+        label: adjectives[_random(adjectives.length)] + ' ' + colours[_random(colours.length)] + ' ' + nouns[_random(nouns.length)],
+        class: ''
       });
     from = from + count;
     return data;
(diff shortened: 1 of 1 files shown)


aadityataparia:
@krausest Please 🙏

aadityataparia:
@krausest I have a question regarding select:

### Before
- Save selected element directly, and update on change of selected element 
This is a no go according to rule 👍 

### Current
- Loop through all the rows and select row which user selected and deselect any previously selected row.

### Question
But will this be considered a rule violation:
- Save selected row id, and deselect old selected row if and select new selected row id?

StefansArya:
It seems you have fall in the same hole [like me](https://github.com/krausest/js-framework-benchmark/pull/519#issuecomment-462603177) :smile: 

The purpose of the test is for measuring the framework speed, and not the direct DOM manipulation speed. I think it's okay to save the id of the item, but you must apply any changes with the framework itself..

krausest:
@aadityataparia You mean like [here](https://github.com/krausest/js-framework-benchmark/blob/master/frameworks/keyed/lit-html/src/index.js#L45-L49)?
If you save it in the model and use it for initial rendering you should be safe.

ryansolid:
The only thing I'd consider with such a solution is ensure that swap still works as intended (if the selected row was the one being swapped). I'm not as familiar with if it is different in non-keyed but make sure the functionality works the same.  While there aren't testing for all the edge cases you want to make sure your implementation makes sense from how a user would expect interactions to work on a list in which items can be re-ordered and selected at will.

aadityataparia:
@krausest Saving it then,

But there's one thing I noticed that if you save the index of selected row, and then delete a row above this row, and then select some other row, it won't deselect old row since the saved index is not correct anymore. 🤷‍♂️

krausest:
Thanks. Results have been updated!

PR #528, opened 2019-02-20
Fix build error for ScarletsFrame

StefansArya:
About this pull request:
- [x] Fix error when compiling the framework with babel because unsupported regular expression.
- [x] Update the framework, test implementation, and the title for non-keyed test
- [ ] Fix [hangs](https://github.com/krausest/js-framework-benchmark/pull/519#issuecomment-464855788) for the test driver

BTW, how to check the the test driver about the hangs?
I'm not sure it was solved or not because I don't see anything wrong when I test it with my computer.


krausest:
It still hangs without the patch. You can try it if you follow the readme and run in webdriver-ts `npm run compile` and then `npm run bench -- --framework  scarletsframe-keyed`.
Just remove https://github.com/krausest/js-framework-benchmark/blob/master/webdriver-ts/src/forkedBenchmarkRunner.ts#L476-L479 and recompile and test again. It only works for the first run and hangs for subsequent ones unless the wait condition is present.

PR #529, opened 2019-03-02
Add ReasonReact implementation

michaelboulton:
This is a (mostly) idiomatic implementation in ReasonML using ReasonReact.

- It uses the `option` type for seeing if a row is selected rather than just having 0 be a magic number for 'nothing selected', or just never resetting it if a row gets deleted like most other implementations
- I did implement the "update every 10th row" using an inplace method but it felt like a bit of a hack so I've left it in there but commented out.
- The `Row` component had to be a `reducerComponent` even though it is pretty much stateless - this is so that it can properly use `shouldUpdate` rather than rerendering all the time

The JSX (and the webpack config) is pretty much just copied from the various react implementations and changed to fit with the Reason syntax. Performance seems to be roughly the same as react-redux, which is about what you would expect.

### DIFF
--- a/frameworks/keyed/reason-react/.gitignore
+++ b/frameworks/keyed/reason-react/.gitignore
@@ -1,0 +1,3 @@
+lib
+**/*.bs.js
+.merlin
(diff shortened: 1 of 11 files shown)


krausest:
Thank you. I‘ll review and merge it right after my holidays (maybe this weekend).

krausest:
Results are in now:
bildschirmfoto 2019-03-07 um 20 46 50

PR #530, opened 2019-03-04
Adds React Hooks implementation

ryansolid:
Been playing around with Hooks a lot recently.  Super interested how this compares and if we can gain any performance insights. Would appreciate some review.

* I went with single state object since data and selected are mostly updated in unison.
* I used useCallback to only have the single fn ref on appendRows (when rows are added in different renders), but I wouldn't be surprised React.memo is slightly faster on its own.
* I didn't bother splicing array on update, not correct but works. Noticed React implementation doesn't either.
* Did a few less optimized patterns to feel idiomatic like react-redux implementation

### DIFF
--- a/frameworks/keyed/react-hooks/index.html
+++ b/frameworks/keyed/react-hooks/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <title>React Hooks</title>
+  <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+  <div id='main'></div>
+  <script src='dist/main.js'></script>
+</body>
+</html>
--- a/frameworks/keyed/react-hooks/package.json
+++ b/frameworks/keyed/react-hooks/package.json
@@ -1,0 +1,38 @@
+{
+  "name": "js-framework-benchmark-react-hooks",
+  "version": "1.1.1",
+  "description": "React Hooks demo",
+  "main": "index.js",
+  "js-framework-benchmark": {
+    "frameworkVersionFromPackage": "react"
+  },
+  "scripts": {
+    "build-dev": "webpack --watch",
+    "build-prod": "webpack"
+  },
+  "keywords": [
+    "react",
+    "webpack"
+  ],
+  "author": "Stefan Krause",
+  "license": "Apache-2.0",
+  "homepage": "https://github.com/krausest/js-framework-benchmark",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/krausest/js-framework-benchmark.git"
+  },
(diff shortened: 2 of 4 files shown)


krausest:
Thanks, surely very interesting. I‘m on holidays this week and will merge it afterwards.

thysultan:
@ryansolid Can we move `run`, `runLots`, `add`, `update`, `clear`, and `swapRows` into the memo  `Jumbotron`?

ryansolid:
Is it good to pass naked setState into the child component? The state to be updated needs to owned by the main component. I've done similar in other fine grained libraries, but I'm unsure if it makes sense for React idiomatically.

I have no problem making this change, if you think that it makes sense.

thysultan:
>Is it good to pass naked setState into the child component?

I'm unsure, my assumption is that since `setState` uses `useReducer` internally, it would be the same across renders, though this could just be a "current" implementation detail and may very well change, maybe @gaearon could comment in this.

gaearon:
Generally `useReducer` + passing `dispatch` down via context is the recommended optimization approach. See https://github.com/ryardley/hooks-perf-issues/pull/3.

ryansolid:
Ok, hard to argue with that.  I have updated the implementation. I might be over using useCallback and I'm banking on the fact the item.id never changes in the closure.

It does produce an interesting refactoring story when a user might start with setState, and then decides to delegate the commands to a child that is mostly static. I see this pattern happening fairly often as an optimization method.

krausest:
Here are the results. All react contenders on the screenshot were re-run for a better comparison:
bildschirmfoto 2019-03-07 um 20 46 50

ryansolid:
That's awesome,

My hunch was that that hooks should be the same or slightly better in performance and this confirms it. If even by a hair, Hooks are the most performant React variant.

But this is a really good demonstration of how optimized approaches for both React Hooks and Classes differ (looking at memory and performance characteristics).

Thanks @gaearon for the clarification on the recommended approach.

tdd:
@gaearon any idea why swapping rows is so horrendously slow regardless of the approach? This is keyed, shouldn't that end up being just two `insertBefore` calls after a quick array traversal with referential comparison of keys?

gaearon:
I think swapping two rows at the beginning and end is the worst case for React because it will `insertBefore` everything else. In this case React tries to be more predictable instead of unexpected perf jumps when a smarter heuristic didn't apply neatly. We could change but in practice we've rarely seen people care about in outside of benchmarks. If you're rendering 1000 rows and reorder them you should use virtualization in which case it's essentially free.

tdd:
I was unconvinced by the main rationale (i.e. I failed to see why having keys didn't let us only move the two relevant lines DOM-side), then the virtualization note basically killed the whole argument 😁

PR #531, opened 2019-03-05
Add sifrr keyed and upgrade sifrr to 0.0.3

aadityataparia:
I hope it is okay to use symlinks.


krausest:
Symlinks work fine on my OSX machine, but I checked on a windows machine and got:
```
D:\stef\Source\js-framework-benchmark\frameworks\keyed\sifrr>npm run build-prod
npm ERR! file D:\stef\Source\js-framework-benchmark\frameworks\keyed\sifrr\package.json
npm ERR! code EJSONPARSE
npm ERR! JSON.parse Failed to parse json
npm ERR! JSON.parse Unexpected token . in JSON at position 0 while parsing near '../../non-keyed/sifr...'
npm ERR! JSON.parse Failed to parse package.json data.
npm ERR! JSON.parse package.json must be actual JSON, not just JavaScript.

npm ERR!     C:\Users\Stefa\AppData\Roaming\npm-cache\_logs\2019-03-09T18_52_16_645Z-debug.log
```
So I suggest going without symlinks (especially since I'm thinking about going back to windows). I'll just make a copy of the symlinked files.

aadityataparia:
😢 Maintaining two same directories will be a pain, as the only difference between two implementations is just one attribute :(

krausest:
There would be another option for maintaining in only once - but I guess you don't want to hear about it (#304) :wink:
I'll make a copy of the linked files as the other implementations also do.

krausest:
Here are the results:
Bildschirmfoto 2019-03-14 um 20 31 59

aadityataparia:
Thanks! But did something went wrong while running, because I don't see results event close to this when I run locally? Maybe a bug in 0.0.3, need to investigate!

PR #534, opened 2019-03-16
EndorphinJS

BusinessDuck:
Added keyed and non keyed new UI Rendering Library https://github.com/endorphinjs

### DIFF
--- a/frameworks/keyed/endorphin/index.html
+++ b/frameworks/keyed/endorphin/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Endorphin</title>
+    <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body >
+    <div id="main" class="container">
+    </div>
+    <script src='dist/main.js'></script>
+</body>
+</html>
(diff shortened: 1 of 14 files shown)


BusinessDuck:
Demo results: [https://i.ibb.co/tbbCz4Z/Screenshot-2019-02-28-at-18-28-40.png](url)

ryansolid:
From your demo results there is a pretty reasonable chance your keyed version is not keyed by this repo's definition for replace rows. Replace rows should not recycle dom nodes which is likely why your preliminary results have that test significantly below VanillaJS.

BusinessDuck:
I am sure, its my mac issues... Test are not clean. I have work in one time. Now we will try to run full production build of all tests in our selenium node (vpn server), than result will be more cleany.

krausest:
I think the keyed behaviour is fine. Here are some results:
Bildschirmfoto 2019-03-20 um 22 08 22
Please don't compare against other frameworks right now since chrome 73 seems much slower for partial update and I need re-run all frameworks (vanillajs is already on chrome 73).

ryansolid:
Yeah it was likely just a coincidence that the inconsistency of the demo results were the exact same tests and only those tests that would be affected by running non-keyed. The final results look proper.

PR #535, opened 2019-03-17
Minor fix to reason-react implementation

michaelboulton:
This probably won't affect the results at all because it only happens after you update every tenth row _then_ do something, but this fixes a bug where it would always re-render every tenth row after it had been updated.


krausest:
Done! Results must be taken with care this time. Seems like partial updates got much worse with chrome 73 (vanillajs got much slower for this benchmark too, I need to re-run all other frameworks)
Bildschirmfoto 2019-03-20 um 21 45 07

PR #537, opened 2019-03-19
Update RE:DOM

pakastin:
Update RE:DOM to the latest version (3.19.0) + dependencies


krausest:
Done! Results must be taken with care this time. Seems like partial updates got much worse with chrome 73 (vanillajs got much slower for this benchmark too, I need to re-run all other frameworks)
Bildschirmfoto 2019-03-20 um 21 45 07

pakastin:
No worries, I'm not that picky about the results (at least not anymore) 😄

PR #539, opened 2019-03-23
Update RE:DOM to v3.20.0

pakastin:
I think this might make a difference: https://github.com/redom/redom/pull/144 😎


krausest:
Thanks. Results have been updated. CPU performance seems unchanged though, startup main thread work cost seems lower. Where did you hope to see improvements?

pakastin:
Just realized elements are memoized and cloned, so startup improvement makes most sense yeah 👍

PR #541, opened 2019-03-29
Add djinn-state

gilmarsquinelato:
Hello @krausest!
Thanks for the great job!
I've added a benchmark of djinn-state with react.


krausest:
Thanks - results have been updated. Here's a screenshot:
Bildschirmfoto 2019-04-11 um 20 58 01

PR #542, opened 2019-04-03
Update san

errorrik:

- update san version to 3.7.3
- add keyed test
- update swap and select implementation of non-keyed


krausest:
Great - a keyed version (and the keyed got faster too). Results have been updated:
Bildschirmfoto 2019-04-11 um 20 58 23

errorrik:
Thank you @krausest .

And, how about non-keyed?

krausest:
Should be published in the result table:

errorrik:
(*˘︶˘*).｡.:*♡

PR #550, opened 2019-04-13
Faster-dom - Optimize call, use version with fragment support

PxyUp:
@krausest Hi, was optimize render list(start use fragment) and optimize class supports and remove some unused stuff

### DIFF
--- a/frameworks/keyed/faster-dom/package.json
+++ b/frameworks/keyed/faster-dom/package.json
@@ -28,7 +28,7 @@
     "webpack-cli": "3.2.3"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


krausest:
It seems to be broken. Click twice create 1,000 rows and then try to remove a row (or selecting a row) which doesn't work on my machine.

PxyUp:
@krausest Thank you very much, was fixed in new version. MR was updated

krausest:
Results have been updated. Here's a screenshot:
Bildschirmfoto 2019-04-22 um 18 36 32

Issue #549, opened 2019-04-13
[Sticky & fixed] It looks the performance logging options affect the performance of frameworks.


ismail-codar:
I'm working on a new framework called [fidan](https://github.com/ismail-codar/fidan).
The actual result on the right according to the screenshot below. Create rows benchmark is slow.

When I investigated the cause of the slowdown I saw that it was related to https://github.com/krausest/js-framework-benchmark/blob/master/webdriver-ts/src/forkedBenchmarkRunner.ts#L342. If I change this line to`traceCategories: "devtools.timeline,blink.user_timing"`, I see the result on the right. Pretty fast.

I don't know why, but there seems to be something wrong.

Additional note: When I logging timings to the console from http://127.0.0.1:8080/frameworks/keyed/fidan, I see fast results similar to those on the left.

<img width="470" alt="fidan" src="https://user-images.githubusercontent.com/282104/56076815-d5584c00-5ddd-11e9-8320-b87954b68480.png">

krausest:
That difference looks worrying!

As a first quick check I ran a single create rows benchmark for a few frameworks. I got the following results:
<img width="276" alt="Bildschirmfoto 2019-04-13 um 13 08 00" src="https://user-images.githubusercontent.com/3374055/56078874-2facd900-5ded-11e9-9c39-4c0780198bc7.png">
Could indeed be that the old code is slightly slower than with your change (but I need more measurements to make sure this isn't due to some background process and last published results are also a bit faster), but it's nowhere close to the difference you observe. 
What machine and platform are you running? 
Is the difference for other frameworks similar large on your machine?
Could you please post a PR for fidan such that I can take a look at the difference on my machine?

ismail-codar:
For example, I saw that vanillajs was no different. But there is a big difference in my framework.  My framework is functional reactive style so maybe related to it or something else.  I will prepare PR. Thanks for the your great working.

ismail-codar:
Hi @krausest  I just sent a PR. 
Please note: Running it from http://localhost:8080/frameworks/keyed/fidan/ looks fast (console logging) similiar to changing [this line](https://github.com/krausest/js-framework-benchmark/blob/master/webdriver-ts/src/forkedBenchmarkRunner.ts#L342) to `traceCategories: "devtools.timeline,blink.user_timing"`

krausest:
I got the PR from İsmail and ran the create 1,000 rows benchmark for all frameworks. Here's the comparison:
[...]

PR #551, opened 2019-04-13
Adding fidan-v0.0.12-keyed

ismail-codar:
Under development [fidan](https://github.com/ismail-codar/fidan) is quite innovative. Benchmarks are works fine and generally fast.  But there is an issue: https://github.com/krausest/js-framework-benchmark/issues/549


krausest:
A big thanks for spotting this issue and reporting it!
I took a look at fidan. I'm afraid to say it's currently not properly keyed for replace rows:
```
Stefans-MBP:webdriver-ts stefan$ npm run isKeyed -- --framework fidan

> webdriver-ts@1.0.0 isKeyed /Users/stefan/Source/Javascript/js-framework-benchmark/webdriver-ts
> cross-env LANG="en_US.UTF-8" node dist/isKeyed.js "--framework" "fidan"

Frameworks that will be checked fidan-v0.0.6-keyed
fidan-v0.0.6-keyed is non-keyed for 'run benchmark' and keyed for 'remove row benchmark' keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
```
It currently recycles the rows when replacing all rows (you can easily check if you change the background-color for a row with the devtools. Then replace all rows. The background color is still there). This is allowed for non-keyed frameworks, but not for keyed.

ismail-codar:
Yeah, that's my fault. I just added `renderMode: "reconcile"` option for array rendering. 
https://github.com/krausest/js-framework-benchmark/pull/551/files#diff-163024d903e225a027d6060c95904ccbR225 So reconcile function given from stage0 thanks for ivi, surplus, stage0 authors.

krausest:
Thanks - looks correctly keyed now. Results have been updated. Here's a screenshot:
Bildschirmfoto 2019-04-22 um 18 36 32

ismail-codar:
Thanks.

PR #552, opened 2019-04-20
Updated literaljs version and updated benchmark code for literaljs.

Mikeysax:


krausest:
Seems much slower than 5.1.62. I measured twice and checked no background process is running wild.
Any idea why it's so much slower?
Bildschirmfoto 2019-04-22 um 18 37 11

Mikeysax:
@krausest Interesting, I'm not entirely sure. That's a drastic difference... I will need to do more testing I believe, in order to clarify why this has happened.

PR #482, opened 2018-11-08
Ember upgrades to 3.5

NullVoxPopuli:
For my own reference:

```
# Terminal 1
cd ./frameworks/keyed/ember
yarn && yarn build-prod
```

```
# Terminal 2
cd ./frameworks/keyed/ember-sparkles
yarn && yarn build-prod
```

```
# Terminal 3
yarn && yarn start
```

```
# Terminal 4
cd ./webdriver-ts
yarn && yarn build-prod
yarn selenium --count 3 --framework ember
yarn selenium --count 3 --framework ember-sparkles
```

```
# Terminal 5
cd ./webdriver-ts-results
npm install && npm run build-prod
```

```
# Terminal 6
cd ./webdriver-ts
yarn results
```

visit http://localhost:8080/webdriver-ts-results/table.html

Results:
image
image
image

looks like there are some improvements over 3.4.1
(also, ignore the 3.4.1 title on sparkles, I forgot to change the package.json referenced framework version)

PR #554, opened 2019-04-26
Ember keyed upstream to 3.9.1

lifeart:
image
image
image

prev results: https://github.com/krausest/js-framework-benchmark/pull/482


frameworks/keyed/ember/app/templates/components/bs-button.hbs
```diff
@@ -1,7 +1,7 @@
 <button
   type="button"
-  class="btn btn-primary btn-block" id={{@id}}
-  onclick={{action @onClick}}>
+  class="btn btn-primary btn-block"
+  onclick={{action @onClick}} ...attributes>
```
NullVoxPopuli: I don't think this `action` is needed. could speed up some things if we aren't doing an extra bind

frameworks/keyed/ember/app/templates/components/my-table.hbs
```diff
@@ -1,73 +1,72 @@
...
-        <div class="col-md-6">
-            <h1>Ember v3.5.1</h1>
+  <div class="row">
+    <div class="col-md-6">
+      <h1>Ember v3.9.1</h1>
+    </div>
+    <div class="col-md-6">
+      <div class="row">
+        <div class="col-sm-6 smallpad">
+          <BsButton id="run" @onClick={{action this.create}}>
```
NullVoxPopuli: I think convention now is to use `@action` in the js, rather than the in the template

frameworks/keyed/ember/app/templates/components/my-table.hbs
```diff
@@ -1,73 +1,72 @@
...
 </div>
 
 
 <table class="table table-hover table-striped test-data">
   <tbody>
-  {{#each this.data key='id' as |item|}}
-
-    <tr class={{if (eq item.id this.selected) 'danger' 'x'}}>
-      <td class="col-md-1">{{item.id}}</td>
+    {{~#each this.data key="id" as |item|~}}
```
NullVoxPopuli: oh what's this squiggly?

frameworks/keyed/ember/app/utils/benchmark-helpers.js
```diff
@@ -31,25 +28,25 @@ export const buildData = (id, count = 1000) => {
       "table", "chair", "house", "bbq", "desk", "car", "pony", "cookie",
       "sandwich", "burger", "pizza", "mouse", "keyboard"];
 
-    var data = [];
+    var data = A();
```
NullVoxPopuli: should we be using tracked properties and native arrays and objects?

lifeart: @NullVoxPopuli  can't get it working in stable release

frameworks/keyed/ember/app/templates/components/my-table.hbs
lifeart: it will be replaced (`~`) - remove space

frameworks/keyed/ember/app/templates/components/my-table.hbs
lifeart: yeah, but `fn` non landed (

frameworks/keyed/ember/app/templates/components/bs-button.hbs
lifeart: it's static rendering, don't think it has an impact

lifeart:
image
updated keying logic

lifeart:
image

lifeart:
@NullVoxPopuli 
image

frameworks/keyed/ember/app/templates/components/my-table.hbs
```diff
@@ -1,73 +1,58 @@
...
-        <a onclick={{action this.remove item.id}}>
-          <span class="glyphicon glyphicon-remove" aria-hidden="true"></span>
-        </a>
-      </td>
-
-      <td class="col-md-6"></td>
-    </tr>
-
-  {{/each}}
+    {{~#each this.data key="@index" as |item|~}}
```
simonihmig: What is `@index` here, as it's not a prop passed to `{{my-table}}`! 🤔

frameworks/keyed/ember/app/templates/components/my-table.hbs
simonihmig: Fwiw, the change I did in #506 should have removed *all* unnecessary whitespace, so not sure if that is needed/makes any difference?

frameworks/keyed/ember/app/templates/components/my-table.hbs
NullVoxPopuli: Found the docs on this! https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/if?anchor=each

NullVoxPopuli: though, more detail would be good there. I still don't know what all the options are for `key`, and why `@index` means the actual index. convention? hard coded?

NullVoxPopuli: https://discuss.emberjs.com/t/new-glimmer-consideration-each-key/8128/9

lifeart: https://github.com/emberjs/ember.js/issues/16975

NullVoxPopuli: I would have thought using the index as the key would result in stale / wrongly updated rows when things are removed / inserted.  🤷‍♂

lifeart: As I understand my goal is - get isolated component-based references https://github.com/glimmerjs/glimmer-vm/blob/master/guides/04-references.md

We have amazing kvo in ember/glimmer, and we need to use all benifits.
Previously ember project in this bench don't use `kvo` only mutable data, and glimmer-vm has only one valid reference (`id`), and it need to recalculate and revalidate all the stuff.
 
Now glimmer has index-like references to KVO objects, and it wins for cases where test suit change data inside rows.

I think @chancancode can clarify this from glimmer engineering side.

simonihmig: > emberjs/ember.js#16975

That's interesting! So I have an idea now what `@index` means. But I still don't get why it made any difference in this case. 

The example in the linked issue was an array of primitive values, but in our case we have objects (EmberObjects now). So I would expect to having no `key` (same as `@identity) to just work the same (now that we mutate the items the identity stays the same). Or - as it was before - to have `key='id'`, which would also correctly match the items with their existing DOM elements.

So why is `@index` any better here?

ryansolid:
While I have no idea how the internals work whenever you see disproportionately low replace rows and swap row scores comparatively it's worth checking your solution is still keyed. While sometimes it is just the inconsistency of benchmark runs, those 2 and remove row are the ones that dramatically improve in non-keyed. It's possible @index is forcing non-keyed (ie.. using the array index as the key).

frameworks/keyed/ember/app/templates/components/my-table.hbs
lifeart: Have no idea( I have screenshots with `@index`/`@identity` and looks like somehow for this case `@index` more performant, glimmer https://github.com/krausest/js-framework-benchmark/pull/555/files works well with default `@identity` (can't get perf impact with `@index`)

We can install DOM mutation observers and measure mutation count for both cases

lifeart: image

lifeart: on 30 runs perf close to preact
image

ryansolid:
Actually seeing the rest of your discussion now everything you are seeing makes complete sense. Basically there are 3 ways to attack array reconciliation. 2 of them are keyed by the definition of this benchmark and 1 is not.

Identity(Keyed)  - uses referential check. This is what KVO libraries use by default generally since their data is mutable and references are fixed. This is usually the most performant keyed approach but it is incompatible with immutable data. VDOM libraries cannot use this approach generally for that reason.

Explicit Key(Keyed) - Uses key check. Necessary for Immutable renderers to maintain keyed updates. KVO libraries could use this approach but usually it's not necessary. This is part of the recommended API for most VDOM/Top Down libraries. 

Index(Non-Keyed) - Uses the array index. Essentially reusing the existing nodes and just updating the values. There are performance benefits in Replace Rows, Swap Rows, and Remove Row, but it comes at the cost of losing DOM state. This is the default behavior for VDOM libraries if a key is not set. For KVO you have to go out of your way to set this mode since the default is identity.

If Ember uses that naming convention for these 3 modes that suggests that while the results are good, you are perhaps looking at the wrong comparison.

simonihmig:
Thanks @ryansolid for chiming in!

@lifeart indeed it seems the current implementation would qualify as "ember-v3.9.1-non-keyed", but not as the keyed version. See also https://www.stefankrause.net/wp/?p=504

Still not sure why in the case of GlimmerVM the keyed version would be so much slower. When I last debugged into it, I remember that for the swap test, basically all rows were moved around, although in the most optimized case only touching those two rows that are swapped would be sufficient.

Here for the record my remarks in Ember's Discord after those findings (end of last year):

image

@chancancode any idea why GlimmerVM is so far behind other frameworks, especially for the swap-benchmark and the keyed version?

ryansolid:
I don't think Ember's performance particularly worse than other libraries in certain areas but more just across the board.  I mean the Swap Rows is likely a very naive approach but half libraries do that. You go up the list one by one until you find an item out of place then insert the right item at that location pushing the rest back. But you will always be off by one index so you basically bubble up the first row all the way to end. You end up moving every element between the 2 indexes. Other libraries use a different set of heuristics to pre-emptively decide to do minimal set of DOM operations since those are the expensive operations. 

It is very obvious Ember takes a hit on initial row spamming since it has to set up all it's reactive data where VDOM libraries get a free pass. But the partial updates is where I'd be concerned. For KVO library you should be owning that. It's the most pure look at the overhead of the change management system. So that is basically your best case. If you are slow there many other benchmarks may be proportionally slow since it's the same change mechanism behind it and you aren't likely winning the row spamming race anyway. Sure using the right DOM API's are part of the puzzle here, but I imagine in many cases Ember is using similar things to other popular libraries.

lifeart:
@simonihmig @ryansolid PR updated to use `@identity`

image

But I'm wondering, why glimmer-vm beat this gap using identity? (see https://github.com/krausest/js-framework-benchmark/pull/555)

NullVoxPopuli:
Shouldn't the tests check for incorrect inserts?

lifeart:
looks like it's an bug in `glimmer-vm - ember` integration

ryansolid:
Look at where Glimmer VM does well. I wouldn't be surprised it's non-keyed.

lifeart:
@ryansolid it's keyed (at least in template syntaxis level - same as ember version)

NullVoxPopuli:
@lifeart, is this something we can make a repro repo for the glimmer folks?

ryansolid:
Maybe it's a bug.? While preact isn't necessarily the fastest benchmark to have on the left side. If it isn't just variation in runs Replace Row is absurdly low. Most libraries are within 20% on replace rows. So when you see 57 there to React and Preacts 130s it may mean something is up. Again it's Benchmark 2, 5, and 6 that are driving the performance numbers up.

krausest:
BTW - The test whether an implementation is keyed can be run with following command in webdriver-ts:
```
npm run isKeyed -- --framework ember-v3.5.1-keyed
```
(The test tries its best to distinguish between keyed and non-keyed but was outsmarted a few times in the past)

@lifeart Do you have an idea why there's no version number for react and preact in your table? As long as there's a package-lock.json in the directory for the react and preact implementation the table should show that version.

simonihmig:
> But you will always be off by one index so you basically bubble up the first row all the way to end.

Yes, this was exactly what I was observing.

> Other libraries use a different set of heuristics to pre-emptively decide to do minimal set of DOM operations since those are the expensive operations.

Do you have any references at hand?

> But the partial updates is where I'd be concerned. 

Yes, indeed. 

Also worth noting here: in this blog post by @wycats there are shown benchmark results of Glimmer vs. Preact/React/Inferno based on this very benchmark, which suggest that Glimmer is substantially faster for "partial update": https://yehudakatz.com/2017/04/05/the-glimmer-vm-boots-fast-and-stays-fast/

But the published results as well as all the screenshots here do not confirm this!? 🤔

> But I'm wondering, why glimmer-vm beat this gap using identity?

Indeed. Especially since in the previous versions Ember and Glimmer were pretty close together (though relatively slow), see my screenshot.

lifeart:
@krausest 555 pr says - it's keyed (for glimmer), for current version of this pr is keyed too.
Regarding versions - I'm trying to follow this https://github.com/krausest/js-framework-benchmark/pull/482 readme, I suggest it can be old results? Don't think `preact`, `react` results tested so mutch (only once - `yarn && yarn build-prod` + `yarn selenium --count 3 --framework preact / react`.

lifeart:
@NullVoxPopuli yeah, it make sence, also I found rendering bug in glimmer implementation with `@identity` key -> after first removal view won't update for other removals. Replacing key from `@identity` to `id`  helped to solve this problem.

frameworks/keyed/ember/app/templates/components/my-table.hbs
lifeart: `~` - removed, thanks!

lifeart:
Using mutation observer:

* Create 1.000 rows -> 13001 mutations
* Create 10.000 rows -> 131002 mutations
* Append 1.000 rows -> 13001 mutations
* Update every 10th row -> 100 mutations
* Swap Rows -> 1996 mutations
* Clear -> 1001 mutations
* Remove Row -> 3 mutations

krausest:
Thanks. Results follow the picture from the glimmer PR and don't look as good as in the first screenshots in that PR.
Bildschirmfoto 2019-04-30 um 20 06 34

lifeart:
@krausest thank you! Yeah, first screenshot taken using `key="@index"` wich is not relevant for keyed case (as we found out)ю

PR #556, opened 2019-04-29
Updated literaljs and added shouldUpdate lifecycle method

Mikeysax:


krausest:
I think I found a bug: Click create 1,000 rows. Then remove the first row and try to remove the row with id 2 (it won't work).

Issue #553, opened 2019-04-22
Update Svelte to latest 3 version


artemirq:

ryansolid:
If there are no takers I will do it. I'm very interested to see Svelte 3 perform in a credible benchmark.

frederikhors:
Thanks @ryansolid!

krausest:
Thanks!
Results have been updated (screenshots are in #557).

PR #557, opened 2019-05-05
Update Svelte to v3.2.0

ryansolid:
As requested on https://github.com/krausest/js-framework-benchmark/issues/553.

I'm not sure performance is any better. I didn't witness that. But on positive the new syntax really was delightful. Idiomatic array updates using spreads like immutable libraries and the bindings more inline with JSX. Really I can only think of a library or 2 that achieve the same with less written code.


krausest:
Thanks a lot!
Overall performance seems about the same as for 2.9.7 (but select row has improved)
Bildschirmfoto 2019-05-05 um 19 38 36
Since I still feel guilty for treating svelte so badly (#549) I'm also posting a screenshot of startup time and memory usage (where svelte looks pretty strong)
Bildschirmfoto 2019-05-05 um 19 49 50
Bildschirmfoto 2019-05-05 um 19 50 01

frederikhors:
I think we're missing something here.

ryansolid:
@frederikhors How so? The performance gain not as much as you expected? The array reconciliation API hadn't changed as far as I could tell so that part stayed the same from implementation standpoint. Admittedly I've never touched Svelte before doing this implementation so I may have missed something.

My understanding was most of the improvements were in syntax (way cleaner I think there is only 1 or 2 libraries that do the benchmark with less implementation code) and bundle size (shaved a whole kilobyte on the implementation). Although I'd obviously love to hear Rich's thoughts.

ryansolid:
@frederikhors @Rich-Harris 
I think I may have one idea. Does Svelte do implicit event delegation? In large row benchmark like this adding 2 event listeners per element has a real cost. Most libraries that don't implement implicit event delegation add some lookup code. On one side it would be a real travesty of muddying up how clean the code is, but it has real benefit on half the performance benchmarks.

deklanw:
> 
> 
> @frederikhors @Rich-Harris
> I think I may have one idea. Does Svelte do implicit event delegation? In large row benchmark like this adding 2 event listeners per element has a real cost. Most libraries that don't implement implicit event delegation add some lookup code. On one side it would be a real travesty of muddying up how clean the code is, but it has real benefit on half the performance benchmarks.

https://github.com/sveltejs/svelte/issues/1098

ryansolid:
@deklanw @Rich-Harris @frederikhors 
I can look at adding it manually. I've done it for a number of implementations in the past. I'm just unclear what is more important to Svelte, "Less code", or "Performance". This part comes down to which best tells the story.

I might lean towards "Less Code" simply because I feel explicit Event Delegation might butcher the aesthetics of the implementation, and will only improve performance a certain amount. There are other optimizations of list reconcilliation and element creation employed at the library level that make more of a difference. If we don't hear from Rich, I will take consensus here. But I have no problem doing a few tweaks to get the performance up.

frederikhors:
@ryansolid For what it's worth I'm on the performance side. I think @Rich-Harris will come soon here.

ryansolid:
Yeah if nothing by the weekend I will add event delegation. I wrote a medium article talking about writing this implementation and experience with Svelte. https://medium.com/@ryansolid/javascript-ui-compilers-comparing-svelte-and-solid-cbcba2120cea. Maybe this will draw someone more involved with the project out.

PR #558, opened 2019-05-08
Update san to 3.7.5

errorrik:

- update san version to 3.7.5
- add `autoFillStyleAndId` property
- use for instanceof forEach


krausest:
Thanks. And I just noticed that I let a caret version slip into the benchmark so it's 3.7.6 now...
Here are the keyed results - nice improvement for swap rows!
Bildschirmfoto 2019-05-16 um 21 58 21

errorrik:
> I let a caret version slip into the benchmark so it's 3.7.6 now...

It's ok. Thanks ^_^

PR #559, opened 2019-05-09
Update Solid, Knockout JSX, MobX JSX, and KnockoutJS

ryansolid:
Update to latest versions of these libraries. This release mostly focused on optimizing code size to combat the new features like:

* Control Flow Fallbacks
* Portals
* Suspense
* forwardRef
* Tagged Template Literals and HyperScript alternate Renderers
* Adding bubbling to Synthetic Events

Taking a page from Svelte the JSX compiler now handles import of methods in so auto detecting only the code that is required to include. However I think my array reconciler alone is the size of Svelte's whole runtime so nothing too amazing here. The price of performance.

Also Knockout 3.5.0 was released end of February so even if it loses a bit on performance probably deserves the update.


krausest:
Thanks. Clear rows got worse for all of them (I ran a few times and am pretty sure it's not an outlier). Please note that there was a chrome security update between both runs (but vanillajs actually improved between both) so comparison with older results maybe a bit shaky - any idea about what might be the cause?  
Bildschirmfoto 2019-05-16 um 21 58 21

ryansolid:
Ok, I was afraid of that. I will have to do some more testing. My windows setup was pretty shaky as of late so I haven't been nearly as diligent on specific tests. Localvoid posted similar results with Solid recently so I must have introduced something unwittingly in the past couple months. 

I will dig in and probably have another pull request in the next couple days.

ryansolid:
I figured it out. Nothing important. Incorrect strict equality check leading to unoptimized code path. I apologize for the unnecessary pull request. (I ended up going back to WSL with chromium 73, chromedriver 73, using x411 as I can atleast get consistent-ish results over several runs)

PR #560, opened 2019-05-16
hullo

arturkulig:
This adds a test for a new renderer in `hullo` libraries family.

### DIFF
--- a/frameworks/keyed/hullo/.babelrc
+++ b/frameworks/keyed/hullo/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["babel-preset-es2015"]
+}
--- a/frameworks/keyed/hullo/.yo-rc.json
+++ b/frameworks/keyed/hullo/.yo-rc.json
@@ -1,0 +1,45 @@
+{
+  "js-framework-benchmark-hullo": {
+    "configuration": {
+      "config": {
+        "topScope": [
+          "const webpack = require('webpack')",
+          "const path = require('path')",
+          "\n",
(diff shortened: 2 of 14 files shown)


krausest:
Thanks. I'm afraid the keyed version isn't keyed for updating the whole table. Click "Create 1,000 rows", set the background of a row to some color in the dev tools and click "Create 1,000 rows" again. A keyed implementation must not keep the color. 

Here's the result from the isKeyed test (run in webdriver-ts)
```
npm run isKeyed -- --framework 'hullo-v1.0.0-keyed' 'hullo-v1.0.0-non-keyed'

> webdriver-ts@1.0.0 isKeyed /Users/stefan/Source/Javascript/js-framework-benchmark/webdriver-ts
> cross-env LANG="en_US.UTF-8" node dist/isKeyed.js "--framework" "hullo-v1.0.0-keyed" "hullo-v1.0.0-non-keyed"

WARNING: Ignoring package keyed/.DS_Store: No package.json found
Frameworks that will be checked hullo-v1.0.0-keyed hullo-v1.0.0-non-keyed
hullo-v1.0.0-keyed is non-keyed for 'run benchmark' and keyed for 'remove row benchmark' keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
ERROR: Framework hullo-v1.0.0-keyed is not correctly categorized
hullo-v1.0.0-non-keyed is non-keyed for 'run benchmark' and non-keyed for 'remove row benchmark' non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
```
Is it possible to create a completely keyed version?

arturkulig:
Thank you for telling this and so quickly :)

This is a very particular case for this renderer.
I think it fulfils other expectations of keyed (vs non-keyed) approach - it will reorder, not update elements in "swap" scenario.
With this renderer should a new node (due to key tracking, but might for another reason) replace an old one and both are of the same tag, the replacement does not occur - old element is cleaned up and passed over as new without even removing it from the DOM. This is quite like simple and momentary object pooling.

Is this a deal breaker for keyed version?

krausest:
Sorry, I'd say yes it is. Should I merge only the non-keyed version?

arturkulig:
This, I changed. Please reconsider the request.

krausest:
Thanks - looks good now. Pretty solid results:
Bildschirmfoto 2019-06-10 um 11 03 55

arturkulig:
SO 🎉 🎉 🎊 🎉 🎉  HAPPY

PR #561, opened 2019-05-16
Add Event Delegation to Svelte Implementation

ryansolid:
This commit adds event delegation to the Svelte 3 implementations as suggested by #557.

I kept this on Svelte 3.2.0 since there are some performance issues with the current version as noted in:  https://github.com/sveltejs/svelte/issues/2768 and I don't want to take any risks in compromising performance.

### DIFF
--- a/frameworks/keyed/svelte/src/Main.html
+++ b/frameworks/keyed/svelte/src/Main.html
@@ -30,14 +30,12 @@
   </div>
 </div>
 <table class="table table-hover table-striped test-data">
-  <tbody>
+  <tbody on:click={click}>
     {#each data as row, num (row.id)}
       <tr class={selected === row.id ? 'danger' : ''}>
         <td class="col-md-1">{row.id}</td>
-        <td class="col-md-4">
-          <a on:click={() => select(row.id)}>{row.label}</a>
-        </td>
-        <td class="col-md-1"><a on:click={() => remove(row.id)}><span class="glyphicon glyphicon-remove" aria-hidden="true"></span></a></td>
+        <td class="col-md-4"><a>{row.label}</a></td>
+        <td class="col-md-1"><a><span class="glyphicon glyphicon-remove" aria-hidden="true"></span></a></td>
         <td class="col-md-6"></td>
       </tr>
     {/each}
@@ -60,8 +58,8 @@
         data[i].label += ' !!!';
       }
     },
-    remove = (num) => {
-      const idx = data.findIndex(d => d.id === num);
+    remove = id => {
+      const idx = data.findIndex(d => d.id === id);
       data = [...data.slice(0, idx), ...data.slice(idx + 1)];
     },
     run = () => {
@@ -72,10 +70,18 @@
       data = buildData(10000);
       selected = undefined;
     },
-    select = (id) => selected = id,
+    select = id => selected = id,
     swapRows = () => {
       if (data.length > 998) {
         data = [data[0], data[998], ...data.slice(2, 998), data[1], data[999]];
+      }
+    },
+    click = ({ target }) => {
+      const id = Number.parseInt(target.closest("tr").firstChild.firstChild.nodeValue);
+      if (target.matches(".glyphicon-remove")) {
+        remove(id);
+      } else {
+        select(id);
       }
     };
 
--- a/frameworks/keyed/svelte/src/main.js
+++ b/frameworks/keyed/svelte/src/main.js
@@ -1,4 +1,4 @@
-import Main from './Main.svelte';
+import Main from './Main.html';
 
 export default new Main({
   target: document.querySelector( '#main' )
--- a/frameworks/non-keyed/svelte/src/Main.html
+++ b/frameworks/non-keyed/svelte/src/Main.html
@@ -30,14 +30,12 @@
 	</div>
 </div>
 <table class="table table-hover table-striped test-data">
-	<tbody>
+	<tbody on:click={click}>
 		{#each data as row, num}
       <tr class={selected === row.id ? 'danger' : ''}>
         <td class="col-md-1">{row.id}</td>
-        <td class="col-md-4">
-          <a on:click={() => select(row.id)}>{row.label}</a>
-        </td>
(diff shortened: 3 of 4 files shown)


frameworks/keyed/svelte/src/Main.html
```diff
@@ -72,11 +70,19 @@ <h1>Svelte (keyed)</h1>
...
     },
-    select = (id) => selected = id,
+    select = id => selected = id,
     swapRows = () => {
       if (data.length > 998) {
         data = [data[0], data[998], ...data.slice(2, 998), data[1], data[999]];
       }
+    },
+    click = ({ target }) => {
+      const id = Number.parseInt(target.closest("tr").firstChild.firstChild.nodeValue);
```
leeoniya: @ryansolid

so, we're still doing this, huh?

ryansolid: Fair is fair? I added implicit Event Delegation to my libraries to align with what the VDOM implementations. And I like the story it tells for code aesthetics ("Write less code"), but it is hardly a fair comparison when implementations aren't allowed to use the exact same mechanisms.

I realize the same could be argued about any trick VanillaJS does, but it just means the floor is pretty low. It's the same old discussion again. Which for me comes down to how does the implementation want to present itself. As long as the minimum requirements are met (Ie.. completely data driven, ie renderable from non-DOM state at any given point) it is up to them how nasty and complicated they want to make their implementation.

So yes I was hesitant to do this initially. Especially without @Rich-Harris approval since I get the impression Svelte is a library that cares more about aesthetics than performance. But absent of that I'd say it's probably fair to align this with other high performing non-VDOM implementations like Surplus, LitHTML, LighterHTML, and so on.

As always ready to open this debate again. But I think it is very difficult for any implementation/implementor that cares about performance and doesn't include implicit event delegation in their libraries for various reasons, to even understand why this is an issue in itself. I implemented implicit event delegation and I still don't.

krausest:
Here's the difference we're be talking about:
Bildschirmfoto 2019-05-16 um 22 51 06

ryansolid:
Yeah that's significant. More than I thought (I'm sort of writing implementations blind right now). It puts it right in the mix with the others. Given that gain I'm gathering for the cost in aesthetics, all being fair, this represents Svelte the best. I'm sold. Does anyone want to veto this?

@leeoniya @localvoid @Rich-Harris

leeoniya:
i'm not brave enough to cast a veto vote. i think the vdom vs non-vdom distinction is irrelevant to whether this strategy should be acceptable, since most vdom impls can do the same thing. we just have different places where we draw the line for what constitutes a fair representation of lib perf comparison. i think by now everyone's aware that i'm in the camp of implicit delegation (that's baked into the lib) rather than manual dom crawling and readbacks in the impls. i had this in domvm's impl long ago, and removed it by adding implicit delegation and taking the small perf hit in favor of ergonomics.

paulocoghi:
I'm impressed with the difference!

krausest:
To be honest: I'm not sure how I should handle this request. If it was just for svelte and would serve as a reference on how much this benchmarks benefits from implicit delegation I could certainly live with it, but I'd like to keep both versions: The old version and an "-optimized" version with explicit event delegation.

I thought a while about allowing a "-optimized" version for all frameworks. Those versions would be allowed to use explicit event delegation (or any other gradual transistion to becoming vanillajs - are there any rules left for them?). 

The next logical step would be to add a filter in the UI to remove -optimized results...

If I end up with most frameworks having two versions, maintenance would become much harder, so maybe a rule will be needed that only some (important) frameworks should be allowed to have an optimised version (angular has that privilege already btw.)

I think I'll come back to it when I've decided how to run the benchmarks on my new (well, not really new anymore) notebook.

ryansolid:
@krausest I understand the difficulty here. I think you are correct not to merge this. Unfortunately I doubt there will ever be agreement on event delegation. I believe that there is a clear line that can be drawn for this benchmark as highlighted in the first post of #430. We had very little difficulty convincing offenders to reach that line, but event delegation could not be agreed upon. But it also became clear from the last round there is a sizeable group that feels that line is way low to be considered acceptable.  I also don't think this is an optimized/non-optimized scenario as it affects every Framework in this benchmark. It isn't some specific tweak for Svelte. Having both isn't worth the complication brings.

The problem here is anecdotally you are looking at a spread something like:

| Explicit  | Parameterized | Implicit | No Event Delegation |
|---------|----------------|--------|----------------------|
|      1.00  |                   1.02 |      1.05 |                             1.15 |

You are penalized way more for not using Event Delegation than you are for small differences on how you do it. That's why when one talks about the cost of their implicit event delegation (and I've done this myself, guilty as charged) it really is of no consequence compared to being blocked from even joining the party.

I created this version since I knew some people were discouraged by the Svelte 3 initial results and didn't understand why Svelte was so much slower. The answer is, this wasn't apples to apples. As to whether it should be is much more complicated. I've done what I can to draw people's attention to it. I'm fine closing this and letting @Rich-Harris or the like submit something if it's important to them. If this goes another week or so I think will just do that.

localvoid:
@ryansolid for libraries like `lit-html` it is not about event delegation at all, they are trying to hide the cost of data bindings.

ryansolid:
@localvoid 
Can you elaborate? Stuff around toggling selected row I can see, but my understanding Event listeners are mostly static so the cost on creation should be basically the same as calling `el.addEventListener`. Unless lit-html has implicit event delegation (perhaps an inefficient version) and is bypassing it for the sake of the benchmark. That would be much more underhanded.

localvoid:
@ryansolid like any other data binding, it won't be able to just clone a node with its subtree, it will need to find a node and assign an attribute, property or add an event listener and the overall cost of data bindings is way much higher than the difference between implicit event delegation and `el.addEventListener`.

ryansolid:
@localvoid Isn't that like anything? I mean I'm looking at domc with it's parameterized event delegation. It still needs to walk the tree to assign the property it will be using "__click" in this case. Solid is essentially the same. I walk then statically assign the property once on creation.

I'm gathering the argument is that lit-html does a lot more. So I shouldn't assume the event handling is as simple as imagined. And that is why event delegation makes a larger difference on some libraries compared to others.

ryansolid:
Closing due to lack of interest from Svelte Community and recognizing short of that this won't get pushed through.

PR #562, opened 2019-05-17
Fix performance regression in Solid, MobX JSX, and Knockout JSX

ryansolid:
I believe I've found the culprit to the issue found in #559 . A too strict null check lead to unoptimized clear all path for all 3 libraries. This pull requests updates the libraries to version with the fix.

### DIFF
--- a/frameworks/keyed/ko-jsx/package.json
+++ b/frameworks/keyed/ko-jsx/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-ko-jsx",
-  "version": "0.6.2",
+  "version": "0.6.3",
   "main": "index.js",
   "js-framework-benchmark": {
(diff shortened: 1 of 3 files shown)


krausest:
Thanks! Seems a bit faster. Here's a screenshot:

ryansolid:
Ok, yeah. In proportion to Vanilla JS seems about right. Although compared to the whole field it's about 10ms slower than I expected.

Is it possible newer versions of chrome have slowed that benchmark down a tiny bit. Vanilla JS at 143, compared to low 130s like Vanilla JS 1 or DOMC or Stage0. I wonder if other libraries would be coming in 10ms or so slower than before.

krausest:
There was a chrome security update and an OSX update, so yes, I really should re-run the whole benchmark.

PR #563, opened 2019-05-18
fidan-v0.0.23-keyed

ismail-codar:
Compile time code generation with jsx version. Thanks @ryansolid for babel-plugin-jsx-dom-expressions. Here is the my results:
image


krausest:
What an improvement! Results have been updated. Here's a screenshot:

PR #566, opened 2019-05-21
Add sinuous-v0.6.0-keyed

luwes:
Adds [Sinuous](https://github.com/luwes/sinuous/) v0.6.0.

Thanks to @ryansolid, @adamhaile and @Freak613 for crucial code in this lib,  
also @developit and @jviide for [htm](https://github.com/developit/htm)!

The idea was to 

- keep the original [hyperscript](https://github.com/hyperhype/hyperscript) idea alive 
- with a better [observable](https://github.com/luwes/sinuous/tree/master/packages/sinuous/observable) API 
- favor template literal over JSX
- no "precompilation", `h` tag is the lowest it can get
- keep bundle size small for lightweight components
A minimal Sinuous app without a list is `< 2kB` gzip
- developer experience++, less library specialized concepts, plain JS feel

Results:

Screen Shot 2019-05-21 at 10 46 30 AM

### DIFF
--- a/frameworks/keyed/sinuous/.eslintrc.yml
+++ b/frameworks/keyed/sinuous/.eslintrc.yml
@@ -1,0 +1,21 @@
+env:
+  browser: true
+  es6: true
+
+extends:
+  - eslint:recommended
+
+parser: babel-eslint
+
+parserOptions:
+  ecmaVersion: 9
+  sourceType: module
+
+globals:
+  html: readonly
+
+rules:
+  semi: error
+  no-unused-vars:
+    - error
+    - varsIgnorePattern: ^h$
--- a/frameworks/keyed/sinuous/index.html
+++ b/frameworks/keyed/sinuous/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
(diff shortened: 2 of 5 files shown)


krausest:
Thanks! Results have been updated. Here's a screenshot:

ryansolid:
@luwes I'd check your disposal logic in your list reconciler. It looks like there might be a memory leak. Usually high memory usage in both of the last 2 memory benchmarks can suggest that computations on each row are not being released explicitly. It isn't necessarily true but those are the benchmarks that would catch memory leaks of that nature.

luwes:
Oh yeah that looks pretty bad, thanks for letting me know @ryansolid

Have you seen Sinuous Observable?
https://github.com/luwes/sinuous/tree/master/packages/sinuous/observable

It's only missing the freeze API and error handling but there might be some useful code in it if you decide to make Solid smaller or dependency free.

ryansolid:
Thanks I will take a look. I have actually been experimenting in that area again recently.

Solid pre v 0.1.0 had its own observable lib but I went to Surplus after complexity got out of hand with my own batching logic(freeze). Most observable libraries don't have this freeze/transaction concept but it is pretty important once you move beyond benchmarks.

PR #569, opened 2019-05-23
Update lit-html to v1.1.0

CaptainCodeman:


krausest:
Here's a screenshot for keyed results (left chrome 75 + updates, right chrome 74 + old version):
Bildschirmfoto 2019-06-10 um 10 30 50

PR #570, opened 2019-05-24
Update sinuous to v0.7.1

luwes:
This should improve memory use for reactive libraries w/ semi-automatic cleanups.
The parent computations take care of that.

Thanks to @ryansolid for detecting this issue!


krausest:
Here's a screenshot for keyed results (left chrome 75 + updates, right chrome 74 + old version):
Bildschirmfoto 2019-06-10 um 10 30 50

Memory results looks much better now!
Bildschirmfoto 2019-06-10 um 10 45 32

luwes:
Thanks @krausest! Real happy with the results.

PR #572, opened 2019-06-05
Updated literaljs to 6.0.5

Mikeysax:


krausest:
Bildschirmfoto 2019-06-10 um 10 48 41

PR #573, opened 2019-06-11
chromedriver 75 + cleanup

krausest:
Chromedriver 75 needs different configuration to allow access to the performance log.


ryansolid:
Awesome it looks like whatever the issue that was introduced in Chrome 73 seems to have phased out again. Not seeing the overhead on my implementations anymore. I wonder if the memory issue is better now too (the one that forced you to do full restarts between tests). I haven't checked it myself as of yet, and it might been coincidental that both performance hits happened at the same time so one resolving doesn't necessarily mean the other has, but it's a thought.

krausest:
Yes, it seems like running the whole benchmark on windows is pretty close. 
On my new laptop it‘s absolutely required to run on AC. Performance and variance is pretty bad on battery (haven‘t found a power setting that avoids that). But it seems stable again.

The one big issue left is that there a pause of a few seconds after the chrome browser launched and before the framework is loaded. This is on windows only, OSX and Linux don‘t have that.  Do you see that pause on your machine too?

ryansolid:
Yeah battery variance is big. Windows has that simplistic power mode setting when you click on the battery symbol but I'm not convinced Best Performance behaves the same on battery.

I did notice the delay when I ran the benchmark the way you described in the other issue building and running the web server in WSL and then running the benchmark in windows. It was the most stable build I had but the results especially with slowdown were sporadic. I went back to running everything in WSL though to get more consistent numbers (and I don't run all the frameworks) but it crashes a reasonable amount. I was hoping you'd report back having figured out some great solution for Windows. It sounds like at least your numbers are more stable now.

PR #575, opened 2019-06-17
Update Sinuous to v0.8.0 and add data selection

luwes:
Under the hood this update does a few things:

- Adds a `template` feature to get the perf benefits of `element.cloneNode()` https://github.com/luwes/sinuous/pull/4
- Optimizes the clear method with `element.textContent = ''` if there are no siblings.

This change also makes the "selection" fully data driven.
https://github.com/krausest/js-framework-benchmark/compare/master...luwes:sinuous?expand=1#diff-625243bf69109917f89eb7c1adf2cef2R58

### DIFF
--- a/frameworks/keyed/sinuous/package.json
+++ b/frameworks/keyed/sinuous/package.json
@@ -17,7 +17,7 @@
     "url": "https://github.com/krausest/js-framework-benchmark.git"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


krausest:
Thanks - faster once again. Here are the CPU results:
Bildschirmfoto 2019-06-18 um 08 12 22

PR #576, opened 2019-06-17
Switch to hooks API in react-redux

localvoid:
1. Reimplemented with a [react-redux hooks API](https://react-redux.js.org/next/api/hooks)
2. Modified state mutations to make sure that it is implemented in a similar way as `lit-html` (they've changed it in one of the latest pull requests and I am just trying to reduce differences)

### DIFF
--- a/frameworks/keyed/react-redux/src/main.jsx
+++ b/frameworks/keyed/react-redux/src/main.jsx
@@ -1,18 +1,18 @@
-import React from 'react';
-import ReactDOM from 'react-dom';
-import { createStore } from 'redux';
-import { Provider, connect } from 'react-redux';
+import React, { useCallback } from "react";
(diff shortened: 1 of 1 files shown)


krausest:
Thanks. Results look pretty bad. I decided to keep two redux implementations until hooks (in redux) are reasonable fast. Did you expect that?
Bildschirmfoto 2019-06-18 um 08 12 22

localvoid:
> Did you expect that?

No, I've submitted implementation that I've used in this run: https://localvoid.github.io/js-framework-benchmark/webdriver-ts-results/table.html . Maybe something changed since `7.1.0-alpha.4`

localvoid:
I've figured out the issue, they've changed `useSelector()` API. I'll submit a fix in a couple of minutes.

PR #577, opened 2019-06-17
Update ivi

localvoid:
1. Updated ivi to the latest version
2. Removed some differences between `react-redux` and `ivi` implementations (state mutations, useless selector)


krausest:
Thanks. Results look like that:
Bildschirmfoto 2019-06-18 um 08 10 46

Issue #578, opened 2019-06-17
Solid implementation doesn't have aria-hidden attributes


localvoid:
@ryansolid glyph icons should have `aria-hidden` attribute:

1. [vanilla](https://github.com/krausest/js-framework-benchmark/blob/a43bb85d29f55240fb6fb2e146146e64d12eb0f3/frameworks/keyed/vanillajs/src/Main.js#L26)
2. [react](https://github.com/krausest/js-framework-benchmark/blob/a43bb85d29f55240fb6fb2e146146e64d12eb0f3/frameworks/keyed/react/src/main.jsx#L28)
3. [ivi](https://github.com/krausest/js-framework-benchmark/blob/a43bb85d29f55240fb6fb2e146146e64d12eb0f3/frameworks/keyed/ivi/src/main.js#L73)

I don't think that it would make any perf difference since you are just cloning subtrees, but it would be better if it produced the same DOM output as other implementations.

ryansolid:
Thanks definitely. I never noticed that. All 3 of my libraries don't have it and I think a few others now may have followed suit. Thanks for bringing this to my attention.

ryansolid:
@luwes I think you ended up here with sinuous as well.  The aria-hidden needs to be on each row on the remove icon span.

leeoniya:
@krausest this is a pretty common problem. it might make sense to add an automated test (similar to isKeyed) to ensure the outerHTML of implementations is correct for initial render of a single item. i guess you can just run1k then trim the tail.

ryansolid:
It's probably a good idea, although I wonder if any libraries add extra elements just because of how they work. Some webcomponent libraries might. Also some templating engines use comment nodes as placeholders (like lit-html).

krausest:
I've added a structure check for a table row to the keyed checker nonKeyed.ts which is invoked with `npm run isKeyed` in webdriver-ts, I guess it's name needs an overhaul yet again).
A few frameworks skipped the fourth td. All of them have been fixed.

Then some frameworks don't output the aria-hidden="true" correctly.
I could fix some frameworks, but some frameworks refused to output the attribute (though some of them already include the attribute in their source code):

- [ ] miso
- [ ] rawact
- [ ] rx-domh-rxjs
- [ ] cyclejs-dom
- [ ] gruu
- [ ] stem
- [ ] tsers

@sokra @sectore @xialvjun @MarekLabuz @ericpts Can you please take a look and add the aria-hidden="true" to the span?

[...]

PR #579, opened 2019-06-17
Add aria-hidden to Solid, mobx-jsx, and ko-jsx

ryansolid:
As per #578 this adds missing attribute to these libraries.


krausest:
Thanks. Results look like that:
Bildschirmfoto 2019-06-18 um 08 10 46

PR #580, opened 2019-06-18
Fix perf regression in react-redux

localvoid:
[react-redux v7.1.0-alpha.5](https://github.com/reduxjs/react-redux/releases/tag/v7.1.0-alpha.5) removed `deps` array as a second argument.


krausest:
Thanks - looks much better:
Bildschirmfoto 2019-06-18 um 19 59 54

PR #584, opened 2019-06-21
Update Sinuous v0.9.0

luwes:
- Added small optimizations and reduced bundle size.
- Removed legacy `observable` library support
https://github.com/dominictarr/observable

### DIFF
--- a/frameworks/keyed/sinuous/package.json
+++ b/frameworks/keyed/sinuous/package.json
@@ -17,7 +17,7 @@
     "url": "https://github.com/krausest/js-framework-benchmark.git"
   },
   "dependencies": {
-    "sinuous": "0.8.1"
+    "sinuous": "0.8.2"
   },
   "devDependencies": {
     "@babel/core": "7.4.4",
--- a/frameworks/keyed/sinuous/package.json
+++ b/frameworks/keyed/sinuous/package.json
@@ -17,7 +17,7 @@
     "url": "https://github.com/krausest/js-framework-benchmark.git"
   },
   "dependencies": {
-    "sinuous": "0.8.2"
+    "sinuous": "0.9.0"
   },
   "devDependencies": {
     "@babel/core": "7.4.4",
--- a/frameworks/keyed/sinuous/src/main.js
+++ b/frameworks/keyed/sinuous/src/main.js
@@ -33,12 +33,26 @@ const App = () => {
 	const run = () => data(buildData(1000)) && selected(null);
 	const runLots = () => data(buildData(10000)) && selected(null);
 	const add = () => data(data().concat(buildData(1000)));
+	const clear = () => data([]) && selected(null);
+	const removeOrSelect = (e) => e.target.matches('.remove') ? remove(e) : select(e);
+	const select = (e) => selected(getParentId(e.target));
+
+	let rowObject;
+	const selected = (id) => {
+		if (rowObject) rowObject.selected = '';
+		const d = data();
+		if (id && (rowObject = d.find(d => d.id === id))) {
+			rowObject.selected = 'danger';
+		}
+	};
+
 	const update = () => {
 		const d = data();
 		for (let i = 0; i < d.length; i += 10) {
 			d[i].label += ' !!!';
 		}
 	};
+
 	const swapRows = () => {
 		const d = data();
 		if (d.length > 998) {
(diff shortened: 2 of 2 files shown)


ryansolid:
@luwes Hmm.. now that I'm looking at this closer I think this implementation might not be considered legitimate (including what is currently posted). I got tipped off when you removed the afterRender hook which I invented to generically solve this problem. You don't need to use the afterRender hook to solve this. Surplus doesn't. But the way it is right now breaks the rules I believe. (If there are any rules).

Basically you can't properly render from the initial state of 1000 rows and id 7 selected. You aren't storing the the id in the state data. There is no way to represent that as an initial state.  While you aren't directly storing the TR in an observable like my first attempt you are essentially just making a closure over it doing the same thing. The id gets fed in for comparison to properly handle changes but it could never be represented as initial state.

I know the @leeoniya @localvoid etc are pretty sick of these "hacks" in general so they aren't really looking at them anymore. But I believe that if there is any rule it should be that the implementation is data driven. That any given state (including selection) should be reproducable from scratch (ie complete render from data snapshot). Perhaps it's weird line to draw since it barely affects performance(and it won't take much to fix up). But it is atleast a clear line from an ideological standpoint.

luwes:
@ryansolid are you sure? if there is a `selected` property in the data it will be set here.
https://github.com/krausest/js-framework-benchmark/blob/9e20691deab04413c9cced795db5f03df3fbad77/frameworks/keyed/sinuous/src/main.js#L76-L76
https://github.com/luwes/sinuous/blob/7ad095899b4b50d74aa37ce9c2b5abc16e5975c3/packages/sinuous/template/src/template.js#L79-L79

leeoniya:
it doesnt look like you can initialize `selected` to a value prior to initial render, since it accepts a `tr` and relies on event handlers to extract the necessary id. right?

https://github.com/krausest/js-framework-benchmark/blob/9e20691deab04413c9cced795db5f03df3fbad77/frameworks/keyed/sinuous/src/main.js#L54-L60

luwes:
@leeoniya the `tr` is just to remember the last selected row, it's the row data object. Surplus did the same thing if I remember correctly.

It's possible to just add `data()[7].selected = 'danger'` and it should render correctly from scratch, fully data driven I think.

leeoniya:
that's probably fine then.

though this brings up another point of whether `selected` may be stored as part of the data or externally as in the original React impl. in my mind adding `selected` to the data is pollution with what's exclusively view state (at least for this benchmark).

for instance, if i switched to a deep-immutable data array and stored `selected` as part of each item, the diff would be significantly cheaper and domvm would do measurably better.

keeping `selected` external has a definite cost (probably in many impls). it might be worth asking @krausest to weigh in on this, too. perhaps it's too prescriptivist for his taste.

ryansolid:
@luwes Surplus has a selected observable that just holds the id. Basically on initialization there is a data array and a selected id.

I'm guessing you are setting 'danger' on the data object behind the scenes during rendering the template and that is what I'm missing. In the code I see you setting it on the tr which is why I would not expect what you wrote to work since the data has no 'selected' anywhere. I guess all observables in the context get reflected on to the underlying data.

Yeah which turns this into the other scenario where it's a selected per node instead of a global one. That ones a bit foggier. We had the same discussion over lit-html. In the end they decided to align with the other implementations but it might be worth a look.

https://github.com/krausest/js-framework-benchmark/pull/521

ryansolid:
@krausest @luwes Now that I understand what is going on I feel less strongly about this. It's arguably not equivalent to what other libraries do but this template approach used by Sinuous has to push everything to the current render context. So it's just how it works. Even if the variable was coming in globally it has to apply it per row for the system to work. I don't think we can say that this is unacceptable any more than libraries that wrap state in observables.

Thanks for the explanations. And sorry for wasting people's time.

luwes:
@ryansolid yup, the `tr` is the row data object. probably bad naming 😄 
I'll change that. 

the `template`'s clone function will only set the property if it's other than `null` or `undefined`. So in buildData that `selected` prop doesn't have to be added.

thanks for explaining and approving.

I wasn't really aware of the difference first but I see it now.

I do like the simplicity currently but I'll see if there is a better way,
might have to bring out the `tr` elements in an array like Surplus but then there is a direct DOM manipulation visible and it's less tidy/clear IMO.

ryansolid:
I think it's even better you don't expose the dom elements. I think the confusion was simply in that it looked like you were caching dom reference and that was the only way it worked. I see now you are caching a model essentially. That is arguably better than what Solid and Surplus are doing. Given the way the templates work I would say this is the idiomatic solution for Sinuous since everything has to be on the current data context. I'd be interested in what that looks like for more complicated perhaps doubly nested structures, but like domc I think it's fair to say this a new approach that is still being worked through. The inner template not being observables is really interesting way to keep things fine grained but not heavy.

krausest:
Here we go:
Bildschirmfoto 2019-06-23 um 10 08 31

luwes:
@krausest thanks! 

that partial update is killing me 😅

ryansolid:
On the otherhand you've cracked 1100 on the 10k row. Sinuous is the first fine grained library to post results in that range for 10k. Good work.

PR #585, opened 2019-06-21
Dominator

Pauan:
Dominator is a zero-cost Rust framework for creating web apps.

I have pre-compiled it (the output is in `bundled-dist`), but the README contains directions on how to compile it.


krausest:
Thanks - first results:
Bildschirmfoto 2019-06-23 um 10 20 52

I'll come back to it in a few days - I want to be able to build the rust-based frameworks on linux/docker.

Pauan:
@krausest Locally, I have updated the other Rust benchmarks so that they now compile correctly (they were broken before). If you want, I can send pull requests for them as well.

krausest:
That would be great. I've opened #586 for that.

PR #588, opened 2019-06-23
Upgrade gruu version

### LINKED ISSUE #578: Solid implementation doesn't have aria-hidden attributes

MarekLabuz:
Fixes #578 for gruu framework, meaning aria-hidden="true" is now correctly added to span

### DIFF
--- a/frameworks/non-keyed/gruu/index.html
+++ b/frameworks/non-keyed/gruu/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8"/>
-    <title>Gruu v1.7.3</title>
+    <title>Gruu v1.7.5</title>
     <link href="/css/currentStyle.css" rel="stylesheet"/>
   </head>
   <body>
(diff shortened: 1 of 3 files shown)


krausest:
Thanks. Results have been [updated](https://krausest.github.io/js-framework-benchmark/current.html).

PR #595, opened 2019-06-28
Fixing stdweb nonkeyed

Pauan:
This fixes the non-keyed stdweb demo to work.

It also updates the stdweb and yew READMEs to match the correct instructions.


krausest:
Thanks - results have been updated. Here's a comparison:
Screenshot2

PR #596, opened 2019-06-29
update domvm & results-ui

leeoniya:
been a while. v3.4.12 looks a bit faster overall on my machine.


krausest:
Thanks - results have been updated. Variance is a bit high for replace and update - I'll check them later. Here's a comparison (I really like that feature of results-ui):
Screenshot2

krausest:
Here's a comparison for non-keyed:
Screenshot2

PR #597, opened 2019-06-29
add angular-xsm, react-xsm, vue-xsm, and vue-vuex

peterluhub:
add angular-xsm, react-xsm, vue-xsm, and vue-vuex
https://github.com/peterluhub/js-framework-benchmark/tree/master/frameworks/keyed/angular-xsm
https://github.com/peterluhub/js-framework-benchmark/tree/master/frameworks/keyed/react-xsm
https://github.com/peterluhub/js-framework-benchmark/tree/master/frameworks/keyed/vue-xsm
https://github.com/peterluhub/js-framework-benchmark/tree/master/frameworks/keyed/vue-vuex


krausest:
Thanks - I've published results here:
https://krausest.github.io/js-framework-benchmark/2019/table-xsm.html

Here's a screenshot:
js-xsm

Results for XSM seem a bit mixed: It's very fast for react mostly due to select rows (and maybe that forceUpdate call that makes it faster than vanilla react and might need some discussion? ). Angular looks more expensive although comparison is limited due to differing angular versions. Vuex vs Vus XSM shows that XSM has not more overhead than Vuex.

But I'm hesitating to merge the PR, since besides some technical issues (including bootstrap, using a meta refresh, broken font in angular) it adds enough weight to the maintenance burden to reconsider. I decided to include only a few of the very popular state management frameworks to get a feeling for the associated cost. I'm afraid I won't be able to maintain this benchmark if we add all state management frameworks available. Sorry for that - I hope you can understand my decision. 
Hope this still serves as a snapshot to get a feeling for the performance cost of XSM.

PR #599, opened 2019-06-30
Update Sinuous v0.9.2

luwes:
Added an optimization where `sinuous/template` can update text nodes w/ `Text.data` instead of `parent.textContent`.

Also changes that the top element in a template can access the props via `node.props` instead of `node.data` which could collide with the native text node `.data` property.

### DIFF
--- a/frameworks/keyed/sinuous/package.json
+++ b/frameworks/keyed/sinuous/package.json
@@ -17,7 +17,7 @@
     "url": "https://github.com/krausest/js-framework-benchmark.git"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


krausest:
Thanks - results have been updated. Here's a comparison:
Screenshot2

luwes:
Thanks @krausest! 

A little surprised it performed almost the same, in this jsperf test it's really visible https://jsperf.com/update-textcontent-vs-data-vs-nodevalue

But I can imagine it's not the biggest cost in the scheme of things. worth a try :)

PR #600, opened 2019-06-30
Update CircleCI version to Node 10

### LINKED ISSUE #593: Circle CI failing?
```
class URLShim extends URL {
                      ^

ReferenceError: URL is not defined
```

`URL` globally requires Node 10
https://stackoverflow.com/questions/52566578/url-is-not-defined-in-node-js#comment92070757_52566578

or add this 

`var URL = require('url').URL;`

luwes:
This is required for Lighthouse.

> Lighthouse requires Node 10 LTS (10.13) or later.

I know Docker is used now but would still be nice to see CI passing :)

```js
/home/circleci/project/webdriver-ts/node_modules/lighthouse/lighthouse-core/lib/url-shim.js:35
class URLShim extends URL {
                      ^

ReferenceError: URL is not defined
```

Fixes #593

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
 
       - run: |
           source $NVM_DIR/nvm.sh
-          nvm install v8.9.1
+          nvm install v10.16.0
 
           npm install
           npm start &


luwes:
Should `--skipIrrelevant` be added back in?

Doing full builds might take too long, I think Circle CI would still be nice to have for checking PR's.

What are your thoughts on this @krausest?

krausest:
Thanks. I'm considering to switch to a custom docker image, but in the meantime this should help.

krausest:
I prefer full builds. I tend to merge PRs without fixed versions and they break. It's just a matter of time how soon. I really like to find those issue with the circleci build.

PR #603, opened 2019-07-01
Update Solid to 0.8.6

ryansolid:
I have the latest version of Solid here. Trimmed some more stuff down. For the life of me cannot figure out why Solid's 10k has taken a hit on the new platform when the rendering techniques should be quite performant (uses same mechanisms as DomC and same reactivity engine as Surplus). Only thing I can think of is ES Proxies are especially heavy on those platforms. To that end I also have a Solid Signals version here which doesn't use Proxies. Now I'm not suggesting you merge 2 versions of the library. I've submitted the Proxy version to date since the code is slim (66 lines), knowing it wasn't the most performant version. But I want to understand if it's just Proxies causing the discrepancy here, to weigh if it is worth submitting the more performant version instead.

Thank you @krausest for humoring me.


krausest:
Results have been updated. Good to see that my new `npm run isKeyed` check reliably detected the missing aria-hidden in the solid-signals implementation (that I added before merging) :smiley: 
Do the results make sense to you?
Screenshot from 2019-07-01 22-03-18

ryansolid:
Yeah sorry about the `aria-hidden` thing. That version I'd been sitting on for quite a while before I added them in.  Numbers-wise Signals is about where I'd expect (next to DomC).

The proxy version seems to still be taking the hit. Given the MobX performance similarly seems relatively a bit worse(both versions) I'm gathering something about Proxies are not running as well here. Similarly Sinuous uses Object getters which also aren't doing quite as well, which explains how Fidan, and Surplus have jumped forward. Not Scientific by any means. Just looking.

All those libraries under the hood do almost the same things except in very specific areas. Technically Solid and Sinuous use the faster DOM render code for large repeated inserts like 10k using Node Cloning (same as domc, sifrr) but outside of the Signals version there is more overhead on setting up the reactive graph. This also adds overhead to partial updates. So either this overhead is proportionally higher (my suspicion) or Node Cloning is no longer that much quicker than `document.createElement`. If might be a little bit of both.

I guess I have a decision to make.

PR #604, opened 2019-07-06
Riot.js 4 keyed and not keyed examples

GianlucaGuarini:
This PR updates Riot.js to v4. It adds also the keyed example


krausest:
Thanks. Good to see riot.js has a keyed mode. Results look pretty good (ahead of react, angular, and vue.js), though maybe there's even a chance to speed up the select row benchmark?
Results have been [updated](https://krausest.github.io/js-framework-benchmark/current.html).

GianlucaGuarini:
@krausest thank you. I'd be curious to see the results of the latest 4.3.5 version. Do you need a PR for it?

krausest:
I‘ll try to bump the version and will report back...

krausest:
I updated to 4.3.5. Sorry to say it seems a bit slower.
Screenshot from 2019-07-11 21-58-32

GianlucaGuarini:
@krausest thank you for the quick update. Actually it does [less work](https://github.com/riot/dom-bindings/commit/dc24195a18ddad26e2cc2d1759ee44f3bcdcc0a6) so it should be faster but maybe I am missing something here 🤕

krausest:
Really hard to say. 
I actually ran 4.3.5 twice to see if I get a run with lower variance for select 1K, but both runs looked very similar (both keyed and non-keyed went up).

GianlucaGuarini:
@krausest ok thank you for the update and your patience. I will figure it out soon

krausest:
I switched only riot and riot compiler back to 4.3.2 and ran the benchmark again with the following result:
Screenshot from 2019-07-11 22-34-30
The 10k clear was faster than the 4.3.2 run above (yep, there are measurement errors...). The other benchmarks look similar. 
 Still the relative difference between 4.3.2 and 4.3.5 stays and it's still the select 1k benchmark that makes 4.3.5 slower.
Let me know when you've got a new version to check.

krausest:
Another note: The observed deviation for select 1k is very high (I's say too high to compute a reasonable average). The actual measured values for 4.3.5 were:
67.212, 78.232, 80.457,  84.44, 81.412, 79.365, 80.976, 128.188, 68.883, 164.078
For 4.3.2 they were closer:
75.974, 101.62, 112.703, 75.182, 71.45, 70.869, 77.633, 80.027, 77.898, 75.736

PR #606, opened 2019-07-06
12.0.0 and row selection

hville:
Minor update to bring it up to the latest 12.0.0 version and to make the selection "more direct" like other implementation.

This is more of a library than a framework so I have no issues if it is deemed not relevant to this benchmark.

As before, the first design goal is to have a light and efficient implementation that is backward compatible with older browsers and phones.

Merci!

### DIFF
--- a/frameworks/keyed/attodom/package.json
+++ b/frameworks/keyed/attodom/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-attodom",
-  "version": "0.11.3",
+  "version": "0.12.0",
   "description": "Benchmark for attodom",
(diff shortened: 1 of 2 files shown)


krausest:
I think the direct setting of the className for the selected row in the event handlers is not compliant with the "... the initial rendering is able to render the selection state" rule. (For some discussion see e.g. https://github.com/krausest/js-framework-benchmark/pull/521). Would you agree?

hville:
Makes sense. Reverted to have all updates from store.

krausest:
Thanks. Has been merged. Results are here: https://krausest.github.io/js-framework-benchmark/current.html

PR #609, opened 2019-07-12
Migrate to reason jsx 3

intellild:
The JSX Record API for Reason React is deprecated
https://reasonml.github.io/reason-react/docs/en/jsx-2

### DIFF
--- a/frameworks/keyed/reason-react/.gitignore
+++ b/frameworks/keyed/reason-react/.gitignore
@@ -1,3 +1,4 @@
 lib
 **/*.bs.js
 .merlin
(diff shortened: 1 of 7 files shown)


krausest:
Thanks, but I'm afraid there seems to be something pretty basic wrong about this update. 
It has the following issues:
* Create 1,000 rows button exists twice
* Selecting a row doesn't highlight it
* Can't delete a row
* Swap rows doesn't work

Can you please check what's wrong? Thanks!

intellild:
@krausest all fixed

PR #611, opened 2019-07-22
Bump miso to 1.1.0.0

dmjio:
Bump to latest version.


dmjio:
@krausest ping :)

krausest:
Not so fast ;-)
I'm afraid the output HTML isn't correct. There's a missing aria-hidden="true" as reported (albeit so far with a misleading message) by `node dist/isKeyed.js keyed/miso`. Can you please add that?

dmjio:
Oh drat. Yes I will fix soon.

dmjio:
So I think something might be up here, the `aria-hidden="true"` property is being set and is present on the DOM. In the chrome console I get:
```
document.getElementsByTagName('span')[0].getAttribute('aria-hidden')
> "true"
```
can also see it in the elements tab in `chrome`
image

cross-referenced with the vanilla-js example as well.

dmjio:
Also, while running these tests, I ran into some `.gc` error. Going to check if this happened in the old tests too.

```
*** duration 208.449 upper bound  442.622
result miso-v1.1.0.0-keyed_09_clear1k_x8.json min 204.623 max 219.889 mean 209.28040000000001 median 208.347 stddev 4.1700707481341945
QUIT
START BENCHMARK. Write results?  true
benchmarking  { name: 'miso',
  fullNameWithKeyedAndVersion: 'miso-v1.1.0.0-keyed',
  uri: 'frameworks/keyed/miso/dist-bundle',
  keyed: true,
  useShadowRoot: false } 21_ready-memory
after initialized  21_ready-memory 1 miso
Benchmark failed { JavascriptError: javascript error: window.gc is not a function
  (Session info: chrome=75.0.3770.142)
    at Object.throwDecodedError (/Users/dmjio/Desktop/js-framework-benchmark/webdriver-ts/node_modules/selenium-webdriver/lib/error.js:550:15)
    at parseHttpResponse (/Users/dmjio/Desktop/js-framework-benchmark/webdriver-ts/node_modules/selenium-webdriver/lib/http.js:542:13)
    at Executor.execute (/Users/dmjio/Desktop/js-framework-benchmark/webdriver-ts/node_modules/selenium-webdriver/lib/http.js:468:26)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  name: 'JavascriptError',
  remoteStacktrace:
```

krausest:
window.gc should be enabled with the chrome option "--js-flags=--expose-gc", but I remember some closure issue https://github.com/krausest/js-framework-benchmark/pull/225. Maybe it‘s something similar.
I‘ll check this evening.

dmjio:
Ah, nice. So `window.gc()` is not present. How do I run `chrome` with those flags ? I just run `npm run bench keyed/miso` and it loads up chrome. I tried running with `npm run bench keyed/miso --js-flags=--expose-gc` but got the same thing . Should I put `--js-flags=--expose-gc` in the `package.json` somewhere?

```
Benchmark failed { JavascriptError: javascript error: window.gc is not a function
```

dmjio:
@krausest also tried 

```javascript
"bench": "cross-env LANG=\"en_US.UTF-8\" node dist/benchmarkRunner.js --js-flags=--expose-gc", 
```

to no avail.

dmjio:
Strangely enough, this did not occur in the old tests (v. `0.18`). I don't think closure compilation would affect this since `window.gc` can't be removed.

krausest:
The flag is set here https://github.com/krausest/js-framework-benchmark/blob/master/webdriver-ts/src/webdriverAccess.ts#L203 . There's no way of passing options to chrome via the npm run bench task. As I said I'll take a look this evening.

frameworks/non-keyed/miso/src/Main.hs
```diff
@@ -0,0 +1,306 @@
...
+            [ text (S.ms rId) ]
+        , td_
+            [ class_ "col-md-4" ]
+            [ a_ [class_ "lbl", onClick (Select i)] [text (rowTitle r)]
+            ]
+        , td_
+            [ class_ "col-md-1" ]
+            [ a_
+              [ class_ "remove" ]
+              [ span_
```
krausest: aria-hidden="true" is missing here, isn't it?

krausest: And similarly in the keyed version. (Leaving out the aria-hidden attribute in the table might be taken as a shortcut,  I currently do not test it for the six buttons)

dmjio: You fixed it ! Thanks ;) @krausest

krausest:
Regarding the window.gc issue. I really think it's a closure problem.
Start chromium like that `chromium --js-flags=--expose-gc`
Open any implementation like http://localhost:8080/frameworks/non-keyed/vanillajs/
Type in the console: `window.gc`:
Screenshot from 2019-07-23 20-01-19
Then open miso http://localhost:8080/frameworks/keyed/miso/dist-bundle/
Screenshot from 2019-07-23 20-03-11
Which can't be called:
Screenshot from 2019-07-23 20-03-48
Would be interesting to see if other frameworks using closure are also affected, but happen to assign a function to gc...

If you look into the generated code you find:
Screenshot from 2019-07-23 20-09-19
I suspect this is causing the issue (but I don't have a idea what to do about it).

dmjio:
@krausest here's an idea, we could cat something like `window['gc']` into the js.

dmjio:
@krausest so doing this before closure compilation works.

```bash
 echo "(window['gc'] = window['gc']);" >> $out/bin/miso-benchmark-keyed.jsexe/all.js
```

It seems we're not completely out of the trap though, a few tests failed. Is this ok?
```
================================
The following benchmarks failed:
================================
[error-miso-03_update10th1k_x16.png]
{ name: 'TimeoutError', remoteStacktrace: '' }

[error-miso-04_select1k.png]
{ name: 'TimeoutError', remoteStacktrace: '' }

run was not completely sucessful Benchmarking failed with errors
```
If so, ready to go I think.

krausest:
Those tests check whether the benchmarks can be run. So it's a very bad sign...

Can you insert a script "window.saveGCFromClosureOverwriting = window.gc" in index.html before loading all.min.js? I could then adjust the benchmark runner to check for existence of that funny named variable and invoke that and all.min.js may do whatever it wants with windows.gc.

dmjio:
> Those tests check whether the benchmarks can be run. So it's a very bad sign...

Yea, I'll look into this more, will it keep us from merging ? (I think the old benchmarks had the same issue, but I can double check). It seems the error was `Timeout` would it be feasible to increase this threshold?

I'll try adding `window.saveGCFromClosureOverwriting = window.gc`

krausest:
I won't merge when the benchmark doesn't run (which is the case when those tests fail). The timeout doesn't happen due to too short amount of time, but due to not performing the expected dom operations (i.e. update and select do not work correctly this should be easily to test in the browser). 
I hope the tests work again when that gc issue is solved. From the logs above it seems like it worked earlier, the aria-hidden test is performed after and only when the benchmark tests work).
I'll take a look at the current commit this evening.

krausest:
Good news. Both keyed and non-keyed pass all tests! (Would be interesting why they don't on your machine)
Both happen to leave window.gc undefined, so the memory benchmarks work too.
I'll merge them and report the results here.

dmjio:
@krausest sounds great thanks. Yea I'm unsure as to why they fail on my machine, I'll attempt to re-run. I'm on OSX Version 75.0.3770.142 (Official Build) (64-bit)

krausest:
[Results](https://krausest.github.io/js-framework-benchmark/current.html) have been updated. Miso did improve but is still far on the right side.

dmjio:
@krausest awesome, thanks again for this. Surprised keyed outperformed non-keyed in these tests. Working on changing some data structures to improve the number, without cheating too much. Will be back ;)

krausest:
You're welcome. Would be nice if you could then add that window.saveGCFromClosureOverwriting in the index.html in the next PR.

dmjio:
Will do!

PR #612, opened 2019-07-22
Build projects in parallel

flaviut:
I have a lot of cores, and I'd like to build everything at the same time.

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,8 +1,12 @@
 var _ = require('lodash');
+var util = require('util');
 var exec = require('child_process').execSync;
 var fs = require('fs');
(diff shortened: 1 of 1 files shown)


krausest:
Thanks, but that means 137 builds are running concurrently, doesn't it?
If so I'd really prefer to have a configurable amount of worker threads that perform the build.

flaviut:
Closing this is likely for the best.

137 concurrent builds works fine for me, but it is unlikley to be right for most users.

Doing this properly would either require bringing in another dependency or me spending a significant amount of time figuring out the logic to run a certain number of promises in parallel.

PR #614, opened 2019-07-24
Update solid, ko-jsx, mobx-jsx

ryansolid:
Pretty big changes. I split out the control flow from the reconciler. This, in theory, has to be a de-optimization but it really opens up the flexibility of the libraries. I was pushing my opinion too strong on these libraries in the name of performance.  Now I can use the underlying renderer to fit whatever fine-grained API makes sense for the library. This commit exemplifies that:

* Solid has new JSX Component control flow
* MobX JSX now uses MobX's built in mutable API, and uses a Class Component with Decorators.
* Knockout JSX uses a custom fn method for memoMapping

Overall the logic is much simpler now, but there is some duplicate work that needs to be done between the reactive systems, and the generalized renderer. I've done a bunch of work to try to optimize performance despite this. Replace rows has taken a small hit, but most benchmarks have stayed relatively the same. I believe I've even managed to gain ground in other areas (like initial render). But I'd like to verify that.

### DIFF
--- a/frameworks/keyed/ko-jsx/package.json
+++ b/frameworks/keyed/ko-jsx/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-ko-jsx",
-  "version": "0.6.3",
+  "version": "0.8.1",
   "main": "index.js",
   "js-framework-benchmark": {
     "frameworkVersionFromPackage": "ko-jsx"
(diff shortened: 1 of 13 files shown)


krausest:
[Wow](https://krausest.github.io/js-framework-benchmark/current.html).

ryansolid:
Thank this is great news. I hadn't slept for like 36 hours now tweaking the last little things. I'm glad it paid off.

PR #615, opened 2019-07-24
Updated Inferno 7.2.0, changed minifier to TerserJS from UglifyJS.

Havunen:


krausest:
[Results](https://krausest.github.io/js-framework-benchmark/current.html) have been updated.

Issue #574, opened 2019-06-16
Update frameworks - help needed


krausest:
The following frameworks could not be updated with a "ncu -u" approach:

Keyed implementations:
- [ ] angular-ivy
- [ ] aurelia
- [ ] bobril
- [ ] domc
- [ ] glasgow
- [ ] hyperoop
- [ ] ivi
- [ ] marko
- [ ] rax
- [ ] react-djinn
- [ ] react-mobX
- [ ] stage0

krausest:
@Freak613 I updated domc. Do you want to review the changes?

Freak613:
@krausest no worries, I just reviewed them, looks fine.
Thanks for update!

PR #617, opened 2019-07-24
Replace Dio with Dyo

thysultan:
Regarding https://github.com/krausest/js-framework-benchmark/issues/574 dio -> dyo


krausest:
Thanks, both for updating dyo and even more so for removing the non-keyed implementations!
I had a small issue with the clear implementation resetting the id which caused an error in the test driver. 
Please review 09e547ab0d87b370c4b485b11c175bf5feae98e2 and send a PR if it's not okay for you.
[Results](https://krausest.github.io/js-framework-benchmark/current.html)  have been updated.

thysultan:
Thank you, it looks good.

PR #618, opened 2019-07-25
Miso 1.1.0.0

dmjio:
Tried a new data structure, added the `saveGCFromClosureOverwriting`.


dmjio:
@krausest it times out for me on a few of them. Will be interested to know if that's the case for you.

dmjio:
@krausest ping :)

dmjio:
where'd he go :)

ryansolid:
While @krausest is fairly responsive, often it will take up to 7-10 days for these merges to happen. Recently due to migration he's been working on this more so merges have been happening more often. Merging generally means quick check over the implementation and manually rerunning the tests on his computer. Sometimes this means a few runs to ensure low variance runs. I imagine it makes it way more convenient to run several updates in batch. But also means putting a side a couple hours to do so and screencap results and share them.

There have been attempts to reduce the overhead of maintaining and they've largely been successful. But as long as we expect the quality of results we get back there is going to be manual work. And that takes finding time. So I wouldn't be surprised on the occasional weekend or perhaps week in the summer the results won't be updated for several days. Historically he's been pretty good notifying people if it's going to be over a week. But anything under that comes with the territory.

I'm not sure he has written anything like this anywhere. And maybe viewed as unnecessary. But he maintains this by himself so overtime the work only goes up.

krausest:
Yes, @ryansolid is absolutely right. I'm trying to batch updates depending on my lust for merging.

In this case your implementation seems buggy. Update every 10th row doesn't update the first row, so I can't merge.

dmjio:
Thanks for the feedback, I'll check it out.

dmjio:
@krausest closing in favor of #657

PR #619, opened 2019-07-25
Updated to Inferno 7.2.1 and clean up controller.jsx

Havunen:
Updated Inferno to 7.2.1, because 7.2.0 had a bug where events were not patched correctly.
Removed old comments from inferno controller.jsx file and fixed wrong title in non-keyed implementation. Also removed `$HasVNodeChildren` optimization flag because its done automatically by JSX plugin.


krausest:
Thanks. [Results](https://krausest.github.io/js-framework-benchmark/current.html)  have been updated.

PR #620, opened 2019-07-28
CRUI benchmark

Iazel:
Add a benchmark for CRUI library. I hope it was correct to put it in "non-keyed", but please be aware that this lib do not use a VDOM, so the keyed concept doesn't really matter.

### DIFF
--- a/frameworks/keyed/crui/index.html
+++ b/frameworks/keyed/crui/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
+    <title>CRUI keyed</title>
+    <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+<div id="main"></div>
+<script src="dist/main.js"></script>
+</body>
+</html>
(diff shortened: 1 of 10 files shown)


ryansolid:
The "keyed" concept here is more generalized than the "key" property you find on React JSX elements. It just means that the data object is tied to the DOM node. Like if you delete a row in the array you delete the corresponding DOM Element and all the items would just be one location higher in the DOM. The "non-keyed" implementation would shift all the data up, repurposing existing DOM nodes, and only delete the last element. It is possible your implementation is already "keyed" depending on how your library works. There is a test to check if your implementation is "keyed".

Generally, we encourage "keyed" implementations since there are many places where "non-keyed" has unexpected side effects.  This is around transitions, animations, custom elements, and so on. Also to be fair the "non-keyed" scenario is more uninteresting from a benchmark scenario. Most libraries that can do "keyed" can do "non-keyed", but vice versa isn't necessarily true.

I'd strongly encourage a "keyed" implementation since it is generally the one that gets included in comparison and is the more defacto view of Framework performance.

Iazel:
Thanks @ryansolid, now the difference is clear. I will update the PR to move CRUI into `keyed` section then :)

krausest:
Ok - I'll leave the PR open until you've made that change.

Iazel:
@krausest moved implementation to `keyed` :)

krausest:
Thanks. Results are [here](https://krausest.github.io/js-framework-benchmark/current.html)
Feel free to disuss the results in issue #622 .

Issue #19, opened 2016-07-13
Highlighted row should stay selected when clicking swap rows, update and append


krausest:
The highlight handling for swap rows, update and append is not consistent among the frameworks. The highlighted row should stay selected when clicking swap rows.

krausest:
Fixed in the repo.

PR #621, opened 2019-07-28
Update Sinuous v0.12.6

luwes:
Shouldn't affect performance too much, might be a bit faster because

- `map` uses an Array instead of Map to keep track of the disposers
- in `map` full replace used to use a loop of `removeChild` calls instead of a simple `.textContent = ''` to clear which is possible since there are no sibling elements.

### Fixed

- Fixed naming issue for umd and iife bundles
- Add observable duplicate edges check
- Replace Map w/ array for `map` disposers

### Fixed

- Fixed [Plain array in root doesn't return `DocumentFragment` #19](https://github.com/luwes/sinuous/issues/19)

### Fixed

- Fixed [template` inserts with duplicate value are empty #14](https://github.com/luwes/sinuous/issues/14)

### Fixed

- Fixed [Incorrect updating of observable #18](https://github.com/luwes/sinuous/issues/18)

### Changed

- Upgraded `sinuous/htm` and `sinuous/babel-plugin-htm` to latest version.

### Added

- Added support for plain array inserts again w/ no size increase.

### Changed

- Make `map` return a DocumentFragment.
- Optimized and re-used code for full replace in `map`.

### Added

- Added SVG support.

### Removed

- Removed plain array inserts. Fragments can be used for this aka

```js
  html` 
    <h1>Title</h1>
    <p>Some text</p>
  `
```

### DIFF
--- a/frameworks/keyed/sinuous/package.json
+++ b/frameworks/keyed/sinuous/package.json
@@ -17,7 +17,7 @@
     "url": "https://github.com/krausest/js-framework-benchmark.git"
   },
   "dependencies": {
-    "sinuous": "0.11.4"
+    "sinuous": "0.12.6"
   },
   "devDependencies": {
     "@babel/core": "7.4.4",


ryansolid:
> in map full replace used to use a loop of removeChild calls instead of a simple .textContent = '' to clear which is possible since there are no sibling elements.

This should be noticeable difference. I had a bug a few versions back I wonder if that had propagated into your codebase and you never got the fix.

luwes:
@ryansolid I think it might have been in this commit
https://github.com/luwes/sinuous/commit/a9036e508a14ccf57ecf7fc70bd8ad749f443ad9

ryansolid:
Yeah, that was all you. In any case, I think that should be a nice boost in benchmark 2.

krausest:
[Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated. No need to look far to the right :+1: !

PR #623, opened 2019-08-10
Update Moon to v1.0.0-beta.4

kbrsh:
Updates Moon to the latest v1 beta. This new beta has a lot of changes to the API and diffing algorithms (it uses a virtual DOM instead of compiled DOM operations).

It uses a memoized `<Row/>` component. Moon doesn't support something like `shouldComponentUpdate` because every component is just a pure function, so standard techniques like memoization are encouraged for performance.


krausest:
Thanks. Nice [results](https://krausest.github.io/js-framework-benchmark/current.html) for moon. Does moon support keyed updates (my opinion about non-keyed only fameworks isn't that good).

kbrsh:
Woah, wasn't expecting such a big jump haha, that's awesome.

As for keyed updates, Moon is non-keyed because there is no concept of local state for components or DOM nodes. Instead, there is a data driver which manages a global state tree and you manage it yourself, similar to Elm. Instead of reordering DOM nodes to preserve local state, you set properties like `focus`, `value`, or `scrollTop` yourself.

Because of this purely functional nature, Moon doesn't _need_ keyed updates to preserve local DOM state. They might be added in the future as an optimization for reordering nodes instead of updating all of them, but not for the reasons they are typically required for today.

PR #626, opened 2019-08-24
Update Mithril to 2.0.4

slegrand45:


krausest:
Thanks - [results]( https://krausest.github.io/js-framework-benchmark/current.html) have been updated
Seems like mithril 2 got a bit faster!

PR #629, opened 2019-08-30
Upgrade the Benchmark for Dojo 6

agubler:
Updates both the keyed and non-keyed benchmarks to Dojo 6. 

Also renamed the `dojo2` directories to `dojo`, let me know if there are other places I need to update (unfortunately, now it says it's all new code, thanks github 😄).


krausest:
Thanks - [results]( https://krausest.github.io/js-framework-benchmark/current.html) have been updated
Scores a better than the for the [old](https://krausest.github.io/js-framework-benchmark/2019/table_chrome_76.html) version, though it looks like create rows got a bit slower. Did you expect that?

agubler:
Thanks @krausest, i don't think there's anything too surprising there to be honest 👍

PR #631, opened 2019-09-06
Update Riot.js version

GianlucaGuarini:
Update Riot.js to v4.5.0


krausest:
Thanks. Results are [updated](https://krausest.github.io/js-framework-benchmark/current.html). Sadly it seems a bit slower than 4.3.5, especially in the select row benchmark. (I ran the benchmark twice to eliminate the chance of a CPU-hogging background process)

GianlucaGuarini:
thank you for the feedback and for merging it

PR #635, opened 2019-09-08
Update Mikado Version

ts-thomas:
I discovered some overhead in the render pipeline. This is fixed in v0.0.7, although it is not very urgent.

### DIFF
--- a/frameworks/keyed/mikado/package.json
+++ b/frameworks/keyed/mikado/package.json
@@ -24,7 +24,7 @@
     "build-prod": "npm run build"
   },
   "dependencies": {
-    "mikado": "0.0.69"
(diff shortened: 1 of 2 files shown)


ts-thomas:
I have repeated the benchmark and there is a noticeable improvement (in this area of course). The issue has affected the render pipeline, the biggest of all hotspots.

krausest:
Results are [updated](https://krausest.github.io/js-framework-benchmark/current.html) but I'm afraid I can't see much of an improvement (but of course the air is pretty thin on the left side of the table :smile: )

ts-thomas:
Thanks a lot, the results differ from my machine, but that is absolutely fine :)

PR #639, opened 2019-09-11
Update Mikado to 0.1.2, cleanup sources

ts-thomas:
I followed the suggestion to render the whole page and removing the static html from index.html and I also cleaned up the directory.  I use an "external/global state" now, which auto references to the template render process. This test implementation is a good tutorial to which I would like to forward.


ts-thomas:
Update: 1.1 kB smaller filesize

krausest:
Thanks - it's merged. I can't publish results right now, since chrome was updated and I have to rerun all frameworks for chrome 77. First runs are alarming - my system with chrome 77 is about 10% slower for vanilllajs create rows. I'll have to take a closer look at that first.

krausest:
Congratulations! Results are available: #641 
Though one could say that it already was much faster... What's going on with chrome? :weary:

PR #648, opened 2019-10-15
Add fntags framework v0.1.7 keyed impl

narcolepticsnowman:
fntags uses hyperscript style templates to build and interact with the dom directly instead of using a virtual dom.

State is implemented using the observer pattern on an es6 proxy.

https://narcolepticsnowman.github.io/fntags

I've also added the --installed option to benchmarkRunner.ts, to run all frameworks that have a package-lock.json file. This is much more convenient than listing out all the frameworks to run, as you can just go install the ones you want and run them.

If you'd prefer I didn't add this option, let me know and I'll happily delete it.


krausest:
[Results](https://krausest.github.io/js-framework-benchmark/current.html) have been updated.

Issue #651, opened 2019-10-18
URL is not defined


mindplay-dk:
I followed the instructions in step 4 of the README to get the automated benchmark driver running, and everything installed without errors.

I've successfully built and run the `keyed/preact` benchmark via the web-server.

I then tried to run the benchmark under the `webdriver-ts`, and it fails:

```
$ npm run bench keyed/preact

> webdriver-ts@1.0.0 bench /mnt/c/workspace/js-framework-benchmark/webdriver-ts
> cross-env LANG="en_US.UTF-8" node dist/benchmarkRunner.js "keyed/preact"

/mnt/c/workspace/js-framework-benchmark/webdriver-ts/node_modules/lighthouse/lighthouse-core/lib/url-shim.js:35
class URLShim extends URL {
                      ^

ReferenceError: URL is not defined
    at Object.<anonymous> (/mnt/c/workspace/js-framework-benchmark/webdriver-ts/node_modules/lighthouse/lighthouse-core/lib/url-shim.js:35:23)
    at Module._compile (internal/modules/cjs/loader.js:654:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:665:10)
    at Module.load (internal/modules/cjs/loader.js:566:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:506:12)
    at Function.Module._load (internal/modules/cjs/loader.js:498:3)
    at Module.require (internal/modules/cjs/loader.js:598:17)
    at require (internal/modules/cjs/helpers.js:11:18)
    at Object.<anonymous> (/mnt/c/workspace/js-framework-benchmark/webdriver-ts/node_modules/lighthouse/lighthouse-core/lib/network-request.js:14:13)
    at Module._compile (internal/modules/cjs/loader.js:654:30)
```

Looks like a similar problem was reported and fixed earlier in #593.

But it looks like we're still missing a `URL` polyfill for node or something?

krausest:
Sorry, seems like I should update the README and require at least node 10.

krausest:
README is updated. Can you try with the current LTS (10.16.3)?

mindplay-dk:
Okay, so now it starts, but just goes into endless errors and retries... 🤷‍♂️

```
$ npm run bench keyed/react-hooks

> webdriver-ts@1.0.0 bench /mnt/c/workspace/js-framework-benchmark/webdriver-ts
[...]

PR #652, opened 2019-10-18
add missing measurements with console-output

mindplay-dk:
Other benchmarks show an approximate measurement in the browser console - this one didn't.

### DIFF
--- a/frameworks/keyed/react-hooks/src/main.jsx
+++ b/frameworks/keyed/react-hooks/src/main.jsx
@@ -1,4 +1,4 @@
-import React, { memo, useReducer, useCallback } from 'react';
+import React, { memo, useReducer, useCallback, useEffect } from 'react';
 import ReactDOM from 'react-dom';
 
 function random(max) { return Math.round(Math.random() * 1000) % max; }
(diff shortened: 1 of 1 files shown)


ryansolid:
Yeah, I was wondering where we stand on this. Almost none of the top 30 implementations actually use the console.log, just to cleanup code and reduce the tiniest overhead. I just stopped adding it to implementations I've done. Does anyone have an opinion on this?

leeoniya:
we should drop them entirely. in the best case they're still wrong.

krausest:
I tried to clarify that in the rules
https://github.com/krausest/js-framework-benchmark/commit/af98c85b40c90b6abf06ecc1bd8a6d3f8f287fc5
I want to keep it in the vanillajs implementation, but it can be dropped anywhere else. It serves as a sanity check for me whenever the next chrome version yields weird results (as it regularly does). 
@mindplay-dk Thus I'm closing this PR.

mindplay-dk:
Yeah, it's totally unreliable of course, I agree - just copy/pasting that into every benchmark is obviously not the right approach...

(I mostly did it because I couldn't get the real benchmark to run - but I understand from your answer in #651 that's because I need to upgrade node to v10...)

Issue #126, opened 2017-02-11
Why isn't jQuery in this Benchmark?


paranoid-android:
This benchmark tool is great! I want you to know you've done some great work and I don't want the issue I'm reporting to detract from what's been done.

Still, I'm surprised jQuery hasn't been included in these benchmarks. I can see an argument being made that it's a not a framework like the others that are benchmarked. However, jQuery can create / read / update rows. For that reason, I'm a bit surprised it's not included. I personally think the addition of jQuery 3.1.1 (`https://code.jquery.com/jquery-3.1.1.min.js`) would be a great addition to this benchmark.

leeoniya:
> I personally think the addition of jQuery 3.1.1 would be a great addition to this benchmark.

Most implementations are submitted as PRs by various authors. You are welcome to submit a jQuery version for inclusion.

trotyl:
> I can see an argument being made that it's a not a framework like the others that are benchmarked.

The problem here is not about whether something is a framework, but whether something is data-driven.

What actually being tested is **how good a library/framework finding out what actual DOM manipulation is by data**, jQuery is not data-driven, so you do can test what's the slowdown of jQuery compare to the VanillaJS, but it does not make much sense to compare jQuery with other data-driven framework.

Besides, in most data-driven frameworks, it's still available for direct DOM manipulation, but that's not what being covered for this benchmark.

krausest:
I must say I'm not particularly interested in adding jQuery. The purpose of this benchmark is to measure the cost of a framework over vanillaJs. And there are many good reasons why one would not use vanillaJs for a real world app. JQuery should be close to vanillaJs and if it's not close, zepto or some low level dom operation mixed with jQuery will be. The existence of vanillaJs should not be interpreted as a recommendation to use it for your next project but just to serve as a baseline (and in some cases it's already behind ;-)

aminya:
Can we open this again? It is not about "the personal interest or preference", it is about having a reference for comparing the JavaScript DOM libraries. I see everywhere people bash jQuery and say that its performance is a bottleneck. 

If this benchmark framework wants to be a reference, it should definitely include jQuery.

PR #654, opened 2019-10-27
Replace swap built-in with reconcile

ts-thomas:
After I got the 3rd shit storm about this swap method I change this built-in method with reconciliation. Please be aware that the keyed test will not become a reconcile test. The first is a paradigm the other is just an implementation detail. It's the same group of library authors which have made its own reconcile libs (which are quite equal^^) and of course they would like to have the best environment for its kind of library. But the truth is that reconcile isn't the holy grail and certainly by far not a standard.

But I would really like to see a 3rd test category especially for this purpose (besides keyed/non-keyed). If you like that idea I would also like to make some suggestions for this new section:

- please please please allow runtime optimizations for the tests instead of reloading, because that isn't natural
- change the concept of test so that each library has to export 1 single function which handles all the data, that will:
  - give you more flexibility in updating/changing tests in the future
  - adding new tests without the need of changing any implementation
  - provides a quite more fairly test base
  - a most simple test implementation for a developer
  - the only way to "force" benchmark library performance and not the performance made by an implementation
  - I do already that way, if you like you can get some suggestions here:
    - https://github.com/nextapps-de/mikado/tree/master/bench
    - look at this tests for example: https://github.com/nextapps-de/mikado/tree/master/bench/test they all have an index.html where a single function is exported to the benchmark test

__Note:__ just the keyed test was updated, the non-keyed stayed as it is

### DIFF
--- a/frameworks/keyed/mikado/package.json
+++ b/frameworks/keyed/mikado/package.json
@@ -14,14 +14,14 @@
   },
   "scripts": {
     "compile": "mikado-compile src/template/app.html && mikado-compile src/template/item.html && echo Compile Complete. && exit 0",
-    "build": "npm run compile && node task/build RELEASE=custom DEBUG=false USE_POLYFILL=false SUPPORT_CACHE=false SUPPORT_EVENTS=true SUPPORT_STORAGE=true SUPPORT_HELPERS='swap' SUPPORT_ASYNC=false SUPPORT_TRANSPORT=false SUPPORT_TEMPLATE_EXTENSION=false SUPPORT_REACTIVE=false SUPPORT_CACHE_HELPERS=false SUPPORT_POOLS=false SUPPORT_COMPILE=false && exit 0",
+    "build": "npm run compile && node task/build RELEASE=custom DEBUG=false USE_POLYFILL=false SUPPORT_CACHE=false SUPPORT_EVENTS=true SUPPORT_STORAGE=false SUPPORT_HELPERS=false SUPPORT_ASYNC=false SUPPORT_TRANSPORT=false SUPPORT_TEMPLATE_EXTENSION=false SUPPORT_REACTIVE=false SUPPORT_CACHE_HELPERS=false SUPPORT_POOLS=false SUPPORT_CALLBACKS=false SUPPORT_COMPILE=false && exit 0",
     "build-prod": "npm run build"
   },
   "dependencies": {
-    "mikado": "^0.6.53"
+    "mikado": "0.7.35"
   },
   "devDependencies": {
-    "google-closure-compiler": "20191012.0.0-nightly",
+    "google-closure-compiler": "20191027.0.0-nightly",
     "mikado-compile": "0.6.5"
   }
 }
--- a/frameworks/keyed/mikado/src/data.js
+++ b/frameworks/keyed/mikado/src/data.js
@@ -10,14 +10,14 @@ let _nextId = 1;
 
 export function buildData(count){
 
-    if(count === 1){
-
-        return {
-
-            id: _nextId++,
-            label: ADJECTIVES[_random(len_ADJECTIVES)] + " " + COLOURS[_random(len_COLOURS)] + " " + NOUNS[_random(len_NOUNS)]
-        }
-    }
+    // if(count === 1){
+    //
+    //     return {
+    //
+    //         id: _nextId++,
+    //         label: ADJECTIVES[_random(len_ADJECTIVES)] + " " + COLOURS[_random(len_COLOURS)] + " " + NOUNS[_random(len_NOUNS)]
+    //     }
+    // }
 
     const data = new Array(count);
 
--- a/frameworks/keyed/mikado/src/main.js
+++ b/frameworks/keyed/mikado/src/main.js
@@ -5,22 +5,28 @@ import { buildData } from "./data.js";
 
 Mikado.once(document.getElementById("main"), app);
 
+let data;
 const state = { "selected": {} };
 const root = document.getElementById("tbody");
-const mikado = Mikado.new(root, item, {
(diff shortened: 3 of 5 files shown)


ryansolid:
@ts-thomas you can do what you want but yes people are going to point out when things are not equivalent. Swap might be the most obvious, but this whole benchmark was setup to compare data-driven libraries. It started with mostly Virtual DOM libraries and continued on to other libraries that can operate similarly. So changing Swap doesn't completely do it. Your Run leverages the fact that you explicitly clear since you know that Replace is a full Replace. Other libraries actually check. Similarly with Append. You Append when other libraries actually check what has changed. I understand it's a by-design thing but it's because they are data-driven. As long as Mikado is singularly different why would anyone take it to be equivalent? Especially when these are easy things that any library could do (like explicitly clearing before replace).

Yes reconciliation is an implementation detail here but it's the very thing that allows comparison. It is literally the only thing we could somewhat agree upon, after tons and tons of discussions. (#430 #230, #126, #104, #22) and while we are at it (#521, #519). 

> But the truth is that reconcile certainly by far not a standard.

Undeniably here it is. And I think that it would be safe to actually say in general it is standard. With the exception of jQuery and Backbone(and maybe Bootstrap if you count that) the top 10 popular JavaScript UI frameworks/libraries use this approach (React, Angular, Vue, Ember, Svelte, Mithril, and so on).

As to the suggestion of this 3rd category. I am not opposed to in general as a way to put aside reference builds but they sort of uninteresting because everyone will just pretty much match Vanilla JS performance. It is already getting so close with reconciled libraries given the overhead in the DOM vs the actually JavaScript. But you can see other overheads here, like Web Components and wrappers. And people can work towards making the idealized Vanilla implementation. However to the suggestions specifically:

>please please please allow runtime optimizations for the tests instead of reloading, because that isn't natural

This is an interesting idea but hard to standardize on or take meaning from. Not sure. A few of the tests have warmup runs so that sort of accounts for a handful of these actions happening. I mean what do think is more natural a page section loading once and being swapped out a couple of times, or that views of the same shape are blasted over and over in succession? It depends on the app I suppose but this sort of optimization feels like view windowing for large lists. It's a thing but it is very specific and is situationally desireable.

> change the concept of test so that each library has to export 1 single function which handles all the data, ... 

I'm sure this was discussed before. You are describing a large category of benchmarks (I'd say most). I call them data snapshot benchmarks. That just isn't what this benchmark is. Data Snapshot benchmarks have just as many quirks in how data is handled, and it doesn't allow libraries that manage their own data primitives to compete. The reason this benchmark is popular is it starts from a natural place. A basic list app. Something that any library could picture implementing in a generic way and then runs it through simulated user action like tests.  Going the snapshot way not only alienates libraries more so, it removes the basis of this at all resembling actual library usage. What I love about this benchmark is I can go in and look at the code and "get" how writing a simple app would look like. For many libraries having a single function driving data is not natural even remotely. And let's face it personally if I was looking for that I would point you to the very good [UIBench](https://localvoid.github.io/uibench/) by ivi's author @localvoid. 

I guess what I'm trying to say is what makes this Benchmark good is it being exactly the way it is. You can go plenty of other places to get those tests. There is value in being able to showcase libraries this way. It's the same reason I'm like do whatever with Mikado implementation. Since all I have do is look at the implementation to understand oh look there is Append and Swap, this library is doing things a little differently. I will have to decide what I think about it. 

I've written a recent article on medium on the topic (here is the link that gets you past the paywall). (Understanding JavaScript UI Benchmarking)[https://medium.com/javascript-in-plain-english/understanding-javascript-ui-benchmarking-afe4278b4b46?source=friends_link&sk=8ad46ce5ee36da0ee9d4c7fd786f03ee]. I hope that it can serve as a good introduction for people coming confused by a lot of the discussion going on here. It has links to many benchmarks in different styles to help people understand the differences.

ts-thomas:
There was nothing to read about a strict data-driven implementation. Also nothing to read about to strictly use the 1-single-magic-reconcile function to do all tests. So what you like to have and what exist are 2 different things.

ts-thomas:
The best of all is I changed this stupid thing and now you continue this discussion. Could you remember on my first post? I supposed that become a never ending story. As almost, I was absolutely right.

ts-thomas:
"and it doesn't allow libraries that manage their own data primitives to compete" that depends on the implementation, I also solved this by providing 3 different modes in my benchmark which basically control how is data created and passed to the render functions ... and it was not that complicated.

ts-thomas:
The uibench is not documented, that's why I did not implement it. Btw also not transparent enough for me that I would thrust them. I wonder if it is so hard to provide a simple, clear and transparent benchmark environment. And that should the basic requirements which every benchmark should have, otherwise it makes it useless.

ts-thomas:
"but every few months there is some new framework believing they cracked the performance boundary by inventing these methods" ... lol, that shows me that you did not know anything about me. When I build a lib, it is supposed to get the best performance and usability by a large margin, otherwise I would not start it. Don't be fules, remember my name, you will thank me later.

ts-thomas:
That you all celebrate yourself all the time is really annoying. Always the same names, always the same libs, always the same crap ... its like a tour in a carousel

ts-thomas:
A last word to you @ryansolid your screenshot in your article did not show the first top ranked and wonder o wonder begins with solid ... You are the one who blinds others with false facts. I'm done with the topic.

ryansolid:
Until Chrome 78 came there were performance issues with Chrome 76/77 so I just grabbed a previous screenshot that I happened to have from Chrome 75 before Mikado was submitted. You are correct it's subtle advertisement but has nothing to do with the content of the article. I clipped off what the benchmarks were refering to etc.. It was just a stock image for all purposes.

ts-thomas:
I'm sorry, I took that too personally.

PR #655, opened 2019-10-29
Update Solid v0.13.0 (and KO JSX, MobX JSX)

ryansolid:
Haven't updated for a while so thought I'd submit. There have actually been a lot of changes. Primarily:
* Removal of explicit `{( )}` syntax for dynamic bindings. Now a heuristic is used to determine if the binding could be dynamic. It simplifies things greatly for end users but requires a bit more diligence for getting optimal performance (ko-jsx, mobx-jsx hoist out their list mapping now in the implementation)
* Batching of computations. No longer are bindings guarenteed to be one per computation. This substantially removes overhead when have many bindings on nodes (not applicable here).
* More robustness around downstream conditional resolution. Some fine-grained libraries acted in unexpected ways under nested contexts dependent on different but related data. This added more size and overhead on all computations to the library unfortunately.
* Unified templating mechanism for HTML and SVG and additional support for Server Side Rendering and Client Side Hydration.

As usual I've learned a couple of things along the way to improve performance, so it is my hope that the performance is fairly comparable to the previous version.

### DIFF
--- a/frameworks/keyed/solid-signals/src/main.jsx
+++ b/frameworks/keyed/solid-signals/src/main.jsx
@@ -1,4 +1,5 @@
-import { createRoot, createSignal, freeze, selectWhen } from 'solid-js'
+import { createSignal, freeze } from 'solid-js';
+import { render, selectWhen } from 'solid-js/dom';
 
 let idCounter = 1;
 const adjectives = ["pretty", "large", "big", "small", "tall", "short", "long", "handsome", "plain", "quaint", "clean", "elegant", "easy", "angry", "crazy", "helpful", "mushy", "odd", "unsightly", "adorable", "important", "inexpensive", "cheap", "expensive", "fancy"],
@@ -25,8 +26,10 @@ const Button = ({ id, text, fn }) =>
   </div>
 
 const App = () => {
+  let rowId;
   const [data, setData] = createSignal([]),
-    [selected, setSelected] = createSignal(null, (a, b) => a === b);
+    [selected, setSelected] = createSignal(null, (a, b) => a === b),
+    applySelection = selectWhen(selected, 'danger');
 
   return <div class='container'>
     <div class='jumbotron'><div class='row'>
@@ -41,14 +44,15 @@ const App = () => {
       </div></div>
     </div></div>
     <table class='table table-hover table-striped test-data'><tbody>
-      <For each={( data() )} transform={ selectWhen(selected, 'danger') }>{ row =>
-        <tr model={ row.id }>
-          <td class='col-md-1' textContent={ row.id } />
-          <td class='col-md-4'><a onClick={ select }>{ row.label }</a></td>
+      <For each={ data() } transform={ applySelection }>{ row => (
+        rowId = row.id,
+        <tr model={ rowId }>
+          <td class='col-md-1' textContent={ rowId } />
+          <td class='col-md-4'><a onClick={ select } textContent={ row.label() } /></td>
           <td class='col-md-1'><a onClick={ remove }><span class='glyphicon glyphicon-remove' aria-hidden="true" /></a></td>
           <td class='col-md-6'/>
         </tr>
-      }</For>
+      )}</For>
     </tbody></table>
     <span class='preloadicon glyphicon glyphicon-remove' aria-hidden="true" />
   </div>
@@ -106,4 +110,4 @@ const App = () => {
   }
 }
 
-createRoot(() => document.getElementById("main").appendChild(<App />))
+render(App, document.getElementById("main"));
--- a/frameworks/keyed/solid/src/main.jsx
+++ b/frameworks/keyed/solid/src/main.jsx
@@ -1,4 +1,5 @@
-import { createRoot, createState, selectWhen } from 'solid-js';
+import { createState } from 'solid-js';
+import { render, selectWhen } from 'solid-js/dom';
 
 let idCounter = 1;
 const adjectives = ["pretty", "large", "big", "small", "tall", "short", "long", "handsome", "plain", "quaint", "clean", "elegant", "easy", "angry", "crazy", "helpful", "mushy", "odd", "unsightly", "adorable", "important", "inexpensive", "cheap", "expensive", "fancy"],
@@ -24,6 +25,7 @@ const Button = ({ id, text, fn }) =>
   </div>
 
 const App = () => {
+  let rowId;
   const [ state, setState ] = createState({ data: [], selected: null }),
     run = () => setState({ data: buildData(1000), selected: null }),
     runLots = () => setState({ data: buildData(10000), selected: null }),
@@ -35,7 +37,8 @@ const App = () => {
     remove = (e, id) => setState('data', d => {
       const idx = d.findIndex(d => d.id === id);
       return [...d.slice(0, idx), ...d.slice(idx + 1)];
-    });
+    }),
+    applySelection = selectWhen(() => state.selected, 'danger');
 
   return <div class='container'>
     <div class='jumbotron'><div class='row'>
@@ -50,17 +53,18 @@ const App = () => {
       </div></div>
     </div></div>
     <table class='table table-hover table-striped test-data'><tbody>
-      <For each={( state.data )} transform={ selectWhen(() => state.selected, 'danger') }>{ row =>
-        <tr model={ row.id }>
-          <td class='col-md-1' textContent={ row.id } />
-          <td class='col-md-4'><a onClick={ select }>{( row.label )}</a></td>
+      <For each={ state.data } transform={ applySelection }>{ row => (
+        rowId = row.id,
+        <tr model={ rowId }>
+          <td class='col-md-1' textContent={ rowId } />
+          <td class='col-md-4'><a onClick={ select } textContent={ row.label } /></td>
           <td class='col-md-1'><a onClick={ remove }><span class='glyphicon glyphicon-remove' aria-hidden='true' /></a></td>
           <td class='col-md-6'/>
         </tr>
-      }</For>
+      )}</For>
     </tbody></table>
     <span class='preloadicon glyphicon glyphicon-remove' aria-hidden="true" />
   </div>
 }
 
-createRoot(() => document.getElementById("main").appendChild(<App />))
+render(App, document.getElementById("main"));
--- a/frameworks/keyed/ko-jsx/package.json
+++ b/frameworks/keyed/ko-jsx/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-ko-jsx",
-  "version": "0.8.1",
+  "version": "0.10.0",
   "main": "index.js",
   "js-framework-benchmark": {
     "frameworkVersionFromPackage": "ko-jsx"
@@ -17,16 +17,16 @@
     "url": "https://github.com/krausest/js-framework-benchmark.git"
   },
   "dependencies": {
-    "babel-plugin-jsx-dom-expressions": "0.11.2",
+    "babel-plugin-jsx-dom-expressions": "0.14.0",
     "knockout": "3.5.0",
-    "ko-jsx": "0.8.1"
+    "ko-jsx": "0.10.0"
   },
   "devDependencies": {
-    "@babel/core": "7.5.5",
-    "rollup": "1.17.0",
+    "@babel/core": "7.6.4",
+    "rollup": "1.26.0",
     "rollup-plugin-babel": "4.3.3",
     "rollup-plugin-commonjs": "9.2.0",
     "rollup-plugin-node-resolve": "5.2.0",
-    "rollup-plugin-terser": "5.1.1"
+    "rollup-plugin-terser": "5.1.2"
   }
 }
--- a/frameworks/keyed/ko-jsx/src/template.jsx
+++ b/frameworks/keyed/ko-jsx/src/template.jsx
@@ -17,6 +17,17 @@ const selectRow = (selected, rows) => {
 }
 
 export default function({data, selected, run, runLots, add, update, clear, swapRows, remove, select}) {
+  let rowId;
+  const items = selectRow(selected, data.memoMap(row => (
+    rowId = row.id,
+    <tr model={rowId}>
+      <td class='col-md-1' textContent={rowId} />
+      <td class='col-md-4'><a onClick={select} textContent={row.label()} /></td>
+      <td class='col-md-1'><a onClick={remove}><span class='glyphicon glyphicon-remove' aria-hidden='true' /></a></td>
+      <td class='col-md-6'/>
+    </tr>
+  )));
+
   return <div class='container'>
     <div class='jumbotron'><div class='row'>
       <div class='col-md-6'><h1>KnockoutJSX-keyed</h1></div>
@@ -30,14 +41,7 @@ export default function({data, selected, run, runLots, add, update, clear, swapR
       </div></div>
     </div></div>
     <table class='table table-hover table-striped test-data'><tbody>{
-      selectRow(selected, data.memoMap(row =>
-        <tr model={row.id}>
-          <td class='col-md-1' textContent={row.id} />
-          <td class='col-md-4'><a onClick={select}>{row.label}</a></td>
-          <td class='col-md-1'><a onClick={remove}><span class='glyphicon glyphicon-remove' aria-hidden='true' /></a></td>
-          <td class='col-md-6'/>
-        </tr>
-      ))
+      items
     }</tbody></table>
     <span class='preloadicon glyphicon glyphicon-remove' aria-hidden="true" />
   </div>
--- a/frameworks/keyed/mobx-jsx/package.json
+++ b/frameworks/keyed/mobx-jsx/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-mobx-jsx",
-  "version": "0.6.3",
+  "version": "0.8.0",
   "main": "dist/main.js",
   "js-framework-benchmark": {
     "frameworkVersionFromPackage": "mobx-jsx"
@@ -17,18 +17,18 @@
     "url": "https://github.com/krausest/js-framework-benchmark.git"
   },
   "dependencies": {
-    "babel-plugin-jsx-dom-expressions": "0.11.2",
+    "babel-plugin-jsx-dom-expressions": "0.14.0",
     "mobx": "5.13.0",
-    "mobx-jsx": "0.6.3"
+    "mobx-jsx": "0.8.0"
   },
   "devDependencies": {
-    "@babel/core": "7.5.5",
-    "@babel/plugin-proposal-decorators": "7.4.4",
+    "@babel/core": "7.6.4",
+    "@babel/plugin-proposal-decorators": "7.6.0",
     "@babel/plugin-proposal-class-properties": "7.5.5",
-    "rollup": "1.17.0",
-    "rollup-plugin-babel": "4.3.2",
+    "rollup": "1.26.0",
+    "rollup-plugin-babel": "4.3.3",
     "rollup-plugin-node-resolve": "5.2.0",
     "rollup-plugin-replace": "^2.2.0",
-    "rollup-plugin-terser": "5.1.1"
+    "rollup-plugin-terser": "5.1.2"
   }
 }
--- a/frameworks/keyed/mobx-jsx/src/main.jsx
+++ b/frameworks/keyed/mobx-jsx/src/main.jsx
@@ -1,4 +1,4 @@
-import { observable, action, autorun, computed } from 'mobx';
+import { observable, action, autorun } from 'mobx';
 import { Component, map, render } from 'mobx-jsx';
 
 function _random (max) { return Math.round(Math.random() * 1000) % max; };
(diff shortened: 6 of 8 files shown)


frameworks/keyed/solid-signals/src/main.jsx
```diff
@@ -1,109 +1,113 @@
...
+  }
+
+  function add() { setData(data().concat(buildData(1000))); }
+
+  function update() {
+    freeze(() => {
+      const d = data();
+      let index = 0;
+      while (index < d.length) {
+        d[index].setLabel(d[index].label() + ' !!!');
```
ts-thomas: Is that the fine-grained strategy to you always talking about?^^

ts-thomas: lol

frameworks/keyed/solid/src/main.jsx
```diff
@@ -1,66 +1,70 @@
...
+  </div>
+
+const App = () => {
+  let rowId;
+  const [ state, setState ] = createState({ data: [], selected: null }),
+    run = () => setState({ data: buildData(1000), selected: null }),
+    runLots = () => setState({ data: buildData(10000), selected: null }),
+    add = () => setState('data', d => [...d, ...buildData(1000)]),
+    update = () => setState('data', { by: 10 }, 'label', l => l + ' !!!'),
+    swapRows = () => setState('data', d => d.length > 998 ? { 1: d[998], 998: d[1] } : d),
```
ts-thomas: Also, that isn't the kind of work you talking about, that's the complete opposite of "when other libraries actually check what has changed" ...

frameworks/keyed/solid-signals/src/main.jsx
ryansolid: Part of it. Fine grained reactive libraries are based on wrapping data access with getters and setters (either functions, object getter/setters, or proxies). So data is trees are broken into reactive atoms. Functions are the classic approach used found in libraries like Knockout, Ember etc and are still the most performant approach. i've written an introduction to the topic: [Finding Fine Grained Reactive Programming](https://levelup.gitconnected.com/finding-fine-grained-reactive-programming-89741994ddee?source=friends_link&sk=31c66a70c1dce7dd5f3f4229423ad127).

Proxies are more common these days being used by MobX and Vue 3 (keep in mind these latter 2 feed into VDOM so while change detection is fine-grained rendering is not). Proxies have taken a while to get browser support so I suspect we will see more in the future. But they definitely come with tradeoffs. Most in lack of explicitness. They hide the fact especially on update that change is happening, that you are dealing with anything other than a POJO. Personally it is why Solid has taken a very different approach to proxies than every other library by choosing not to use simple assignment on purpose. To understand that better I've written: [What Every JavaScript Framework Could Learn from React](https://medium.com/@ryansolid/what-every-javascript-framework-could-learn-from-react-1e2bbd9feb09?source=friends_link&sk=75b3f6f90eecc7d210814baa2d5ab52c)

frameworks/keyed/solid/src/main.jsx
ryansolid: Sort of. This is expressive syntax for the same iterative processes you would be using based on [Netflix's Falcor](https://netflix.github.io/falcor/) and [ImmutableJS](https://immutable-js.github.io/immutable-js/). This still does the for loop to update the data and where applicable runs it through reconciliation. You can see this as a declarative replacement of using Proxy assignment. This way writes require access to the setter in order to update values for control, but under the hood we have mutation.

So is this different than a Virtual DOM library? yes. Could it be used to solely directly proxy DOM changes? yes. In fact a few libraries like [crui](https://github.com/cruijs/crui) are experimenting in this direction to avoid reconciliation where possible. It isn't something I would discount if I could make it work well enough in the general case without increasing the API surface. Obviously the examples I gave were all true, since all the libraries use some sort of list reconciliation, but you are correct that fine grain libraries use events to update data rather than reconciling everything. The only reason the VDOM authors don't cry foul, and some have tried (https://github.com/krausest/js-framework-benchmark/issues/230), is that these are the tests that make the least difference. That and the API surface is still about the data, and not being tied to a single view (even if in this test the data is only tied to a single view).

I was pointing out what I did in this issue, so that you understood how changing swap wouldn't just end the discussion. There is no problem with it being the way it is. I'm kind of glad you didn't change it if you didn't feel it was idiomatic Mikado. Sure I shot down down making a different sort of benchmark here, but my opinion of Mikado is still the same as when I first saw it here https://github.com/krausest/js-framework-benchmark/pull/632#issuecomment-529171129. It isn't the developer experience I prefer even remotely(personal opinion) but I'm not surprised it exists. In fact I anticipated your arrival https://github.com/krausest/js-framework-benchmark/pull/398#issuecomment-392296780 (Read the Edit, if you extend it beyond the specific cheat being discussed). If I seem adversarial it is just that the more combative your tone got the more incentivized I am to correct you. I have no issue with Mikado being how it is.

frameworks/keyed/solid-signals/src/main.jsx
ts-thomas: That is an explicit and absolutely non-generic call ... and please stop posting your articles, that will doesn't change that fact.

frameworks/keyed/solid/src/main.jsx
ts-thomas: OMG

ts-thomas: You may simply forget, that I got 3 shit storms in a row like a witch-hunt ... and you was sledging the hammer also.

ts-thomas: Nobody of yours makes anything better ... if you would like to see a 100% strict implementation, then take a look here: https://github.com/krausest/js-framework-benchmark/pull/656 and learn from it.

frameworks/keyed/solid-signals/src/main.jsx
ryansolid: No you are incorrect. In a sentence: you are pointing at a syntax difference that doesn't dictate either way whether the approach is data driven or not.

But I suppose it's a cop-out a bit to just point you to articles. It's a decent chunk of reading, and I was doing it just to hopefully save myself from explaining concepts again. But I will attempt again, under the assumption you are interested in understanding. I hope by the end of this you will understand the subtle differences which is why your posts gained some heat.

Ok first consider the differences between these:
```js
data = [/*... some data */]

// 1
for(let i = 0; i < data.length; i++) data[i] += 10;

//2
data.forEach((_, i) => data[i] += 10);

//3
function add10(list) { return data.forEach((_, i) => data[i] += 10));
add10(data);
```
The 1st is our typical for loop. The 2nd method is mechanically similar but now it is being called in a function form. The 3rd is slightly different in that the predicate is no longer in the callers control. `add10` might be incredibly useful in your domain and used everywhere but it is less general. Pretty straight forward. Now Lets consider the following:

```js
// 1
function action1() { doUpdate(); }

//2
function action2() { emitter.trigger("update"); }
// somewhere else
emitter.addListener("update", doUpdate);
```
Again these will mechanically do very similar things once wired up. But the subtle difference is the inversion of control. The coupling is looser and a different party could be responsible for wiring it up. A similar conversation comes up when comparing 1 vs 2 way binding. In React you might write(although you would probably use defaultValue):
```jsx
<input type="text" value={state.value} onInput={e => setState({ value: e.target.value })} />
```
This is still a full cycle equivalent to 2 way binding with 2 differences: It is much more verbose to wire up, and the flow is much more explicit. You see(and control) the direction data is travelling. This is known as uni-directional flow.

Finally, consider the difference between Proxies and function calls:
```js
function getValue(key) { return data[key]; }
function setValue(key, value) { data[key] = value; }

const pData = new Proxy(data, {
  get(target, key) { return target[key] },
  set(target, key, value) {
    target[key] = value;
  }
});

setValue("name", "John");
const name1 = get("name");

pData.name = "John;
const name2 = pData.name;
```
As you can see again mechanically accomplishing the same thing. There however is one key difference. The functions have the ability to be managed separately. I can pass around the getter or setter independently. And yes I can do that with the proxy by wrapping it in 2 functions. But my point is like the example above about 2 way binding this is just a syntax choice that has data flow repercussions although mechanically they can work the same.

---------------
Ok tying this all together. What is the difference here between the highlighted code snippet and Mikado's swap? Consider the first 2 examples where we have the different versions of the loop, and where we do direct updates versus registering an event emitter. Using the same language here is how Solid and Mikado differ on say Swap:

```js
// mikado
swapDOM(index1, index2) { 
  // ... do DOM work
}
swapDOM(1, 998);

// Solid
updateIndex(index1, value) {
  emitter[index1].trigger("update", value)
}

const data1 = data[1];
const data2 = data[998];
updateIndex(1, data1);
updateIndex(998, data2);

// somewhere else
data.forEach((_, i) => emitter[i].register("update", () => /* do DOM work */));
```
Even if on a different granularity than say a Virtual DOM library Solid's approach is detached from the updates. The data is independently managed than any DOM updates. You can use the data without ever touching the DOM, or you bind the same data to 10 views. Once wired yes updates can happen rather directly but the approach opens the door for the data to drive any update you want. Where as Mikado's helper method is just a glorified wrapper of direct DOM updates. The end user does not set the predicate so to speak (from our first example). It isn't just about reconciling or diffing approach, but that you control the application through the flow of data.

Now not all libraries need to be Data Driven. But this is what I mean by data driven and it checks with definition of the Virtual DOM libraries. They don't use a number of event emitters but they still update the data with no mindful consequence of what the DOM updates may be. The data update function doesn't know, the consuming template does.

Finally, since Proxies are just a syntax wrapper over functions. I could choose to use a Proxy to wrap either approach. I can have my set handler directly update the DOM or have it trigger registered updates. There is nothing inherent about it being a proxy that makes things different. Nothing makes using Proxies more pure or not. What makes something Data Driven is not its ability to tightly bind to DOM updates. No one would call jQuery Data Driven. Rather that conceptually updates are made to Data rather than the DOM. That the Data and DOM updates are not implicitly tightly coupled. That the data can update the DOM as desired without the person updating the data needing to be conscious of the DOM updates.

frameworks/keyed/solid/src/main.jsx
ryansolid: I know I sound like a broken record at this point. But I will say it again. I have no issue with how Mikado works. Your impression of me slamming down the hammer, is simply because you set yourself up by saying things are just not true or lead me to believe you are missing the point.

Like this statement you just made about your Proxy solution. In this case I haven't had the chance yet to check whether your statement is true, just follow me for a second. Let's pretend I look and based on what I explained in https://github.com/krausest/js-framework-benchmark/pull/655#discussion_r341264466 I realize that in fact this isn't "100% clean implementation(probably the most clean/strict of all)". Because of the absoluteness of your statement that pretend, I know not to be true at all, I respond and drop the hammer so to speak and my response carries weight. You leave feeling dejected and frustrated, and I felt I had no choice but correcting you because the statement was so absolute that I felt it could be misleading. Does it any point mean I dislike what Mikado is doing? No. Do I think that Mikado using Proxies is a super cool feature that could move Web Development forward definitely. I want to see all the potential of Proxies. It would just that I felt the need to clarify. I'm pretty sure there are ways to avoid this in the future. The easiest might simply be conscious of talking in absolutes unless you want to invite longer discussions.

ts-thomas: Ok sorry, it might we are different. I just need some seconds to look on the code to realize what it is. The implementation of my proxy benchmark is so simple, that I couldn't believe that I need to make any explanations to it.

frameworks/keyed/solid-signals/src/main.jsx
ts-thomas: Thanks for explanation, but I know these things already. By the way IoC is something I use very often, when it fits to the given task. Your example how mikado does apply swap is wrong for the proxy version. The truth is it looks very similiar to the approach you posted below. There is just one detail which makes my proxy implementation special, maybe unique, it completely derives from data to provide this: "That the data can update the DOM as desired without the person updating the data needing to be conscious of the DOM updates". If you find something similar, please show me, I'm interested in.

ts-thomas: BTW, bi-directional binding is a complete different topic ...

ryansolid: > BTW, bi-directional binding is a complete different topic ...

Definitely. It is just a good example that we would all be familiar with where something effectively ends up being the same thing but the concept/spirit of the thing is very different. With direct updates or triggered, the very same code could more or less be run the same but there are fundamental differences on how you approach it.

And I mean I haven't checked but I assumed your Proxy implementation was different than the original Mikado implementation in that way. I just wanted to spell out how Solid or the new Proxy implementation is different from that.

> There is just one detail which makes my proxy implementation special, maybe unique, it completely derives from data to provide this: "That the data can update the DOM as desired without the person updating the data needing to be conscious of the DOM updates".

This is what I don't understand. I believe most implementations do this for the most part. Maybe comment on that in the other thread.

ts-thomas:
I ask my self, what would happened when I say that the mikado proxy implementation is the only one which is data-driven __without__ using tricks like `data(items)` or `setState()` or `setData()` to __explicitly invoke the render__. I'm pretty sure that you all began to cry, because it simply doesn't fit into the essence of your libs. But that would be absolutely the same kind of discussion, and also I could claim that this is the "one-and-only-legal-way-on-how-doing-things-right". But I won't. As you realize, there never can't exist a "global" rule of what is good and what is not. And I expect the same democratic thinking from every other developer too, especially from library authors. What I have experienced so far from you special group of community is the complete opposite of what I know about others. And it feels absolutely unprofessional to prefer kicking someone instead of proofing the facts.

frameworks/keyed/solid/src/main.jsx
ryansolid: I mean from what you described above I believe the Proxy implementation is as you say. I just was trying to come up with some hypothetical example of what has been happening to this point.

ryansolid:
> I ask my self, what would happened when I say that the mikado proxy implementation is the only one which is data-driven without using tricks like data(items) or setState() or setData() to explicitly invoke the render

What do you mean? Those aren't tricks. I mean sure VDOM triggers a render/diff routine. But your proxies technically are triggers. They cause the DOM to be updated. To me that is a distinction on syntax. Solid could use Proxies the same way but choose not too for reasons (in the article link and explained in my other posts). All you are describing is fine grained versus course grain updates. Fundamentally it is still the same. End user sets data, triggers update, DOM is updated. Data-driven. The means of the trigger is syntax. In this case it doesn't have the same spirit versus mechanism thing I was describing above since regardless it works the same way. You perform a mutation on the data independent of thinking of how the DOM will update and DOM updates. It isn't any different.

ts-thomas:
Let me summarize:
- I build the fastest template render library, which couldn't be beaten by any of existing libs (regardless of how it is implemented)
- It uses close to plain web standards by the same time, and provides ob of the lowest complexity for developers
- I discovered 3 new strategies from scratch for reconciliation, from where I implemented 2
- Mikados reconcile are the most effective today (the code to solving this is super tiny compared to all anything)
- Mikado is also is providing a reactive paradigm based on proxy (with full fallback) which allows the rare feature (I never seen this before) to apply absolutely all changes via data-driven. Just change the Array, the dom reacts respectively. In my eyes this is unique.
- All that is provided by one single library where you can choose the preferation.
- The best of all it is open source and free.

No one of you never has said anything positive ... that's really crazy by this list^^
Either Mikado is getting the respect to which it is entitled (no one says you should have to like me), or it makes no sense to continue talking.

ts-thomas:
"To me that is a distinction on syntax" is also `view.swap(1, 998)` ... can you finally decide?

ryansolid:
> "To me that is a distinction on syntax" is also `view.swap(1, 998)` ... can you finally decide?

As I've been saying that swap alone told me nothing. I looked and saw what Mikado keyed implementation was doing and was like. "Oh this is just directly wrapping DOM operations. It is tightly coupled like jQuery. Hmm.. I'm sure we've all thought about doing this. Surprised no one has bothered building around it. Cool." Look back at my first impression(https://github.com/krausest/js-framework-benchmark/pull/632#issuecomment-529171129). Clearly your proxy doesn't work the same way. That's also cool.

I don't know what you expect. I mean do you want me to dissect?

> I build the fastest template render library, which couldn't be beaten by any of existing libs (regardless of how it is implemented)

Definitely. However the library author audience isn't going to give accolades here simply because I'm sure they've all thought of doing stuff similar, but it didn't fit with their experience goals etc.. I acknowledge it is the fastest. Before you were the fastest Solid was, before that it was DomC, before that it was Surplus, before that it was ivi, before that it was inferno.. and so on. Libraries have been in that place at various times. Some people might not agree with how you got there but ultimately bottom line is it is undeniable. As I've always said let the end user look at the implementation and decide what experience they prefer.

> It uses close to plain web standards by the same time, and provides ob of the lowest complexity for developers.

Opinion. Many library author believe they have accomplished this. Talk to lit-html, talk to Svelte and so on.

> I discovered 3 new strategies from scratch for reconciliation, from where I implemented 2
Mikados reconcile are the most effective today (the code to solving this is super tiny compared to all anything)

Is it? Cool. It is good when people push what can be done here. So it's all positive. Although it wasn't what you submitted when you submitted your keyed implementation. So no one here even knows about it.

> Mikado is also is providing a reactive paradigm based on proxy (with full fallback) which allows the rare feature (I never seen this before) to apply absolutely all changes via data-driven. Just change the Array, the dom reacts respectively. In my eyes this is unique.

I think this very cool. This is an area I'm super interested in general. I'm going to check it out when I get the chance. Notice no one has said anything negative about this. I mean we've all seen MobX, Vue3, Svelte (compiler rather than proxy but looks similar), not to mention some of us are familiar with Knockout, Solid, Surplus, Sinuous etc.. which work off similar concepts. I think this is still relatively new ground so it is unique in that sense. I personally look forward to it. I was actually going to write something positive on that PR and thumbs up, until I read the post where you yet again came off as combative.

> All that is provided by one single library where you can choose the preferation.
The best of all it is open source and free.

I think that's great.

ts-thomas:
It is hard to believe that Svelte has the same learning curve than Mikado. But yes, developers differs. I forced myself so often to use Angular or React, because I always said to me, they must have advantages because they are such a common standard. But it never feels good to me, anytime, nor it makes any fun. As a full stack developer I'm not a fan of client-weighted ecosystems. Also it destroys the classical MVC, one of the biggest milestones in software engineering. I used mithril in the past, then I build Xone (a MVC on top of Inferno) before I switched to Meteor from which I've got many impressions. The experience shows me that it gives an advantage when the possibility of mixin the MVC exists (like in Ruby on Rails). That's why Mikado would never become a MVC like mithril. It should provide just the "V" within a MVC as good as possible with all freedoms to integrate into every stack. That's why Mikado has functional over declarative paradigm, because the latter is often a closed road. Just provide an adapter and you can adopt Mikado into just every ecosystem you like. But thats of course is nothing new.

ryansolid:
On an aside, have you thought of how to best benchmark the Mikado-Proxy implementation (reuse: false, store: store) submitted here? I'm gathering it would have a similar issue in your benchmark even under data-driven. Just something to think about. Because if you can handle that case the other reactive libraries would also probably be better represented.

ts-thomas:
Absolutely, but it has one downgrade. Actually I show how the same library runs under different situations. Just the keyed test has a limitation. Forces the data-driven test as strict data-driven-only has the drawback that the comparison against all others is not possible. I know these are in fact aren't comparable, but from the developers perspective they all made the same things at the end, rendering an UI to the users screen. I never saw a comparison like this before, because I can compare the benefits and downgrades coming from data-driven against the other paradigm for the same render job. But I will add mikado-proxy to any of these 3 test categories, so reactive becomes comparable against functional. If you like, there exist a test where all different modes are compared: https://raw.githack.com/nextapps-de/mikado/master/bench/#modes ... when I will add the "mikado-observer-proxy" (aka mikado-proxy from my PR) the next problem is the astronomical score just from the "repaint" test. Maybe not a problem, but I need a kind of normalize it ... but that's also has an effect of other tests (like sinuous repaint test).

localvoid:
> And it feels absolutely unprofessional to prefer kicking someone instead of proofing the facts.

Mithril author is probably one of the nicest people in a javascript reddit community and he genuinely tried to help you when it was obvious that you didn't even understand the basics.

> I build the fastest template render library, which couldn't be beaten by any of existing libs

Nobody cares because you aren't even competing with libraries like React/Solid, your library in the same category as stage0.

> It uses close to plain web standards by the same time, and provides ob of the lowest complexity for developers

Every library author here is probably thinking that their library has "the lowest complexity for developers" :)

> Mikados reconcile are the most effective today (the code to solving this is super tiny compared to all anything)

Nested loop, wow, such effective :man_shrugging: 

> Mikado is also is providing a reactive paradigm based on proxy (with full fallback) which allows the rare feature (I never seen this before) to apply absolutely all changes via data-driven. Just change the Array, the dom reacts respectively. In my eyes this is unique.

All reactive libraries that applied DOM mutations synchronously died in early 2010s :man_shrugging:

> I forced myself so often to use Angular or React, because I always said to me, they must have advantages because they are such a common standard.

How is it possible that you [didn't even understood](https://github.com/nextapps-de/mikado/blob/e7aa98bc86b138be6c3c46f9e7b5b8711cfbd357/bench/test/inferno/keyed.html#L47) initially how to create keyed implementations with libraries like React in your benchmark? I seriously can't imagine how it is possible for someone who used React, because React will print a lot of warnings as soon as you start render dynamic lists.

> I build Xone (a MVC on top of Inferno)

Just looked at [Xone sources](https://github.com/nextapps-de/xone), it looks like a library from early 2010s. 

> the next problem is the astronomical score just from the "repaint" test

We've learned all this techniques how to inflate benchmark numbers 5 years ago, we've done all this benchmarketing bullshit a long time ago. Do you seriously think that nobody thought about adding keyed recycling in a benchmark that applies DOM mutations without flushing DOM ops? We've seen this 100x benchmark improvements, nobody cares about this useless "optimizations".

ts-thomas:
@localvoid it looks like you've learned a much the last 5 years ... but what you doesn't learned is to thinking before writing. When I first time looked into ivi ... I thought omg what a crap. But slowly I understand why these lib is so strange ...

ts-thomas:
It looks like you've learned a lot, but sadly by far not enough :(  ... although I wish you good luck with celebrating yourself

localvoid:
> although I wish you good luck with celebrating yourself

https://en.wikipedia.org/wiki/Psychological_projection

ts-thomas:
lol

https://en.wikipedia.org/wiki/Envy

fabioricali:
> > Mikado is also is providing a reactive paradigm based on proxy (with full fallback) which allows the rare feature (I never seen this before) to apply absolutely all changes via data-driven. Just change the Array, the dom reacts respectively. In my eyes this is unique.

My library uses also proxy for reactive changes... is the best solution.
if it could be useful:
https://github.com/dozjs/doz

ts-thomas:
https://github.com/dozjs/doz/blob/master/example/todolist.html
@fabioricali very nice, thanks

localvoid:
@fabioricali if I understand correctly, your library works like Vue and it just provides a different API on top of the proxy API, it is a perfectly fine solution. With mikado, when you remove element from an observable array, [it will immediately trigger DOM mutation](https://github.com/nextapps-de/mikado/blob/9e413f385e107a220fc9c1314a2f2921e85e73a5/src/array.js#L254) like it was done in old reactive libraries, until library authors realized that it is important to have batching, computeds, glitch-free computations etc,etc...

fabioricali:
@localvoid also in my library at each mutation of an array/object is called the "render" method. Doz uses the virtual DOM anyway.

ts-thomas:
wrong, in Mikado you can choose between batching and choose between immediately ... it so stupid that others mean they know my lib better than me^^

localvoid:
> @ts-thomas it so stupid that others mean they know my lib better than me^^

1. https://github.com/nextapps-de/mikado/blob/9e413f385e107a220fc9c1314a2f2921e85e73a5/src/array.js#L242
2. https://github.com/nextapps-de/mikado/blob/9e413f385e107a220fc9c1314a2f2921e85e73a5/src/array.js#L254
3. https://github.com/nextapps-de/mikado/blob/9e413f385e107a220fc9c1314a2f2921e85e73a5/src/mikado.js#L1471
4. https://github.com/nextapps-de/mikado/blob/9e413f385e107a220fc9c1314a2f2921e85e73a5/src/mikado.js#L1504

ts-thomas:
omg, you just pick one paradigm which is supported by this lib ...

ts-thomas:
https://github.com/nextapps-de/mikado#extern
That does fully support batch updates ... but you couldn't find it because you are just focused to find the negatives.

ryansolid:
@ts-thomas To be fair while Mikado supports many different modes, the only thing really relevant here is what is in the implementations that are submitted in this benchmark. So in a sense @localvoid  didn't pick the paradigm, you did in how you chose to represent your library here. But I don't think you should be necessarily too swayed by opinion and do whatever feels right for Mikado.

ts-thomas:
@ryansolid I've followed your suggestion and added "mikado-proxy" to all 3 test categories: https://github.com/nextapps-de/mikado/blob/master/bench/README.md#keyed

krausest:
@ryansolid Thanks for the update. Results are here: https://krausest.github.io/js-framework-benchmark/2019/table_chrome_78.html

PR #656, opened 2019-10-30
ADD mikado-proxy

ts-thomas:
This is a 100% reactive data-driven approach. It uses Mikados __proxy__ and the new feature __observable array__ (Virtual NodeList) under the hood. Also this one might be the smallest and most simple implementation on the table. Technically it completely differs from my other implementation, so it would be really nice to see both paradigm compared :)


krausest:
Thanks - results are here: https://krausest.github.io/js-framework-benchmark/2019/table_chrome_78.html

PR #657, opened 2019-11-03
Miso 1.3

dmjio:
@krausest this should solve the previous issues that plagued the miso-1.1 PR. Including swap issues. For both keyed and non-keyed. Should be ready to go though.


krausest:
Thanks - I changed the version information in package.json to 1.3.0 - I hope that's correct.
Results are here: https://krausest.github.io/js-framework-benchmark/2019/table_chrome_78.html

PR #659, opened 2019-11-05
Initial commit of miso.js keyed implementation.

dmjio:
@krausest I'd like to try out the underlying javascript that https://haskell-miso.org uses as a standalone framework. I think this will be very telling where the higher-level language bottlenecks reside. I might add non-keyed as well, depending on how this goes. Should be ready to go. Swap issues won't be in this version, nor update on every tenth row issues.


dmjio:
@krausest also, if you'd first like me to put this into an `npm` package I can do that first.

krausest:
Wow, nice - [that's](https://krausest.github.io/js-framework-benchmark/current.html) surprising considering how the miso implementation performs.
My recommendation is to drop the non-keyed implementation and just focus on keyed implementations :smiley:

dmjio:
Heh thanks 😅 seems like the Haskell Miso versión could be implemented much more efficiently. If not, GHCJS is just sandbagging us.

PR #661, opened 2019-11-13
Update lighterhtml to v2

WebReflection:
The [lighterhtml](https://github.com/WebReflection/lighterhtml#readme) library reached version 2.

This MR would like to have it updated so that tests would be more aligned with current performance.

Thank you.


krausest:
Thanks - here is the [update](https://krausest.github.io/js-framework-benchmark/current.html).

PR #662, opened 2019-11-14
Update lighterhtml test

WebReflection:
~~Both _hyperHTML_ and _lighterhtml_ have a special `data` behavior [since about ever](https://viperhtml.js.org/hyperhtml/documentation/#essentials-6-2), which would make the selection or removal of an item straight forward, as it is already for [domc](https://github.com/krausest/js-framework-benchmark/blob/master/frameworks/non-keyed/domc/src/app.js#L57-L67) or other cases, where the `item` is passed right away.~~

~~This behavior allows me to also cleanup a bit redundant layout details, so I hope it's OK to ask for this update.~~

~~Only thing that is slightly different, but it's the way I'd solve this "challenge", is that the item is not re-created each time, but I'm not sure that pattern is mandatory.~~

~~If that's the case, I'd be more than happy to update this MR to still retrieve the index via `data.findIndex`, instead of simply using `data.indexOf`.~~

Never mind, I've realized if I need to retrieve the id, like any other test does, there's no reason to store the item as `data`, so this is just a basic cleanup that I hope will be merged at some point.

Best Regards.

### DIFF
--- a/frameworks/keyed/lighterhtml/src/index.js
+++ b/frameworks/keyed/lighterhtml/src/index.js
@@ -27,27 +27,24 @@ const clear = () => {
 };
 const interact = event => {
   event.preventDefault();
-  event.stopPropagation();
-  const {target} = event;
-  const id = +target.closest('tr').id;
-  if (target.getAttribute('data-interaction') === 'delete') {
-    del(id)
-  } else {
-    select(id)
+  const a = event.target.closest('a');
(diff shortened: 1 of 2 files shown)


WebReflection:
Actually, never mind, as I've tested this and the non-keyed doesn't even work, so since I'm sure you've better things to do, I'll close this MR and re-submit one when I've done all my duties as dev.

Apologies for the time taken to read this.

PR #663, opened 2019-11-14
Remove the non-proxy version, make mikado-proxy as default

ts-thomas:
Done.

Also:
- update to latest
- swap helper removed
- direct dom manipulation (classname) in select test replaced by idiomatic style of the lib
- first time build flag "cache" enabled (improves at least the two new replacements)


krausest:
Thanks! Results are [here](https://krausest.github.io/js-framework-benchmark/current.html). 

BTW: I'm always running keyed/vanillajs when updating a framework to make sure that system conditions are comparable - just in case you're wondering why vanillajs numbers change from run to run.

leeoniya:
nice.

PR #669, opened 2019-11-24
neverland: hooks for lighterhtml

WebReflection:
The [neverland](https://github.com/WebReflection/neverland/#readme) project just reached its 3rd major update.

Even if I know it cannot be faster than _lighterhtml_, I'd like to add this library to the list of tests.

Thanks for considering this PR.


krausest:
Thanks. Results have been updated. Here's a screenshot (for the keyed implementation):
Screenshot from 2019-11-27 19-49-34

PR #670, opened 2019-11-26
heresy: custom elements for lighterhtml

WebReflection:
The [heresy](https://github.com/WebReflection/heresy/#readme) project can perform both keyed and non-keyed updates through Custom Elements, via _lighterhtml_ engine, without name clashing and through simplified literal non-dom-coupled definition.

The project is close to reach v1 and I would like to compare it with other projects, even if Custom Elements re not super common in this benchmark.

Thanks for eventually considering this PR.

### DIFF
--- a/frameworks/keyed/heresy/index.html
+++ b/frameworks/keyed/heresy/index.html
@@ -1,0 +1,6 @@
+<!DOCTYPE html>
+<meta charset="utf-8" />
+<title>heresy keyed</title>
+<link href="/css/currentStyle.css" rel="stylesheet" />
+<div id="container"></div>
+<script src='dist/index.js'></script>
--- a/frameworks/keyed/heresy/package.json
+++ b/frameworks/keyed/heresy/package.json
@@ -1,0 +1,37 @@
+{
+  "name": "js-framework-benchmark-heresy",
+  "version": "1.0.0",
+  "description": "heresy demo",
+  "main": "index.js",
+  "js-framework-benchmark": {
+    "frameworkVersionFromPackage": "heresy"
+  },
+  "scripts": {
+    "build-dev": "rollup -c -w",
+    "build-prod": "rollup -c"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/krausest/js-framework-benchmark.git"
+  },
+  "keywords": [
+    "heresy"
+  ],
+  "author": "Mathis Zeiher",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/krausest/js-framework-benchmark/issues"
+  },
+  "homepage": "https://github.com/krausest/js-framework-benchmark#readme",
+  "dependencies": {
+    "heresy": "^0.23.2"
+  },
+  "devDependencies": {
+    "@ungap/degap": "^0.1.4",
+    "rollup": "^1.27.5",
+    "rollup-plugin-includepaths": "^0.2.3",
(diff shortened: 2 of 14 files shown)


ryansolid:
Personally,  think it would be great to have another web component comparison.

WebReflection:
But that would defeat the idea that Web components (custom elements only in heresy case) are the tool to solve Web problems, right?

ryansolid:
Ok yeah, I didn't follow that at all without the edit.

I've been doing performance comparisons myself in this area and the cost of components in general Web or otherwise in different rendering techniques. Admittedly it is never a fair comparison against a simple component-less implementation but it is very interesting to me that different component solutions scale differently, yet we all see the need to modularize our code. The challenge as always with Components is that we split code for many different reasons and there is a blur on understanding of how different libraries use them differently. Like Components in a Virtual DOM library are just another node, and in efficiently written ones have very minimal overhead, the same is not true for say Svelte where they define reactive boundaries. In a Top Down (including Virtual DOM libraries) you often want to break into more Components for finer grain control to better handle change management, where a Reactive library doesn't pay the cost of of grouping together dynamic structural reconciliation with it's generally static surroundings since it uses finer grained resolution. It is only on Component boundaries where unwrapping/wrapping of reactive values occurs that you incur costs. This leaves you more free to organize code as fits your developers cognitive model. To split code as you see fit for code organization. However if that split is basically make everything a Component like React you will see the cost much more rapidly.

Web Components are that second case where you pay the cost on having them, on the boundaries. Which is an interesting thing because it means when talking about Components Virtual DOM libraries and Custom Element aficionados are often talking about different things. Vue might be an exception because it sort of suffers from both shortcomings with none of the benefits, so I guess Evan is fine talking to both sides?

But I should try to limit this to speaking to raw performance as it pertains to benchmarks. There are other frictions. But that is a different discussion. I think it is good to see in practice from a performance standpoint how much overhead is here. The pure vanilla web component implementation has ~ 10% overhead. Does it scale the same when you have a full library behind it like Heresy or Stencil? That's what I want to see, I suspect they don't scale any worse than most libraries that are in the 2nd category (Angular, Svelte,  Solid, Surplus, etc...). Unfortunately we probably won't get that comparison here as those libraries are incentivized to do this benchmark in a single Component. And who could blame them, since it isn't like the raw renderers(domc, mikado, lit-html, lighterhtml, etc..) in this comparison are going to use Components. But atleast if we can compare heresy to say lighterhtml we can get some sort of indicator.

WebReflection:
> Unfortunately we probably won't get that comparison here as those libraries are incentivized to do this benchmark in a single Component

but that's not fun, and yet I'd love to see those libraries you mentioned playing a role in here too, and `domc`, in that regards, uses the same "_single component_" concept already, while _heresy_ uses a more dev friendly (imho) approach: `App`, `Button`, and `Row`, and it's in its core that heresy shines, enabling the ability to define a `<tr>`, 'cause every other Web Components library you mentioned would fail there, as they believe custom elements built-ins have no role in Web development, to which statement _heresy_ answers saying "_there's literally no other way to create a `<tr>` component otherwise, something every non Web Component based libarry is capable of_".

If nothing, _heresy_ could show that development through standards, in a componentized way, is possible, so I think it should compete with other solutions that componetize `Buttron` and `Row` too.

krausest:
Thanks. Results have been updated. Here's a screenshot (for the keyed implementation):
Screenshot from 2019-11-27 19-49-34

PR #673, opened 2019-11-27
domdiff: raw diffing performane comparison

WebReflection:
The [domdiff](https://github.com/WebReflection/domdiff#readme) module would like to be the fastest, general purpose, non virtual DOM based, diffing utility out there.

This benchmark is a perfect example of both how to use _domdiff_, and show its raw performance (including memory usage).

In an attempt to figure out where is the _lighterhtml_ bottleneck, I ended up benchmarking its diffing core, to also realize how big of an impact is using `WeakMap` instead of arrays to relate, keyed or not (indexed), data to nodes.

Thanks for considering adding this too.

As side note, since I'm pushing like crazy in this repo these days, I can assure this is likely the last benchmark I've interest in putting in the table.

The only exception could be for a simplified _domtagger_ based one, but I'm likely keeping that for myself to keep working, behind the scene, into _lighterhtml_ performance improvement so ... thanks for your patience, this last one should be it from me.

Best Regards.

P.S. this utility is in _hyperHTML_, _lighterhtml_, _heresy_, and _neverland_ core, which is why I believe it's important to see "_how the engine perform_" by itself, so that a _features to perf degradation_ comparison becomes implicitly evident too.


krausest:
You're welcome. I somehow messed this merge up. Seems like there was a commit after merging locally (though the PR shows only one commit). The results shown in the other PRs are from the earlier 5e0feed commit.
I'll check later how to clean it up.

krausest:
Current approach: I'll just delete domdiff and merge the PR again.

WebReflection:
Alternatively, I can close this and create a new one, let me know if that works, thanks!

WebReflection:
OK, I should've resolved all conflicts. Basically I think there was a race condition during my last cleanup, but current state should be the right one for this PR, as I've removed unnecessary cruft in the package.json, 'cause I didn't really need any of that.

krausest:
I'm starting to merge - no more commits, please :smiley:

krausest:
I'm starting to merge - no more commits, please :smiley:

PR #674, opened 2019-11-27
lighterhtml: dependencies update

WebReflection:
Thanks to _domdiff_ investigation and benchmarks, I've eventually found out what was making _lighterhtml_ way slower on the "_select rows_" part of the benchmark, and not only.

Turns out, a single `WeakMap` could compromise this benchmark [extra miles](https://jsperf.com/weakmaps-vs-arrays), and it's only via _domdiff_ PR that I could see where was my main bottleneck.

Meanwhile, both _domdiff_ and _domtagger_ has been updated to produce best possible results, so that _lighterhtml_ 2.0.6 should be (hopefully) right beside _lit-html_, as it uses a single WeakMap too.

This update would like to place _lighterhtml_ in a more fair place, so I hope it will get merged at some point.

Thanks again for your patience, and effort, with this project 👍

### DIFF
--- a/frameworks/keyed/lighterhtml/package.json
+++ b/frameworks/keyed/lighterhtml/package.json
@@ -24,11 +24,11 @@
   },
   "homepage": "https://github.com/krausest/js-framework-benchmark#readme",
   "dependencies": {
-    "lighterhtml": "^2.0.2"
+    "lighterhtml": "2.0.7"
   },
   "devDependencies": {
     "@ungap/degap": "^0.1.4",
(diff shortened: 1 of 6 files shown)


WebReflection:
So, just to contextualize why I am pushing again for this update, I know this is my computer and maybe here it'll be different, but this is the current status, and I honestly couldn't be happier for reaching such state.

EKc4PMMWoAE_Bzo

EKc4PMjWoAA85zy

Best Regards.

PR #675, opened 2019-11-27
hyperHTML: benchmark update

WebReflection:
It's been long time since I have checked the _hyperHTML_ status in here, and beside updates related to every other libraries of mine, sharing same core components, _hyperHTML_ was apparently using `performance.now()`, and other measurements that made absolutely no sense whatsoever compared to other frameworks.

To have a more update, and fair, comparison, I've decided to follow up same approach used in every of my libraries, in this case _lighterhtml_, to see where _hyperHTML_ would position these days, as it's the most deployed UI library (production speaking) I have.

I've said I would've stopped pushing libraries, but please bear with me, these are fair updates, that required some maintenance.

Best Regards


krausest:
I'm starting to merge - no more commits, please :smiley:

krausest:
Done. Results have been updated. 
Here's the relevant screenshot:
Screenshot from 2019-11-28 21-12-36

PR #676, opened 2019-11-29
lighterhtml: final test-case update

WebReflection:
This is the last thing I can change to have a better comparison (as in: more fair one).

As every other competing framework is using `item.id` to reference rows, and as each item lives as long as the database has it, I've finally found a way to not create O(n) WeakMap references, but just O(1) related to the data.

This provides the possibility to use the `item.id` in _lighterhtml_ too (and every other framework of mine, PRs will follow) so that instead of benchmarking WeakMap performance, we can now compare raw libraries logic/core/engines.

As result, and since there's literally nothing else I can do to make it faster so this is really my last update 'til 2020, the new keyed test drops the infamous _selectRow_ slowdown, placing _lighterhtml_ way up in the table.

EKh5_EuWsAAkd5y

Once again, thanks a lot for this benchmark, as it gave me the opportunity to learn all possible bottlenecks to boost up real world applications (and as there's nothing else I can do, I won't torture you anymore with this library).

Best Regards.

### DIFF
--- a/frameworks/keyed/lighterhtml/package.json
+++ b/frameworks/keyed/lighterhtml/package.json
@@ -24,11 +24,12 @@
   },
   "homepage": "https://github.com/krausest/js-framework-benchmark#readme",
   "dependencies": {
-    "lighterhtml": "2.0.7"
+    "lighterhtml": "2.0.7",
+    "js-framework-benchmark-utils": "0.2.5"
   },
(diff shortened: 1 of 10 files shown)


WebReflection:
Screenshot from 2019-11-29 14-51-14

- - -

Screenshot from 2019-11-29 14-51-48

PR #679, opened 2019-11-29
neverland: final test-case + version update

WebReflection:
For all the exact same reasons specified in #676, this is the ultimate change to drop WeakMap slowdown from the equation.

This change also brings in the very latest version of the library, which includes all core updates.


WebReflection:
There is something super weird with latest update, I might have broken everything ... keeping WIP until I am sure it works.

WebReflection:
Heads up: `terser` might break come code, I had to switch to uglify-es here. Now it builds and shows correctly.

WebReflection:
FYI the terser path 4.4.1 broke neverland outcome over the weekend.

Full report here: https://github.com/terser/terser/issues/527

The 4.4.0 works as expected, but whoever is using `rollup-plugin-terse` might have issues too as they use `"terser": "^4.1.0"` [in their package.json](https://github.com/TrySound/rollup-plugin-terser/blob/master/package.json#L38).

PR #680, opened 2019-11-29
heresy: final test-case + version update

WebReflection:
For all the exact same reasons specified in #676, this is the ultimate change to drop WeakMap slowdown from the equation.

This change also brings in the very latest version of the library, which includes all core updates.

### DIFF
--- a/frameworks/keyed/heresy/package.json
+++ b/frameworks/keyed/heresy/package.json
@@ -24,11 +24,12 @@
   },
   "homepage": "https://github.com/krausest/js-framework-benchmark#readme",
   "dependencies": {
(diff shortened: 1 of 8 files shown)


WebReflection:
There is something super weird with latest update, I might have broken everything ... keeping WIP until I am sure it works.

WebReflection:
Somehow heresy didn't seem to be affected by terser, so it builds and shows correctly.

krausest:
[Results](https://krausest.github.io/js-framework-benchmark/current.html) have been updated. Here's a screenshot of the last PRs:
Screenshot from 2019-12-04 21-25-35

PR #686, opened 2020-01-03
Use Riot.js shouldUpdate to avoid unecessary updates

GianlucaGuarini:
I have seen that many other frameworks use the `shouldComponentUpdate`  to improve the rendering performance I thought it wasn't allowed so now I have implemented it in Riot.js as well.

### DIFF
--- a/frameworks/keyed/riot/package.json
+++ b/frameworks/keyed/riot/package.json
@@ -10,12 +10,12 @@
     "build-prod": "webpack -p"
   },
   "devDependencies": {
-    "@riotjs/compiler": "4.5.4",
+    "@riotjs/compiler": "4.5.5",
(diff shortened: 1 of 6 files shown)


krausest:
Hi @GianlucaGuarini . Using shouldComponentUpdate is indeed allowed.
I took a look at it but I'm afraid to say that this seems slower. I've run it twice (as always with keyed/vanillajs to make sure that results are comparable):
riot
Should I merge it anyways or do you want to take a look at it before? I'd expect select rows to be much faster with shouldComponentUpdate.

GianlucaGuarini:
@krausest many thanks for the screenshot I will close this PR because I wanted to improve the performance of the framework. BTW these tests allow me to improve massively the Riot.js performance thank you!

PR #689, opened 2020-01-28
Ember 3.17 - Octane

lifeart:
* updated ember version to Octane  https://blog.emberjs.com/2019/12/20/octane-is-here.html
* updated glimmer-vm version
* code cleanup
* embroider build pipeline


krausest:
Thanks - I‘ll take a look at it this weekend!

krausest:
Looks good: 
(Results for 3.11 from last run, Results for 3.11 when repeating the run today and results for 3.17)

ryansolid:
@lifeart If I'm reading this right other than improving the row reconcilliation method Ember Octane (3.17) is actually slower across the board? Is this expected?

lifeart:
@ryansolid yes, it's pretty ok, because this pr use `modifiers` (functions over DOM nodes) to bind event listeners. Each row has 2 modifiers and it can relatively reduce overall perfromance, I don't want to cheat on it and add only 2 listeners into the table itself to skip event binding in cycle.

PR #691, opened 2020-02-05
Add vue-next and update Svelte to latest version

filipsobol:
Adds benchmark for vue-next (alpha version of Vue.js 3.0) and updates Svelte to latest available release.


krausest:
Thanks. At the moment I'm rebuilding and trying to run on chrome 80. I promise to merge soon after this is done such that we can see the effect of updating svelte. I'm looking forward to seeing vue-next's performance!

krausest:
Thanks - results are [updated](https://krausest.github.io/js-framework-benchmark/current.html). Svelte seems pretty much unchanged, vue-next looks like a nice improvement over vue.

exodusanto:
May be intresting to test also this [suggestion](https://gist.github.com/yyx990803/6e9ee3dff878ad4202ce71ac6b9e51cf) from Evan

filipsobol:
@exodusanto I tried Evan's approach before making this PR, but it didn't make any difference in this tests.

ryansolid:
I'm super impressed that Vue was able to improve even 10% across the board. For a large framework like that it takes considerable effort. I've seen libraries claim huge performance gains, and then make no noticeable difference here. From my perspective they've really delivered here. The difference between Svelte 2 and 3 was a kb or so of size but nothing noticeable on the performance side by comparison.

This benchmarks tests really core level rendering and list management. It is brutally efficient and is unforgiving on any library overhead. It also doesn't give much opportunity for library overhead. DOM is where the majority of the cost is. Assuming they haven't really dug into the DOM operations they do and focused on the scoping the VDOM updates with the reactive system this is quite impressive.

No matter how efficient they make their reactive system their VDOM is redoing some of the work.  However, to see these sort of gains suggests they've managed to cut significant library time costs. I wouldn't be surprised if they increased reactivity system interopt performance by like 50% to get numbers like this. It is really awesome to see a big library spend this sort of effort to reduce size and improve performance like this.

leeoniya:
> Vue was able to improve even 10% across the board.

i think this is extremely non-obvious to most people who look at this benchmark, which is rather unfortunate.

i've suggested multiple times that we should somehow deduct the baseline DOM cost [of the fastest implementation] from all remaining libs and *then* scale them against each other. this would really show the true efficiency differences between them - the spread would be WAY larger than what is shown today. it would be interesting at least to add such an option to the web UI.

ryansolid:
Sort of case and point about performance claims. https://github.com/krausest/js-framework-benchmark/pull/689#issuecomment-589991849. As I said Vue has more than delivered in my eyes. Evan and co. are definitely doing some good work.

PR #692, opened 2020-02-06
Miso 1.4

dmjio:


krausest:
Thanks - results are [updated](https://krausest.github.io/js-framework-benchmark/current.html).

Issue #694, opened 2020-02-14
Strengthen isKeyed test for swap rows


krausest:
As discovered in #693 currently the isKeyed test for swap rows is too weak to detect implementations that remove the two rows and simply insert two new rows.
I really think it makes much more sense to require a keyed implementation to actually move the rows, i.e. add exactly the rows that were removed (otherwise any dom state would be lost when swapping rows).

If isKeyed checks whether the rows are actually moved for swap rows the following implementations fail:

- [ ] aurelia-v1.3.0-keyed
- [ ] binding.scala-v10.0.1-keyed
- [ ] crui-v0.1.0-alpha.13-keyed
- [ ] datum-v0.12.2-keyed
- [ ] dojo-v6.0.4-keyed
- [ ] fidan-v0.0.23-keyed
- [ ] maquette-v3.3.0-keyed
- [ ] mikado-v0.7.57-keyed
- [x] ractive-v1.3.6-keyed  
- [ ] react-hookstate-v16.8.6 + 1.5.2-keyed
- [ ] reflex-dom-v0.4-keyed
- [x] san-v3.7.7-keyed

Can you please take a look at your implementations? Is it possible to fix it for your frameworks?
@Alexander-Taran @Atry @Iazel @MartinRixham @agubler @ismail-codar @johan-gorter  @ts-thomas @alexfmpe @errorrik

Atry:
I think it’s better to create a third category other than keyed and unkeyed
instead of changing the definition

Stefan Krause <notifications@github.com>于2020年2月14日 周五下午12:20写道：

leeoniya:
the name of the metric has always been "swap", unambiguously. this simply improves the assertion of the metric.

krausest:
@Atry I don't think we're changing the [definition](https://www.stefankrause.net/wp/?p=342) of keyed, but indeed we're changing the test for keyed that wasn't sufficient (and I'm not sure it's now watertight).

johan-gorter:
Maquette does not support swapping nodes, you can exclude maquette from this test for this reason.

krausest:
I decided to update the results table such that it show any open issues for the implementations. The information is kept in package.json (js-framework-benchmark.issues).

ts-thomas:
Hello @krausest, the mikado keyed implementation was tagged with the issue "Keyed implementations must move the DOM nodes for swap rows", but that's not right. The implementation must assign the new values to the data array (like all others).

```js
.route("swaprows", () => {
[...]

PR #695, opened 2020-02-14
update ractive keyed row swap per #694

evs-chris:
Prior to 0.10, ractive didn't really support moving more than one item at a time, which means that the row swap only really worked for one item and the other was recreated. This is now possible, though not exactly an optimized case.

See #694


krausest:
Thanks a lot! Results have been [updated](https://krausest.github.io/js-framework-benchmark/current.html).

PR #696, opened 2020-02-17
Added Seed 0.6.0

MartinKavik:
Added Rust framework Seed - https://seed-rs.org/

---

`npm run check non-keyed/seed` result:

```
Non-keyed test for remove failed. Expected that the dom node for the 2nd row would NOT be r
emoved, but it was.
seed-v0.6.0-non-keyed is non-keyed for 'run benchmark' and keyed for 'remove row benchmark'
 and non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
```
- ^ I don't see `ERROR` so I assume it's ok. 
- `npm run bench non-keyed/seed` passes without errors.
- `bundled-dist` content included because of Rust dependencies - `package.json` is inspired by the Yew's one.

Thank you!


krausest:
Thanks! Results have been [updated](https://krausest.github.io/js-framework-benchmark/current.html).
Requirements for qualifying as **non**-keyed aren't high :smile: .  Any working implementation should be fine that creates the correct html.

MartinKavik:
Thanks. Keys are WIP so I hope I'll add keyed version soon :)

PR #700, opened 2020-02-25
Update ScarletsFrame

StefansArya:
I have updated the implementation of ScarletsFrame and remove the sleep/delay from the benchmark runner.


krausest:
Nice [improvement](https://krausest.github.io/js-framework-benchmark/current.html)!

PR #704, opened 2020-03-01
incr_dom 0.13

michaelboulton:
I was trying to figure out how https://github.com/janestreet/incr_dom worked so I did this benchmark using it. I also did a wrapper for the incr_dom vdom library so that the react JSX syntax worked. A lot of it is just copied from my initial reason-react implementation.

I think this library is intended for much more complicated systems (part of why the javascript bundle is so huge), and it does actually have special functionality for not displaying all data in a big table (https://github.com/janestreet/incr_dom/blob/master/example/ts_gui/README.md) but it gets very complicated very fast so I kind of gave up on using that - it also seemed to confuse the webdriver tests that tried to select items which were now not being rendered.

This can take a really long time to build because it needs to download and build ocaml and then all the dependent libraries.

### keyed vs non-keyed

I'm not explicitly using any keys and I don't think the library does, so I would expect this would go under non-keyed.

However, if I run `npm run isKeyed --framework non-keyed/incr_dom` I get 
```
ERROR: Framework incr_dom-v0.13.0-non-keyed html is not correct
Non-keyed test for remove failed. Expected that the dom node for the 2nd row would NOT be removed, but it was.
```
and if I move it to the keyed/ folder and run ` npm run isKeyed --framework keyed/incr_dom` I get
```
ERROR: Framework incr_dom-v0.13.0-keyed html is not correct
```
I don't really know the best folder to put it in.


krausest:
I took a look at it. The issue isn't keyed vs. non-keyed but the html structure (though the log makes it a bit hard to read). The problem is:
```
Expected to find 'aria-hidden'=true on span in third td, but found  1
ERROR: Framework incr_dom-v0.13.0-non-keyed html is not correct
```
`<td class="col-md-1"><a><span class="glyphicon glyphicon-remove"></span></a></td>`
should be 
`<td class="col-md-1"><a href="#"><span aria-hidden="true" class="glyphicon glyphicon-remove"></span></a></td>`
Can you take a look at that please?

The isKeyed check will continue to complain that remove behaves as keyed, but non-keyed frameworks are allowed to do that.

michaelboulton:
Trying to figure it out but it seems like there's some issue with setting custom properties using virtual_dom that I'm not seeing, will try to figure it out tomorrow.

michaelboulton:
There seemed to be some issue with the custom properties, I've changed the way it calculates them and it should work correctly now.

krausest:
Thanks - results have been [updated](https://krausest.github.io/js-framework-benchmark/current.html).
I'll keep an eye on build times and further development. Maybe I'll check in the resulting js files and make the build optional.

PR #709, opened 2020-03-16
Fix #708 - Updated lighterhtml, neverland, heresy, uhtml & hyperhtml

### LINKED ISSUE #708: Neverland build is broken
Usually this is due to a version that isn't locked, but not in this case.
Both keyed and non-keyed version are broken when rebuilt.

@WebReflection Can you please take a look at it? I tried to downgrade  some packages that differed between the old package-lock.json and the current but didn't find the guilty one.

WebReflection:
There have been quite some consistent refactoring of my libraries so that I've thought it might make sense to update all their score/benchmarks at once.

### DIFF
--- a/frameworks/keyed/domdiff/package.json
+++ b/frameworks/keyed/domdiff/package.json
@@ -24,12 +24,12 @@
   },
   "homepage": "https://github.com/krausest/js-framework-benchmark#readme",
   "dependencies": {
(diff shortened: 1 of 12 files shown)


krausest:
Thanks - results are [updated](https://krausest.github.io/js-framework-benchmark/current.html).

PR #710, opened 2020-03-25
Update Solid, ko-jsx, and mobx-jsx

ryansolid:
Big rewrites to reduce size for Solid. Too many changes to list. Most code changes were the core reactive library so performance could have changed.

Renamed solid -> solid-state, solid-signals -> solid to reduce confusion.


krausest:
Thanks - all benchmarks have been [updated](https://krausest.github.io/js-framework-benchmark/current.html) to Ubuntu 20.04 beta.

ryansolid:
Thanks @krausest. Hard to really tell if anything is really any different. So close at the top. But I will keep chipping away.

PR #714, opened 2020-04-03
Added ResonateJS

TerraTkai:
I have added ResonateJS under the keyed frameworks, hopefully it is setup as required, but any problems and I will rectify them. ResonateJS is a new framework I created for a project I am working on but I thought I would share it, why not? The website is ResonateJS.org if you need any details. Thanks for the hard work on this benchmark it was a real time saver and it was good to see I am in the right ballpark for performance.


krausest:
Thanks, looks good and performance looks fine too! 
All benchmarks have been [updated](https://krausest.github.io/js-framework-benchmark/current.html) to Ubuntu 20.04 beta.

TerraTkai:
Thanks for merging it in. I'm happy as a first version I'm sure there is room for improvements. I don't suppose anyone but me will ever use it but its good to know my projects are not going to suffer for it. Thanks again for maintaining this!

PR #715, opened 2020-04-03
Added benchmarks for lit-element

iFaxity:
Benchmarks for lit-element was missing from the frameworks, only lit-html existed.
Would be great to add this framework too.


krausest:
Thanks - all benchmarks have been [updated](https://krausest.github.io/js-framework-benchmark/current.html) to Ubuntu 20.04 beta.

PR #716, opened 2020-04-05
Update Halogen benchmarks for Halogen 5 and ecosystem changes

thomashoneyman:
This update to the Halogen benchmark makes a few tweaks:

1. Halogen is now on version 5, which tweaks how components like this one are written. I've updated the component to be compatible with Halogen 5.
2. Random numbers in PureScript are implemented via the FFI using JavaScript's random functions. For that reason, PureScript functions that call `random` repeatedly are, in production code, also implemented via the FFI. That's because it isn't necessarily stack-safe to call an effectful function like this one say 10,000 times. The previous benchmark used a type class, `MonadRec`, to make this repeated call safe, but an ordinary application would just use the FFI. For that reason I've updated the random number generation.
3. The PureScript ecosystem has moved entirely off of Bower in favor of package sets managed via the Spago dependency manager & build tool. This PR switches to package sets.

This isn't meant to be a huge performance change for the benchmark, but it does move from the current benchmark's ~2.1x slowdown to around a ~1.8x slowdown.

Screen Shot 2020-04-04 at 8 37 14 PM

I am not sure why there is such a big outlier wrt `select row` -- running on my machine I see the same thing in the existing Halogen 4 benchmark, but Halogen 4 has closer to a 2x slowdown in the official benchmark for this particular task.

I'm not sure what might be causing this; perhaps there is something odd with my system as I'm clearly out of line with the official benchmark for the existing Halogen 4 implementation.

Anyway -- I think this ought to be merged simply as a keeping-up-to-date commit for Halogen.


krausest:
Thanks, all benchmarks have been [updated](https://krausest.github.io/js-framework-benchmark/current.html) to Ubuntu 20.04 beta.

thomashoneyman:
Thanks!

PR #718, opened 2020-04-05
Upgrade AppRun to v2.23.10

yysun:


krausest:
Thanks, I merged it. Due to the arrival of chrome 81 I have to run all implementations before publishing any numbers.

krausest:
Results are here: https://krausest.github.io/js-framework-benchmark/2020/table_chrome_81.html

PR #719, opened 2020-04-11
Riot.js improved event handling

GianlucaGuarini:
I wasn't happy about this PR https://github.com/krausest/js-framework-benchmark/pull/707, I felt like cheating (even if other frameworks use this cheap cheat to improve their performance)

I was able to solve the problem to  the root improving the way Riot.js handles its events internally.

### DIFF
--- a/frameworks/keyed/riot/package.json
+++ b/frameworks/keyed/riot/package.json
@@ -10,12 +10,12 @@
     "build-prod": "webpack -p"
   },
   "devDependencies": {
-    "@riotjs/compiler": "4.7.0",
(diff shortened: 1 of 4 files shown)


krausest:
Thanks, I merged it. Due to the arrival of chrome 81 I have to run all implementations before publishing any numbers.

krausest:
Thanks a lot. It's really a hard to judge how much cheating explicit delegation really is - so thank you for that clean version!
Sadly chrome 81 mixes results a bit up (and performs generally worse than chrome 80).
Results are here: https://krausest.github.io/js-framework-benchmark/2020/table_chrome_81.html

PR #721, opened 2020-04-13
Add Isotope UI library

areknawo:
Hi there!
First off, great thanks for this benchmark - I've been using it as a reference and performance comparison resource for a long time now!
With this PR, I'd like to add my new open-source UI library called **Isotope** to the comparison. If you're interested, the repo for the library is [here](https://github.com/Isotope-js/core), and the docs [here](https://areknawo.com/isotope).

### DIFF
--- a/frameworks/keyed/isotope/index.html
+++ b/frameworks/keyed/isotope/index.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Isotope</title>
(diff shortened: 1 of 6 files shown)


krausest:
Thanks, I've merged it. Due to the arrival of chrome 81 I have to run all implementations before publishing any numbers.

krausest:
Results are published now: https://krausest.github.io/js-framework-benchmark/2020/table_chrome_81.html
Really not bad!

PR #724, opened 2020-04-20
upgrade AppRun examples to use TypeScript strict

yysun:


krausest:
Sorry, the keyed version appears to be broken.It doesn't recreate the table when updating all rows
```
npm run isKeyed -- --headless keyed/apprun non-keyed/apprun
```

PR #727, opened 2020-04-28
add helix framework

lucywang000:
[Helix](https://github.com/Lokeh/helix) is a react wrapper library for ClojureScript which tries to provide as much of original react to cljs as possible.

The demo code is ported from frameworks/keyed/reagent


lucywang000:
@krausest would you please take a look?

krausest:
Thanks. I changed to things: Select row was't working correctly and I removed the deprecated start/stop measurement code.

PR #729, opened 2020-05-03
Simplify hyperapp implementation

jorgebucaran:
- Remove useless non-keyed Hyperapp
- User hyperscript
- Update developer deps
- Remove rollup config file
- Refactor imperative code to store


krausest:
Thanks, especially for removing a non-keyed version!
[Results](https://krausest.github.io/js-framework-benchmark/current.html) are almost exactly like in the old implementation.

jorgebucaran:
@krausest Thank you for merging! I was still on the fence about a few cosmetic things, like using hyperscript or JSX, but why not! 🙌

jorgebucaran:
@krausest I've noticed Hyperapp "underperforms" in the script bootup time test. Do you have any general pointers for improving in this test? Could this be a compilation/minification setting I'm missing? 🤔

PR #731, opened 2020-05-14
Add Crank.js

brainkim:
Please don’t merge or run the benchmarks yet!

This is a work in progress and it only doesn’t run like ass on a development branch of Crank.js. I need to deploy a release with performance fixes!

Thanks for creating this project. It was super helpful for finding performance problems and watching the benchmarks run was equal parts nerve-racking and exciting. I’ll have more thoughts on what I think about the benchmarks soon!

### DIFF
--- a/frameworks/keyed/crank/index.html
+++ b/frameworks/keyed/crank/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <title>Crank</title>
+  <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+  <div id='main'></div>
+  <script src='dist/main.js'></script>
+</body>
+</html>
--- a/frameworks/keyed/crank/package.json
+++ b/frameworks/keyed/crank/package.json
@@ -1,0 +1,35 @@
+{
+  "name": "js-framework-benchmark-crank",
+  "version": "0.1.0",
+  "description": "Benchmark for Crank.js",
+  "main": "index.js",
+  "js-framework-benchmark": {
+    "frameworkVersionFromPackage": "@bikeshaving/crank"
+  },
+  "scripts": {
+    "build-dev": "CRANK_ENV=development webpack",
+    "build-prod": "webpack"
+  },
+  "keywords": [
+    "crank",
+    "webpack"
+  ],
+  "author": "Stefan Krause",
+  "license": "Apache-2.0",
+  "homepage": "https://github.com/krausest/js-framework-benchmark",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/krausest/js-framework-benchmark.git"
+  },
+  "devDependencies": {
+    "@babel/core": "7.4.5",
+    "babel-loader": "8.0.6",
+    "babel-preset-crank": "^0.1.0",
+    "terser-webpack-plugin": "1.3.0",
+    "webpack": "4.34.0",
+    "webpack-cli": "3.3.4"
+  },
+  "dependencies": {
+    "@bikeshaving/crank": "0.2.1"
+  }
+}
--- a/frameworks/keyed/crank/src/main.jsx
+++ b/frameworks/keyed/crank/src/main.jsx
@@ -1,0 +1,204 @@
+import {Copy, createElement} from "@bikeshaving/crank";
+import {renderer} from "@bikeshaving/crank/dom";
+
+function random(max) { return Math.round(Math.random() * 1000) % max; }
+
+const A = ["pretty", "large", "big", "small", "tall", "short", "long", "handsome", "plain", "quaint", "clean",
+  "elegant", "easy", "angry", "crazy", "helpful", "mushy", "odd", "unsightly", "adorable", "important", "inexpensive",
+  "cheap", "expensive", "fancy"];
+const C = ["red", "yellow", "blue", "green", "pink", "brown", "purple", "brown", "white", "black", "orange"];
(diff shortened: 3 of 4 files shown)


brainkim:
This PR is ready for review. Sorry for taking so long: I made a lot of wild refactorings and kept breaking things. On my MacBook Air, Crank is neck and neck with Preact, which is where I said I’d stop and cut a release. But I’m definitely interested in making Crank even faster, and would love any advice or code review from the js-framework-benchmark community. The core of the library is now a single file and I’ve tried to comment it for readability! (https://github.com/bikeshaving/crank/blob/master/src/index.ts?ts=2)

One implementation note that might be interesting to other framework authors: I was always interested in making Crank generic for all rendering targets, because I want to be able to use it to render Three.js/Pixi.js scene graphs for games/interactive stuff. The first pass at the Renderer API used generators to encapsulate host elements, with the idea that  each host node steps through its associated generator whenever it needs to rerender. This turned out to be a bad idea because retaining a generator object for every single DOM node turned out to require a lot of excess memory.

However, one nice thing about the first approach is that I may have stumbled upon a simpler algorithm for rendering, especially for keyed elements. Basically, what Crank does is rendering in two steps. The first step is the actual element diffing, where virtual elements between renders are compared by tag, position and key. This happens without any actual DOM manipulation. The second step is where we pass both the parent and children at each level of the DOM tree to a [method called “arrange.”](https://github.com/bikeshaving/crank/blob/master/src/dom.ts?ts=2#L149-L221) This method compares the DOM nodes of a parent with an array of children, and inserts/appends the nodes as needed.

I was worried that this two-step approach would be slow, but it doesn’t seem to be that much slower than doing the diffing and arranging in a single pass. The nice thing is that this means that custom Crank renderers don’t have to define methods like insertBefore or appendChild as the custom React/Vue renderers do, which don’t make sense for rendering targets where children aren’t ordered, like three.js scene graphs. And it also makes the key matching algorithm a lot simpler as well; I just compute a map of elements to key on the fly when a key mismatch is detected between renders. I have some ideas to make arranging even faster, like skipping arranges if none of the children have changed or moved, and passing the indexes of the first and last dirty child into the arrange method, but I’m deferring this stuff for a later release.

One thing I’m worried about is that maybe the benchmark isn‘t actually robust enough to tell me if this approach is actually flawed. The swaprows benchmark seems arbitrary; it does a UI change which would likely never occur naturally due to user action or data updates, and I’m worried that [Goodhart’s law](https://en.wikipedia.org/wiki/Goodhart%27s_law) might be coming into play and a lot of people are overfitting their frameworks for that test. Ideally I think three more benchmarks should be added to test more natural DOM mutations: 1. prepending rows to the table. 2. reversing the order of the table 3. sorting the table alphabetically. These benchmarks would be much more robust and informative.

Again, thanks for creating this project. It’s been very helpful!

thysultan:
>One thing I’m worried about is that maybe the benchmark isn‘t actually robust enough to tell me if this approach is actually flawed. 

@localvoid Has a benchmark that might help discover such issues: https://localvoid.github.io/uibench

brainkim:
@thysultan Might be a little too burned out on perf stuff to subject myself to another benchmark right now, but thanks for the tip!

ryansolid:
I do second giving uibench a shot if you are worried about consistency issues and a wider number of reconciliation tests. This benchmark is great for raw speed on most common tasks but if you are concerned that your approach is general enough UIBench is great. 

Swap row is probably the most unlikely actual action and it's currently the only actual reconciling one that involves moving rows. It's more likely to drag a single row in position and shuffle everything in the middle between source and destination places but I suppose most naive approaches would do well there. Reverse also probably is handled well by the short cuts libraries use as it is a bunch of swaps if you go outside in as these libraries do. Sorting has a large degree of randomness which unless all libraries were submitted the same list would introduce way too much variation.

@brainkim You might be interested to know Solid does a 2 pass approach as well to list since I wanted to pull the reactive part out of the renderer so end users could write their own. As you noticed compared to the DOM operations it hardly makes a dent. In my case it actually made things faster since I removed places I was touching the DOM entirely since I resolved most fragment/positioning concerns upfront.

brainkim:
> Reverse also probably is handled well by the short cuts libraries use as it is a bunch of swaps if you go outside in as these libraries do.

Tests which do well with swaprows probably do well with reverse too. The main test which I think would reveal some overfitting is prepending rows. Some of the algorithms I’ve read about explicitly state that they diff the children from last to first which means they’re assuming append is the more common case, and any framework which naively overfits the swaprows case (like Crank) will likely add and remove all original rows from the `tbody` while prepending rows.

> Sorting has a large degree of randomness which unless all libraries were submitted the same list would introduce way too much variation.

I agree and I think the benchmark could definitely benefit from some pseudorandom utility which all frameworks share.

> You might be interested to know Solid does a 2 pass approach as well to list since I wanted to pull the reactive part out of the renderer so end users could write their own.

Wasn’t aware of that. As you can probably guess, I’m a big fan of frameworks being reactivity-agnostic, so this is a pleasant surprise.

If Solid is doing the 2-pass approach I guess I can stick with it haha. The next step for Crank is probably to figure out why the selectrow test is doing so poorly, and implement some more optimizations when there are no shared children between renders, or when there are only deletions. I think it’s interesting that Crank is so skewed between tests which run cold vs tests which run hot, and I wonder if Chrome is just bad at stepping through generators cold or something like that. Probably it’s just some obvious bottleneck which I’m missing.

If anyone has a rig of js-framework-benchmark which works with deoptigate or similar tools lemme know! I’m too lazy right now to try and set it all up myself at the moment.

PR #732, opened 2020-05-17
Add VueRX JSX

ryansolid:
@yyx990803 posted that Vue 3's Reactive library was stand-alone so I couldn't help myself. Let's see how it does when not tethered to Vue 3's hybrid VDOM solution. Introducing [VueRX JSX](https://github.com/ryansolid/vuerx-jsx) (Vue JSX was already taken). It uses the DOM Expressions library to compile JSX down to optimized DOM statements and reactive bindings.

Quick tests show this Vue variation outperforms MobX, but does it have what it takes to go up against the VDOM heavy-weights like Inferno or ivi? We shall see.

### DIFF
--- a/frameworks/keyed/vuerx-jsx/index.html
+++ b/frameworks/keyed/vuerx-jsx/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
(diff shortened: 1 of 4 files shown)


krausest:
Thanks - impressive [results](https://krausest.github.io/js-framework-benchmark/current.html)!

PR #733, opened 2020-05-19
Optimize Reagent select row

Deraen:
This changes the implementation to only render the selected (and deselected) rows, instead of the whole list. Code is still idiomatic and matches what Reagent apps would normally do (though often using Re-frame which adds more state management tools over Reagent built-in features).

Also cleaned the code a bit, by removing performance measure calls. I guess they were there just to help development, other frameworks don't seem to have such. (89b546fe1b772e0f9f0b94da81b83513b6759e4b seems to have removed these calls for JS frameworks)

### DIFF
--- a/frameworks/keyed/reagent/src/demo/main.cljs
+++ b/frameworks/keyed/reagent/src/demo/main.cljs
@@ -1,5 +1,6 @@
 (ns demo.main
   (:require [reagent.core :as r]
+            [reagent.dom :as rdom]
             [demo.utils :as u]))
 
 (defn is-selected? [selected id]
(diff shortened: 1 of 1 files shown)


krausest:
Thanks - not sure what to make of the [results](https://krausest.github.io/js-framework-benchmark/current.html).
Select row is now much faster, but did you expect create rows and remove row to change?

krausest:
Thanks - nice [improvement](https://krausest.github.io/js-framework-benchmark/current.html)

Deraen:
Results probably make sense - there is a bit more more to construct and destruct due to the "track" value. Maybe I even tried this before and selected the another implementation because of this, but this is fine for now.

PR #734, opened 2020-05-19
update helix to be more in line with react-hooks implementation

Lokeh:
This updates the helix framework to be more similar to the idiomatic React Hooks example.

It uses `use-reducer` and memoizes the row and jumbotron components.

It also does a trivial optimization of constructing and then immediately realizing a lazy sequence in `build-data`.


frameworks/keyed/helix/src/demo/state.cljs
```diff
@@ -0,0 +1,67 @@
...
+            (let [row (get data index)]
+              (assoc data index (assoc row :label (str (:label row) " !!!")))))
+          data
+          (range 0 (count data) 10)))
+
+(defn swap-rows [data]
+  (if (> (count data) 998)
+    (-> data
+        (assoc 1 (get data 998))
+        (assoc 998 (get data 1)))
```
lucywang000, line 44: `data[998]` would not change after this, right?

frameworks/keyed/helix/src/demo/state.cljs
```diff
@@ -0,0 +1,67 @@
...
+      (vec data))))
+
+(defn add [data]
+  (let [last-id (:id (last data) 0)]
+    (into data (build-data 1000))))
+
+(defn update-some [data]
+  (reduce (fn [data index]
+            (let [row (get data index)]
+              (assoc data index (assoc row :label (str (:label row) " !!!")))))
```
lucywang000, line 36: the let block could be simplified as

```clojurescript
(fn [data index]
  (update data index update :label #(str % " !!!"))
```

frameworks/keyed/helix/src/demo/state.cljs
Lokeh: I haven't changed this from the previous implementation. `data` is still an immutable vector, so `(get data 1)` will still be the old value.

frameworks/keyed/helix/src/demo/state.cljs
Lokeh: Nice

frameworks/keyed/helix/src/demo/state.cljs
lucywang000: Oh right, I'm obvously affected by the imperative style thinkging here :P

krausest:
Thanks - nice [improvement](https://krausest.github.io/js-framework-benchmark/current.html) and sorry for the delay. The update to chrome 83 was challenging and took much time (#741).

PR #739, opened 2020-05-22
Add Doz non-keyed with new version

fabioricali:
This PR update the version to 3.0.1 and inclues non-keyed test. I still apologize for the previous PR.. -.-"


krausest:
Thanks. Results are [updated](https://krausest.github.io/js-framework-benchmark/current.html).

PR #743, opened 2020-06-01
Dojo: Update to dojo 7 and fix non-keyed impl

### LINKED ISSUE #742: Fix dojo non-keyed
For some reason select row is broken for dojo in the non-keyed version.
Fixing the versions doesn't help.
@agubler Can you please take a look at it or should be just remove the non-keyed version?

agubler:
Fixes the non-keyed version of Dojo and also upgrades both version to the latest release (7.0.0).

cc @krausest 

With regards to https://github.com/krausest/js-framework-benchmark/issues/742#issuecomment-634919576, we are working on improving the core support in the rendering engine which hopefully will be available  soon. I tested a draft implementation with the keyed check and it passed so should resolve the issue.

Resolves #742

### DIFF
--- a/frameworks/keyed/dojo/package.json
+++ b/frameworks/keyed/dojo/package.json
@@ -4,17 +4,19 @@
   "js-framework-benchmark": {
     "frameworkVersionFromPackage": "@dojo/framework",
     "customURL": "/output/dist",
-    "issues": [694]
+    "issues": [
+      694
+    ]
   },
   "scripts": {
     "build-prod": "./node_modules/.bin/dojo build"
   },
   "dependencies": {
-    "@dojo/framework": "^6.0.0"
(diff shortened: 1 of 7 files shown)


krausest:
Sorry, the non-keyed version seems to have an issue with remove row.
When I click on the delete row button in one row it doesn't remove that row but the next one. Can you take a look at that please?

agubler:
Oh that’ll teach me for not testing properly. Will do.

agubler:
@krausest this should be updated now.

krausest:
Thanks - that worked fine.  [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated.

agubler:
Thanks @krausest, once we've fixed the move scenario for keyed, we'll probably remove the non keyed implementation.

PR #744, opened 2020-06-01
Update RE:DOM and it's dependencies

pakastin:


krausest:
Thanks.  [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated.

PR #745, opened 2020-06-03
Dojo: Improve the dojo keyed implementation

agubler:
A couple of tweaks to hopefully improve the results in a couple of the benchmark tests.


krausest:
Results are [updated](https://krausest.github.io/js-framework-benchmark/current.html) and indeed a bit faster.

PR #750, opened 2020-06-16
Migrate yew to use wasm-bindgen instead of stdweb

dodomorandi:
Yew now supports `wasm-bindgen` instead of `stdweb`, and this new approach is recommended (and the benchmarks more realistic).

The major difference is that `wasm-pack` must be used  instead of `cargo-web`, and the `view` method must be implemented in the `impl Component for Model`.

Unfortunately `npm run check` did not pass before my PR and nothing has changed about that.


krausest:
Thanks. [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated, but it looks like overall  yew's results haven't improved much in comparison to the [old](https://krausest.github.io/js-framework-benchmark/2020/table_chrome_83.0.4103.106.html) results.

dodomorandi:
Thank you @krausest! Yes, I noticed that performances are almost the same, but it is fine as long as the approach is comparable to how yew is recommended to be used. Let's see if with better JS-WASM interop something will change :blush:

ryansolid:
Yeah to be fair perf difference beween StdWeb and WASM-Bindgen is noticeable. But once you had a higher level library abstraction there is probably enough overhead that without the most efficient approach it isn't big enough compared to that overhead to really notice a difference.

PR #752, opened 2020-06-19
Fixed MobX benchmark after upgrading by swapping minifier and enabling batching

mweststrate:
This fixes: https://github.com/mobxjs/mobx-react/issues/877

React became really slow after upgrading to 16.6+, which is caused by a bug in the default webpack minifier (https://github.com/facebook/react/issues/13987). This might apply to other React based frameworks as well. 

Also enabled ReactDOM batching support according to best practice.

### DIFF
--- a/frameworks/keyed/react-mobX/package.json
+++ b/frameworks/keyed/react-mobX/package.json
@@ -30,13 +30,14 @@
     "babel-preset-react": "6.24.1",
     "babel-preset-stage-0": "6.24.1",
     "jsx-loader": "0.13.2",
+    "terser-webpack-plugin": "^3.0.6",
     "webpack": "4.16.3",
     "webpack-cli": "^3.1.0"
   },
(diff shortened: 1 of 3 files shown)


krausest:
Thanks. [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated. I hoped to see a clear improvement in comparison to the [old](https://krausest.github.io/js-framework-benchmark/2020/table_chrome_83.0.4103.106.html) results, but sadly results didn't change much. (System was rebooted before runnning and I'm always running keyed/vanillajs as a check and its results look fine)
How much of an improvement did you expect?

mweststrate:
I'd expect them to be roughly the same indeed. We didn't make much
performance improvements that would show up in this benchmark. MobX is
kinda optimized for fine grained updates to widely shared state and when
complicated derived data comes into play. But that is not the case in this
benchmark :)

Issue #749, opened 2020-06-07
Better name for vanillajs-1


benmccann:
What is vanillajs-1? Could we come up with a more descriptive name?

krausest:
Maybe vanillajs-#2 would be better :rofl:
I always wanted to get back to a single vanillajs implementation, but I haven't managed to decide which one to keep. 
Maybe I just should stick with the one that will be faster on chrome 84.

noelmace:
I need to compare performance variances between different rendering approaches using vanillajs (plus some micro-libraries) and, after some investigation on my own, this project seems the best fit for that.

So, I forked the repo and will create some more vanillajs implementations. @krausest Should I make a PR with those when I'm done? (I guess not based on your comment `"I always wanted to get back to a single vanillajs implementation"` :smile:) If so, I could rename existing vanillajs implementations for more clarity.

Also, if you would be kind enough to share more information about your own investigations on this specific topic, it would help a lot.

krausest:
The only thing better than one vanillajs implementations would be several vanillajs implementations :smile: but with clear and distinct implementation strategies (and appropriate naming). So I'd say we should take a look at those implementations and discuss here which should be added. But I want to restrict it to adding only keyed vanillajs implementations.

Sadly I have failed to make vanillajs the fastest implementation so I can't give you good advice.

ryansolid:
The difference between the 2 implementations is VanillaJS keeps an array of nodes around and VanillaJS-1 does not. This reduced memory usage(although as of late that advantage has almost vanished) and code size.

You would think improve performance not keeping an array in sync. However, as we know the DOM is expensive. So stuff like this hasn't really made a big impact to get past any variance. Sometimes I think that walking down the 2 nodes from either end on the Swap test is more expensive than having the array for the entirity of the tests.

noelmace:
> I want to restrict it to adding only keyed VanillaJS implementations. - @krauset

IMO, the distinction between keyed & non-keyed approaches isn't as relevant for VanillaJS as when using frameworks. Because a VanillaJS approach could be "partially keyed", and it would be nice to compare all the VanillaJS implementations between them more easily.

As I'm only focusing on VanillaJS implementations right now, I was thinking about creating a third category for them. This way, we could just show one of them in the other categories, for comparison purposes.

noelmace:
> I have failed to make VanillaJS the fastest implementation - @krausest

[...]

PR #754, opened 2020-06-19
(WIP) [discussion] investigate vanillajs approaches

noelmace:
**This PR is for feedbacks and discussions only**

> follows https://github.com/krausest/js-framework-benchmark/issues/749#issuecomment-646592970 & https://github.com/krausest/js-framework-benchmark/issues/749#issuecomment-645554827

As I'm currently using this project in order to investigate different VanillaJS practices and their impacts on rendering performance, I'm creating some new poorly optimized implementations. Those aren't intended to be merged as is, as they are out of this project's scope.

Yet, sharing those implementations with contributors to this project may help some discussions about the current ones.

Especially:
- I would be happy to make (based on `vanillajs-wc-dx`, but in another PR) the `vanillajs-wc` implementation follow good practices in order to be closer to "real-life" projects, even if it means worse results.
- I'm currently trying to list as many vanillajs rendering approaches and their impact on performance. So any advice, feedback, and help are more than welcome.

FYI: in order to allow me to follow how rendering approaches compares without creating a bunch of implementations, I commit every results, so of course, those are conflicting with the master branch.


ryansolid:
@noelmace Awesome work on so many variants. Any observations so far?

noelmace:
Whoops, I forgot about this PR :sweat_smile: Sorry for the flooding, here is a quick summary of some of my observations so far.

## the "good" VanillaJS approach

So, first, the generally advocated VanillaJS approach is kind of dumb, as it leads to ~30% worse rendering performance than an optimized one, and even ~10% compared to most rendering micro-libs (e.g. lit-html).

> Plus, it's less than 10% better than Preact & Co, which offers a way better DX.

This is mostly because:
- using innerHTML w/ a template sting is ~10% (9-13%) slower than the programmatic approach (createElement etc.)
- just using a custom element makes rendering ~10% (7-15%) slower
    - w/o even following good practices like reflecting properties and attributes (which makes it even slower)

I'm still investigating different "minimal vanillajs component" approaches (no custom element, pre-parsed or programmatic template, no advanced function, etc...) w/ some helper functions.

But I'll need to work on a solution to test shadow dom to compare those with "real life" Web Components.

### Some other general considerations

- it's generally better to use just one global event handler instead of local ones
    - even if the callback becomes very slow, performance is better most of the time
    - DX is often better
    - this is mostly related to templating/rendering, as setting local event handlers often leads to more code
    - this is just a global constatation, I'll need to give more thoughts to that
- use a "DOM buffer" (i.e. store the elements in collections) when you can, it can increase performance a lot

## In general

- `textContent = ""` is always the best solution
    - but `parentNode.replaceChild()` leads to pretty much the same performance
- `requestAnimationFrame` can lead to better rendering performances
   - but not enough to make it very useful considering to awful associated DX (except for specific use cases of course)
- `append()` is slightly faster than 2 or more `appendChild()` (yeah, obviously :shrug:)

## Some comments on the pre-existing examples

- "selection" with lit-html is very slow, yet this doesn't seem to be related to lit-html itself. I'll investigate the actual demo later, let me know if you have any idea...
- globally, vanillajs only uses "one-time rendering components" for rows, which IMO is kind of misleading. For example, if I add any `disconnectedCallback()` to a row custom element, this slows down clear by x2. I can remove it, but only because I don't have to care about the component being reconnected after that in this specific use case. IMO, this is far from a "real use case" example
- `vanillajs-wc` always skipp `rowId = item.id` and `selectedRow = null`, which makes it a little bit inconsistent with some other examples

noelmace:
I made a lot of changes in my fork, and this branch has been removed in favor of `fullwebdev/new-vanillajs-implementation`.

Please check the dedicated issue I created to discuss differences between this fork and this project: https://github.com/fullwebdev/benchmark/issues/1

PR #755, opened 2020-06-20
Add @vue/composition-api

antfu:
[@vue/composition-api](https://github.com/vuejs/composition-api) is the official plugin for Vue 2 to support Vue 3's Composition API. 

Adding this framework to the benchmarks is mainly for users to know its extra costs.

Thanks.

### DIFF
--- a/frameworks/keyed/vue2-composition-api/.babelrc
+++ b/frameworks/keyed/vue2-composition-api/.babelrc
@@ -1,0 +1,11 @@
+
+{
+    "presets": [
+        ["@babel/preset-env", {
+          "targets": {
+            "browsers": ["last 1 chrome versions"]
+          }
+        }]
+    ],
+    "plugins": []
+}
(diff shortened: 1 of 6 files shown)


ryansolid:
That's interesting. It reflects the difference between VDOM approaches between Vue 2 and 3 too I imagine since it's already been shown the Composition API(Reactivity) without Vue has very little overhead(vuerx-jsx). There can be potentially a real cost when combining VDOM with reactivity depending on approach.

krausest:

ryansolid:
Super interesting. Not saying something might be here in the implementation but the overhead composition API adds to Vue 2 is significantly more than the overhead of Vue Reactivity over VanillaJS. Which makes me think that a decent amount of gains were made in how they interface their VDOM with the reactivity system in Vue 3. Basically Vue 2 can't leverage it in the same way and its pure overhead, but Vue 3 they much have changed some very fundamental things.

Now it's possible there is still some benefit of not using composition API bits in Vue 3 for similar reasons and it is just their VDOM performance that has improved. But that would suggest that the interopt between Reactivity/VDOM is really expensive as we know the raw reactivity system isn't much overhead. More than likely they've drastically improved both areas, but that is interesting.

PR #759, opened 2020-06-30
update san to 3.9.0, for #694

### LINKED ISSUE #694: Strengthen isKeyed test for swap rows

errorrik:
to fix #694

### DIFF
--- a/frameworks/keyed/san/package.json
+++ b/frameworks/keyed/san/package.json
@@ -1,10 +1,9 @@
 {
   "name": "js-framework-benchmark-san",
-  "version": "3.7.6",
+  "version": "3.9.0",
   "description": "Benchmark for san framework",
   "js-framework-benchmark": {
-    "frameworkVersionFromPackage": "san",
-    "issues": [694]
+    "frameworkVersionFromPackage": "san"
   },
   "scripts": {
     "build-dev": "webpack -w -d",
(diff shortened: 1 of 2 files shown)


krausest:
Thanks for the fix. Results are [updated](https://krausest.github.io/js-framework-benchmark/current.html).

PR #760, opened 2020-07-02
Update ScarletsFrame

StefansArya:
Hi, I updated the framework version and simplify the implementation


krausest:
Thanks looks like append rows got faster. Results are [updated](https://krausest.github.io/js-framework-benchmark/current.html).

StefansArya:
Thanks,
Because I changed `concat` it with `push` for appending new value, the framework didn't need to recheck between 2 different arrays of 10k item from the beginning.

PR #761, opened 2020-07-04
Move yew from non-keyed to keyed framework

jstarry:
- Updates Yew to v0.17.2 which has keyed list support
- Split Yew implementation into multiple components
- Specify release profile config in Cargo.toml
- Use lighter weight random number generator


krausest:
Thanks. It's in the right table now :smile: and performance is much better. Results are [updated](https://krausest.github.io/js-framework-benchmark/current.html).

PR #762, opened 2020-07-04
build & install implementations using lerna

noelmace:
Add two new scripts using lerna in order to ease examples' dependencies install & build:
- `npm run bootstrap`: install all implementations' dependencies
- `npm run build-all -- --scope <glob>`: build implementations, filtered by their package names
    - example: `npm run build-all -- --scope "*vanillajs*"`

This required to rename most packages in order to avoid conflicts, as some haven't been updated after copy.

> FYI lerna can't be used if several packages have the same name.

### DIFF
--- a/frameworks/keyed/angular-ng/package.json
+++ b/frameworks/keyed/angular-ng/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "angular-ng",
+  "name": "js-framework-benchmark-angular-ng",
   "version": "0.0.0",
   "js-framework-benchmark": {
     "frameworkVersionFromPackage": "@angular/core",
--- a/frameworks/keyed/angular-noopzone/package.json
+++ b/frameworks/keyed/angular-noopzone/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "angular-closure",
+  "name": "js-framework-benchmark-angular-noopzone",
   "version": "1.0.0",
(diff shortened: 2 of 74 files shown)


krausest:
Thanks. It's merged, but I'd like to name the targets lerna-bootstrap and lerna-build-all.
I'm not yet sold on switching to lerna since I had same painful experience with yarn (#221, #182) and so I avoided looking into lerna so far.
Currently I'm having a rather robust workflow for updating implementations with the docker-rebuild target to do a safe rebuild. Do you have any recommendations how to improve this with lerna?

noelmace:
Well, actually, I only used lerna here as a quick and dirty way to ease implementations install & build for other contributors (me included). I'm not familiar enough with the docker-rebuild workflow to have any useful opinion on that. I'll try to give it a try when I'll have access to a more powerful computer :sweat_smile: 

But I don't expect Lerna to break anything like yarn used to, as it simply use npm under the hood.

krausest:
I put a description of the docker build here: https://github.com/krausest/js-framework-benchmark/wiki/Build-all-frameworks-with-docker
I guess I wouldn‘t use it for development, since the turnaround times of a clean npm install and npm build-prod is much higher than hot reloading like the build-dev target often allows. But it‘s really great for maintenance since the docker-rebuild target allows rebuilding and checking one implementation in a controlled environment.

PR #763, opened 2020-07-05
Add sexy framework benchmark

lastonoga:
Add [sexy-framework](https://github.com/sexy-framework/sexy) client-side performance


krausest:
Updating every 10th row doesn't work yet. Can you please fix it?

PR #765, opened 2020-07-07
remove vanillajs build

noelmace:
Using Babel, Webpack & Rollup could lead to different performance results.

This PR removes all VanillaJS implementations build step in order to keep to 100% "Vanilla".


krausest:
Thanks! It's merged now.

PR #767, opened 2020-07-08
remove umaintained implementations

noelmace:
Remove implementations using dependencies without activity on github or npm for more than a year.

See README.md for the details.


krausest:
Thanks - I decided to keep crui (a week to early for retirement) and marionette (too famous).
React-easy-state was updated to @rising-stack/react-easy-state.

PR #768, opened 2020-07-09
new vanillajs implementation with shadow dom on table rows

noelmace:
attachDom can't be called on a tr element. Therefore, using an autonomous custom element is mandatory.
This leads to some issues with bootstrap, as it's expecting .table > tbody > tr.
In order to test this correctly, the list of elements shouldn't be a table, and bootstrap shouldn't be used.

See https://github.com/krausest/js-framework-benchmark/pull/764#issuecomment-655116868 & https://github.com/krausest/js-framework-benchmark/issues/753#issuecomment-656050278

### DIFF
--- a/frameworks/keyed/vanillajs-shadow/index.html
+++ b/frameworks/keyed/vanillajs-shadow/index.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
+    <title>VanillaJS-shadow-"keyed"</title>
+    <link href="/css/currentStyle.css" rel="stylesheet" id="root-style"/>
+</head>
+<body>
+<div id='main'>
+    <div class="container">
+        <div class="jumbotron">
+            <div class="row">
+                <div class="col-md-6">
+                    <h1>VanillaJS-shadow-"keyed"</h1>
+                </div>
+                <div class="col-md-6">
+                    <div class="row">
+                        <div class="col-sm-6 smallpad">
+                            <button type='button' class='btn btn-primary btn-block' id='run'>Create 1,000 rows</button>
+                        </div>
+                        <div class="col-sm-6 smallpad">
+                            <button type='button' class='btn btn-primary btn-block' id='runlots'>Create 10,000 rows</button>
+                        </div>
+                        <div class="col-sm-6 smallpad">
+                            <button type='button' class='btn btn-primary btn-block' id='add'>Append 1,000 rows</button>
+                        </div>
+                        <div class="col-sm-6 smallpad">
+                            <button type='button' class='btn btn-primary btn-block' id='update'>Update every 10th row</button>
+                        </div>
+                        <div class="col-sm-6 smallpad">
+                            <button type='button' class='btn btn-primary btn-block' id='clear'>Clear</button>
+                        </div>
+                        <div class="col-sm-6 smallpad">
+                            <button type='button' class='btn btn-primary btn-block' id='swaprows'>Swap Rows</button>
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </div>
+        <table class="table table-hover table-striped test-data">
+            <tbody id="tbody">
+            </tbody>
+        </table>
+        <span class="preloadicon glyphicon glyphicon-remove" aria-hidden="true"></span>
+    </div>
+</div>
+<script src='src/Main.js'></script>
+</body>
+</html>
--- a/frameworks/keyed/vanillajs-shadow/package.json
+++ b/frameworks/keyed/vanillajs-shadow/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "js-framework-benchmark-vanillajs-shadow",
+  "version": "1.1.1",
+  "description": "Vanilla.JS demo",
+  "main": "index.js",
+  "js-framework-benchmark": {
+    "frameworkVersion": "",
+    "useRowShadowRoot": true
+  },
+  "scripts": {
+    "build-dev": "exit 0",
+    "build-prod": "exit 0"
+  },
+  "keywords": [
+    "react",
+    "webpack"
+  ],
+  "author": "Stefan Krause",
+  "license": "Apache-2.0",
+  "homepage": "https://github.com/krausest/js-framework-benchmark",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/krausest/js-framework-benchmark.git"
+  }
+}
--- a/frameworks/keyed/vanillajs-shadow/src/Main.js
+++ b/frameworks/keyed/vanillajs-shadow/src/Main.js
@@ -1,0 +1,383 @@
+'use strict';
+
+function _random(max) {
+    return Math.round(Math.random()*1000)%max;
+}
+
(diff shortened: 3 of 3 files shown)


ryansolid:
You are hitting on why you never see shadow Dom in benchmarks. The need to mess with styles and change html structure basically makes it non-comparable. I hit this repeatedly when trying to make demos and benchmarks early on. As said in a previous thread i've built versions of this using native built-ins but never Shadow Dom for this reason.

Even with all custom styles and a non-restrictive HTML structure WebComponent version is creating more elements and doing more work. Trying to accommodate this across all benchmarks is a sizeable change.

noelmace:
> You are hitting on why you never see shadow Dom in benchmarks.

IMO this project is pretty specific. Styles modularity can be achieved with or without a shadow dom. Here, the main issue is that this project is using a CSS library (bootstrap). And using `table > tbody > tr` isn't mandatory by itself. This wasn't built with this kind of problem in mind, and this is OK. But now it's too late for that.

> Even with all custom styles and a non-restrictive HTML structure WebComponent version is creating more elements and doing more work. Trying to accommodate this across all benchmarks is a sizeable change.

Yep, this is exactly what I'm thinking too. Maybe a good fit for my fork, but not for this project.

I don't believe this PR should be merged (it's just an answer to https://github.com/krausest/js-framework-benchmark/pull/764#issuecomment-655116868). Neither should have been #764 (which was a WIP) without discussing more about this.

ryansolid:
> IMO this project is pretty specific.

Other good or popular benchmarks:
https://github.com/localvoid/uibench
https://github.com/mathieuancelin/js-repaint-perfs
https://github.com/somebee/dom-reconciler-bench

Demos
https://github.com/gothinkster/realworld
https://github.com/tastejs/todomvc
Most hackernews implementations

The one thing they always have in common is non-modular CSS and more often than not for measuring purposes specific DOM structures. I do think the CSS can be worked through in a lot of the demos. And we've started to see WebComponent specific benchmarks(https://github.com/vogloblinsky/web-components-benchmark, or https://webcomponents.dev/blog/all-the-ways-to-make-a-web-component/). But it definitely highlights why you don't see even webcomponent libraries like Stencil or LitElement use the Shadow DOM in these demos.

I was talking to @trusktr recently about the desire to take a fork of JS Framework Benchmark in this direction, and I admitted I probably wouldn't have time but I think there would definitely be an interest there. Most webcomponent benchmarks are simplistic based on creation time or things like size. I will say from having participated in a few there are some standardization considerations. Some libraries using passed handlers versus DOM events to communicate upwards makes a pretty decent performance difference. But that's the whole thing. Maybe I've witnessed this, but this kind of data isn't out there for general consumption. I'm positive there is an interest.

krausest:
Since this benchmark includes rendering in the duration measurement it's very important to prevent optimizations for the styles for individual implementations.

I see two options:

1. It should pretty easy to change the style for all benchmarks. Maybe it's time to drop the alternating row colors, create simple but still nice looking styles that are compatible with web components (but I'm not sure what's the best way to achieve that...) and see what impact it has on keyed vs. non-keyed.

2. Allow additional css rules for webcomponent frameworks that create exactly the same look as bootstrap does. But I'm not sure how much those different styles influence the measurement. This might be dangerous.

What's your opinion?

noelmace:
> @krausest It should pretty easy to change the style for all benchmarks

Big :+1: on this option.

> @ryansolid I was talking to @trusktr recently about the desire to take a fork of JS Framework Benchmark in this direction, and I admitted I probably wouldn't have time but I think there would definitely be an interest there.

Me too, definitely. Like I said in https://github.com/krausest/js-framework-benchmark/issues/749#issuecomment-646012074, I'm/was thinking about coding:
1. A project benchmarking just small VanillaJS approaches (e.g. innerHTML vs createElement) w/o anything else, in order to minimize the risk of unwanted side effects on performance. Of course, when talking about rendering, we need to get the whole parse+paint timing, like with this project. Maybe using puppeteer `trace`, or simply `webdriver-ts`.
2. A tool allowing the automate web components' performance testing based on [custom-elements-json](https://github.com/webcomponents/custom-elements-json) files.

Unfortunately, I won't have time for that until the end of the year.

Actually, I spent too much time on this :sweat_smile: I gathered plenty enough information (thx to you :pray:), and I'll need to postpone all other discussions in order to focus on my actual work for some time. But I'll definitely get back to it ASAP.

ryansolid:
Yeah I think the first option seems reasonable for CSS. I think the more interesting thing is necessity of injecting the element in the table. @krausest does any of the checks look at the structure of the HTML. Correct me if I'm wrong @noelmace , but basically with webcomponents and shadow dom the structure is going to become:

```html
<table>
  <tbody>
    <custom-row>
      #shadow-root
        <style />
        <tr>
          ...standard row stuff
        </tr>
    </custom-row>
  </tbody>
</table>
```

noelmace:
@ryansolid nope, that it. Using an element compatible with  `attachShadow` would be better, but this would need too many changes.

krausest:
So I created a very simple style which makes vanillajs and vanillajs-shadow look equal:
image
The results then look like that:

I know too little about web components. How often is the linked style evaluated when appending 1,000 web components?

noelmace:
How did you load the CSS in the shadow dom? Are you using a `<link>`? The shadowed element seems very slow to render, compared to some other benchmarks I made in the past. But those other benchmarks were using very few CSS, so maybe this is only due to an heavy CSSOM to construct for each shadow.

Maybe @ryansolid will have some insights about that.

krausest:
There‘s a style link cloned from the host document in the PR. See line 41 in Main.js:
https://github.com/krausest/js-framework-benchmark/blob/e5d50cea6b35a454c270ca167e36d33cd2a73423/frameworks/keyed/vanillajs-shadow/src/Main.js#L41

ryansolid:
That's interesting. There were a lot of things in spec or intended but I'm not sure how they landed in the browser. Chrome team at one point had talked about intention to optimize the processing of such style tags given that they could appear on the page multiple times but I'm not sure that ever happened. Honestly I'm unclear the fastest way anymore for Web Component styles. You should expect link tag to have cached the content so it should be able to pull already processed CSS.

noelmace:
A `<link>` will always lead to a fetch event. The only optimization a browser can do here is to use the cache. This is why I believe this is always slower than using a style.

I did some experiments recently to find a solution to pre-load an external CSS. First, to avoid any duplicate fetch, and second to minimize FOUC. Unfortunately, I was unable to find any good solution for that. This is why I recommend avoiding `<link>` in shadow dom as much as possible. Yet, I wasn't thinking about the use case we have here. Maybe the link could allow us to optimize rendering as parsing is already done? :shrug: 

@krausest could you share your code and benchmark another implementation using a `<style>` tag instead, with the same CSS, just to be sure?

krausest:
@noelmace I pushed the code to the branch vanillajs-shadow: https://github.com/krausest/js-framework-benchmark/tree/vanillajs-shadow
The vanillajs-shadow is your version and vanillajs-shadow-inline-style uses only inline styles.

I got the following results:

krausest:
I'm closing this PR. I don't think we're getting it into master.

PR #769, opened 2020-07-14
add @fullweb/helpers 0.1.0 implementations

noelmace:
Small helpers library to demonstrate some of the stuff I learned thanks to this project.

Check https://github.com/fullwebdev/fullwebdev/tree/master/packages/helpers for the source.

### DIFF
--- a/frameworks/keyed/fullweb-helpers/index.html
+++ b/frameworks/keyed/fullweb-helpers/index.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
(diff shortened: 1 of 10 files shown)


krausest:
Thanks. Since we've had many discussions about data-driven frameworks and apple-to-oranges discussions and both fullweb implementations would easily allow to open that discussions again i decided to mark those implementations. Maybe this is an acceptable solution. Of course vanillajs had to marked too. (Maybe I'll remove the red color in the th for this issue when I find some time since it's not really an issue but just a note.)

Issue #239, opened 2017-08-28
Lock down dependencies


alexfmpe:
We probably should add a [package-lock.json](https://docs.npmjs.com/files/package-lock.json), to fix dependency problems.
[Build appears to be broken right now](https://circleci.com/gh/alexteves/js-framework-benchmark/32), and since the same commit passed CI[ just a bit ago](https://circleci.com/gh/krausest/js-framework-benchmark/38), I'm guessing some package is picking up a new version.

Note: circleci dashboard is having issues at the moment.

alexteves:
Apparently whether package-lock 'locks' depends on nvm version: https://github.com/npm/npm/issues/17979

krausest:
Just ran into the same topic when ncu and npm install actually updated the libraries despite package-lock.json being present.
See also 
https://stackoverflow.com/questions/45022048/why-does-npm-install-rewrite-package-lock-json
I'll stick with requiring fixed version numbers in package.json - all build problems I've encountered so far were caused by version ranges in package.json.

PR #773, opened 2020-07-22
add package-lock files

noelmace:
Not committing `package-lock.json` files may introduce several issues when installing the dependencies from scratch, especially when using Lerna.

The only related issue I could find was #239.


krausest:
Thanks!

PR #775, opened 2020-07-26
feat: sinuous without htm build

noelmace:
New Sinuous implementation without `babel-plugin-htm`, in order to keep it closer to other implementations (like lit-html for example).

It seems to perform pretty much the same tho :shrug: 
image

I hope to find some time soon to add an implementation with `htm` in order to investigate this further (maybe on top of `@fullweb/helpers` in order to keep it simple, and/or with Preact).

@developit & @luwes, kudos to you, you are doing an amazing job here! Do not hesitate to share some information about this, it would help a lot.

### DIFF
--- a/frameworks/keyed/sinuous-run/.eslintrc.yml
+++ b/frameworks/keyed/sinuous-run/.eslintrc.yml
@@ -1,0 +1,21 @@
+env:
+  browser: true
+  es6: true
+
+extends:
+  - eslint:recommended
+
+parser: babel-eslint
+
+parserOptions:
+  ecmaVersion: 9
+  sourceType: module
+
+globals:
+  html: readonly
+
+rules:
+  semi: error
+  no-unused-vars:
+    - error
+    - varsIgnorePattern: ^h$
--- a/frameworks/keyed/sinuous-run/index.html
+++ b/frameworks/keyed/sinuous-run/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
+    <title>Sinuous-run-keyed</title>
+    <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+<div id='main'></div>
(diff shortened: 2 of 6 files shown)


ryansolid:
You should check out the `template` function. The overhead of tagged template literals is not going to change much here since most of the benchmark is just running the same one. But the `template` method is clever.

See pre-compilation and just-in-time compilation are more typical at the front end of this benchmark since you can optimize away decision code. Solid, Mikado, DomC, Fidan, Stage0, etc... All make use of this. Even Solid's tagged equivalent to Sinuous uses JIT compilation. The performance has overhead is almost nothing there either. The trick is to use the template strings to analyze slots and generates code almost as performant as pre-compilation.

But of the JIT solutions other than Solid they are top down so they take the hit in the more granular benchmarks like partial update or select row.

But Sinuous approach actually works with just the HyperScript even without tagged templates. But eschews the heavier more flexible reactive system for a streamlined single purpose one in the scope of the template. It also caches the nodes to perform cloning the way the compiled libraries do. So it doesn't suffer the slow down of top down nor the memory overhead and GC of a typical (including its own) reactive system.

It has limitations from a app scale perspective but for a benchmark like this it is very difficult to match. You won't see compilers use an approach like this as it isn't generalizable. But it feels like the right sort of tool for Sinuous a la carte approach.

Like without this approach a library could only hope to score in the 1.18 range. (I know I've written one). But it's a complete game changer in benchmarks.

luwes:
thanks for the addition @noelmace! 🙌 this is great but yes like @ryansolid pointed out in performance it's pretty much the same because of the way how the `template` module works, the biggest difference will be in size, +700B or so for the Sinuous HTM module.

krausest:
I included it in the latest [results](https://krausest.github.io/js-framework-benchmark/current.html)


Results are so close that I hesitate to merge this PR. Do you see much benefit from keeping it?

luwes:
I don't see much benefit in adding the results to the table (would add noise) but it's good to see the difference is so small. Unless @noelmace feels very strongly about it...

PR #776, opened 2020-08-02
Add non-keyed yarte-wasm-app

botika:
Add valid isomorphic wasm app for [`yarte`](https://github.com/botika/yarte). Later all this will be done with a `derive` but the runtime is already stable and I want to bench it.


botika:
Any problem? It passes the same checks as the vanillajs version.

krausest:
Sorry, no problems except for a lack of time. I'll review the PR this weekend.

krausest:
I guess you know that I'm not a fan of non-keyed, but yarte is quite fast - nice!

PR #777, opened 2020-08-05
update Doz to 3.4.0

fabioricali:
Hi, in this release I have rewritten much of the template engine. Should improve a little.


krausest:
Very nice. Doz 3.4 is now 1.95 on average, down from 3.5

PR #778, opened 2020-08-06
Update fntags to 0.2.4

narcolepticsnowman:
This update changes to use a simple observer instead of the es6 proxy.


krausest:
Sorry, overall it's a bit slower. 0.2.4 gained a bit on most benchmarks, but lost quite a bit for append to large table.

narcolepticsnowman:
so close, but so far away

PR #779, opened 2020-08-10
Update fntags to 0.2.10

narcolepticsnowman:
This has a new mapChildren function, should fix the slow append to large table


krausest:
Thanks - results have been updated. Package.json is still on 0.2.4. Is that intended?

narcolepticsnowman:
Nope, I should have updated that. I'll fix it when I figure out how I screwed up select row lol.

Thanks a ton for creating and managing this awesome project!

krausest:
You‘re welcome! I‘m on holidays for the next two weeks so there‘s no need to hurry with a PR :-)

PR #780, opened 2020-08-11
Update ScarletsFrame

StefansArya:
Hi, I want to update the framework with "scarletsframe.es6.js" that doesn't include hot reload module and may be optimized for modern browser.


krausest:

StefansArya:
Thanks, I though the memory and the startup time would be reduced a lot. but it was not 😅

PR #782, opened 2020-08-15
update Doz to 3.4.1

fabioricali:
This update improve non-keyed performance... maybe even for keyed!


krausest:


And non-keyed got indeed much faster:

fabioricali:
Thanks... There are still some red boxes🤣 .. But much better than the first PR.

PR #783, opened 2020-08-16
use faster select row impl

narcolepticsnowman:


StefansArya:
It seems some implementation was using direct DOM manipulation while it already mentioned on [How To Contribute](https://github.com/krausest/js-framework-benchmark#how-to-contribute) section. The purpose of this project is to compare a framework with another framework where vanilla JS is the main baseline for the benchmark.

Adding vanilla JS on the implementation means the framework itself isn't really benchmarked, [(ref)](https://github.com/krausest/js-framework-benchmark/pull/521#issuecomment-463026879). If other benchmark implementation begin to do this, it's possible that every implementation could have identical result.

ryansolid:
Yeah that thread has the whole conversation which is a good reference. But to summarize, while the direct DOM manipulation is controversial, what needs to be present is keeping the selected state in the data and not in the DOM. Otherwise it is not data-driven. If the initial state was row 10 was selected would this render properly?

Mind you this whole implementation sort of sidesteps the library in a big way. Like the row implementation doesn't use the templating system anymore and breaks into imperative DOM construction. I mean if that is how the library wants to showcase itself. But I'm a bit surprised you wouldn't use the same HyperScript syntax throughout when building apps.

narcolepticsnowman:
@StefansArya 
I see your point, I will close this pr and I'll make my library suck less.

@ryansolid 
It certainly was prettier before I changed the row implementation to the clone style it is now, but this being a benchmark I switched it to that faster clone implementation because the library allows for it and certainly doesn't discourage it.

As far as this being a showcase, I think it's fine. I don't agree with the opinion that we shouldn't manipulate the dom. Some things really make more sense, and are easier to do that way, and It would be beneficial if everyone understood the dom better (not saying you don't, but the general population of developers), but that won't happen if we keep hiding it in our frameworks. I feel this demonstrates the flexibility of the library to behave in either way. The premise of this library was to step back a bit from the enclosed worlds of react and angular, and to allow it to be more flexible and easier to get started.

narcolepticsnowman:
@ryansolid  

I'd also be interested in the other "big ways it sidesteps the library"

StefansArya:
@narcolepticsnowman

Well I'm not forcing you to close the PR or giving pressure 😅 , but actually you can still commit on your branch to do the improvement. Personally in the past I also disagree that we can't direct manipulate the DOM, but now I think it's acceptable.

You can expose the DOM on some of your library's feature if you allow developer for accessing it. Some developer may like an expressive library where it does the job in single short line, while using vanilla would take 2 or more line.

Dev time cost could be faster and the only matter is the library's performance cost. Maybe like comparing jQuery and Zepto.

ryansolid:
@narcolepticsnowman I probably looked to quickly at it. I see the template now for the row. I was sort of focused on https://github.com/krausest/js-framework-benchmark/blob/master/frameworks/keyed/fntags/src/Main.js#L45 and just saw a bunch of directed dom node crawling(`nextSibling`, `firstChild`) and attaching event handlers and setting dom attributes. I see now it was originally rendered with the HyperScript.

What I mean by sidesteps is you are no longer using the templating syntax(HyperScript) to decorate and declaratively bind your elements. That could have effectively been an innerHTML on a template element. And call the same function and be similar to the vanillajs implementation. I see there is a `fnbind` method which ties state updates so to be fair it looks to be considered use case. It just is sort of a gray area when we aren't binding the elements using the templates. I made a prediction in the other threads and in #430 that more people looked at lighter weight it's just the DOM abstractions we'd have a hard time justifying not doing some level of DOM updates as DOM awareness is part of the library.

narcolepticsnowman:
@StefansArya No pressure felt. I agree that the current impl is more appropriate, since its initial state is rendered correctly if it were selected. Any changes that need to happen should be in the library, so this pr is moot. No hard feelings, promise. I say my stuff sucks all the time, not raging, I promise.  If we don't acknowledge our failures we can't improve.

@ryansolid 
It definitely could've been an innerHTML, and I totally shamelessly ripped this off from the vanillajs impl. Part of my goal of the library was to ditch html, and let the user directly build the dom. Maybe I'm strange, but ever since I could build a UI, I preferred to completely build my ui programmatically and to forego using htlm/xml templates. I just saw more code to learn, and more mental load, because the user has to now map these two things together in their mind. I also never liked how hard it is to do certain simple things when you don't have access to dom elements, hence I made a library that deals in dom elements directly, and only deals in es6 to close the gap. 

I do appreciate both of your time very much, and I love to see all of these projects growing together, learning from each other. This is what open source is about :).

PR #785, opened 2020-09-01
Update reaml dependencies

utkarshkukreti:
Just updating all dependencies. The benchmark runs a tiny bit faster with the latest version of bs-platform (the `ml` -> `js` compiler) for me.


krausest:

PR #786, opened 2020-09-05
add glimmer2 benchmark

lifeart:
image
image


lifeart:
I'm getting runtime error if I use npm 
image


With yarn all ok

krausest:
Sorry, doesn't work for me I'm getting the following error:
```
app.bundle.js:1 Uncaught Error: The `@glimmer/validator` library has been included twice in this application. It could be different versions of the package, or the same version included twice by mistake. `@glimmer/validator` depends on having a single copy of the package in use at any time in an application, even if they are the same version. You must dedupe your build to remove the duplicate packages in order to prevent this error.
    at Module.<anonymous> (app.bundle.js:1)
    at s (app.bundle.js:1)
    at app.bundle.js:1
    at app.bundle.js:1
```

lifeart:
@krausest is it possible to add yarn support to benchmarks?
Trying to force resolutions for `npm`, but no luck. 
For now, I suppose to wait for https://github.com/glimmerjs/glimmer.js/issues/303 (likely any new release will fix it)

krausest:
There once was yarn support, but it caused some issues like https://github.com/krausest/js-framework-benchmark/issues/182 
I‘ll try to build with yarn and post the results here and then I‘ll come back to answer the question.

krausest:
@lifeart Building with yarn worked. Here are the results (vanillajs and glimmer-2 were rerun. vanillajs looks reasonable):

I'd rather prefer to wait for a fix for npm. We've had a few nasty bugs (or at least inconsistencies from npm) in the past and the version lookup would have to be fixed.

lifeart:
@krausest thank you for this run! I Agree, with your concerns. I will convert MR to draft and will appear it back once new versions will be released (with fixed `npm`)

lifeart:
@krausest now it's alive for me

krausest:
This works fine. #800 holds for this implementation too.

PR #787, opened 2020-09-08
Add react tracked

dai-shi:
https://react-tracked.js.org

---
The src/main.jsx is based on react-redux-hooks and react-hooks.


krausest:

PR #789, opened 2020-09-13
Add LS-Element

lsegurado:
My implementation for [LS-Element](https://github.com/lsegurado/ls-element).


krausest:
I'm a bit short of time so I plan to merge the PR this weekend

lsegurado:
Thank you very much!

krausest:
There's an issue with remove row which doesn't work in the test driver. You can reproduce the issue with those steps:
1. Click create 1,000 rows
2. Delete row with number 1
3. Try to delete row with number 2, which now doesn't work

One minor note: Please change the package.json to the following:
```
  "js-framework-benchmark": {
    "frameworkVersionFromPackage": "ls-element",
    "useShadowRoot": false,
    "customURL": "/dist"
  },
```

lsegurado:
Hi @krausest! Thank you for your answer! Its fixed now!

krausest:
I've changed onpointerup event to onclick since the test driver wouldn't work otherwise.
I'll post results tomorrow.

krausest:
I'm sorry to say there's room for improvement...

PR #792, opened 2020-09-23
Add etch benchmark (revive)

aminya:
This adds etch back which has become active again:
https://github.com/atom/etch

Fixes: https://github.com/atom/etch/issues/87


krausest:
Thanks. Here are the results for the keyed version:

PR #793, opened 2020-09-24
keyed/preact: update preact to 10.5.2

utkarshkukreti:
I'm seeing some impressive improvements: in my local run, the slowdown before this PR = 1.34, after = 1.26!

### DIFF
--- a/frameworks/keyed/preact/.babelrc
+++ b/frameworks/keyed/preact/.babelrc
@@ -1,11 +1,22 @@
 {
-    "presets": [
-        ["@babel/preset-env", {
-          "targets": {
-            "browsers": ["last 1 chrome versions"]
-          }
-        }],
-        "@babel/preset-react"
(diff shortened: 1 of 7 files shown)


utkarshkukreti:
Pushed an update to switch to [React's new jsx-runtime](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) (which Preact already supports in 10.5.2). The slowdown went from 1.26 to 1.22!

utkarshkukreti:
(The slowdowns are on my system which seems to be running vanilla much slower than the results in current.html.)

utkarshkukreti:
I also updated Preact on keyed/reaml-preact to 10.5.2. I hope that's fine to do in the same PR.

krausest:
Thanks. Indeed a bit faster. Here are the new results:

And for reference the old results:

PR #798, opened 2020-09-27
updated svelte to v3.28.0

TheComputerM:


krausest:
Thanks - results are updated: https://krausest.github.io/js-framework-benchmark/current.html

Issue #772, opened 2020-07-21
Note: Implementation uses manual DOM manipulations [sticky]


krausest:
These implementations use direct DOM modification in the end user code. This means specific DOM updating code is written tailored for these specific tests. 
Those implementation are expected to perform very close to vanillajs (which consists of manual DOM manipulations).

ryansolid:
I believe the WASM implementations like `wasm-bindgen` and `stdweb` are in this category as well.

Worth mentioing `domdiff` (it's the underlying reconciler behind uhtml, lighterhtml, hyperhtml). I think `stage0` is worth consideration too, but it is bit more debatable. These libraries have a reconciler atleast but they apply "bindings" by using DOM operations in the open. Worth atleast taking a look.

aminya:
This benchmark should not force any programming style, it should just be a framework for exercising different libraries that perform certain functionality.

If you want to consider the "effort to write the code", there are better ways to measure that. You can add "lines of code" of the implementation as a measure that shows how much code a developer will need to write to do the included functionalities.

ryansolid:
I think lines of code are poor metric for effort. I can write very concise code that is hard to reason about yet is very performant. Or I can write Elm which is easy to write but has a disproportional amount of LoCs. That being said people can just look at the implementation and make their own judgement call assuming they do equivalent things.

The problem with non-data driven approaches is they become ultimately meaningless from a comparison perspective outside serving as a guidepost for optimal performance. The scenario presented here is an approximation of a category of the solution but shouldn't represent a specific one. Let me explain why this matters. I think there are 3 factors at play.

Having knowledge of the specific problem allows certain types of optimizations that would make no sense for a general solution. However, restricting these lower-level libraries makes no sense as there is no reason jQuery should adopt React's abstraction. It's an artificial constraint.

Conversely, every library has an escape-hatch back to the VanillasJS. So an implementor could just grab a ref and wire everything up by hand. Sure it bypasses the library but it is no more effort or complex than the VanillaJS implementation. We've seen that in implementations here. It is a tempting approach to get performance. One could argue it is even idiomatic. But any library could do the same thing.

We've gotten to a place where declarative libraries are knocking on the door of vanilla performance anyway. So it's a shrinking zone, since I'm not even sure what the idiomatic jQuery implementation looks like anymore. How Vanilla is it?

[...]

PR #799, opened 2020-09-27
[do not merge] dirty selection model for angular and react

krausest:
This is currently not intended to be actually merged.
This implements the dirty model (selected is part of the row model) for discussion in #772


frameworks/keyed/angular-ng-model-hack/src/app/app.component.ts
```diff
@@ -0,0 +1,88 @@
...
+        return Math.round(Math.random() * 1000) % max;
+    }
+
+    itemById(index: number, item: Data) {
+        return item.id;
+    }
+
+    select(item: Data, event: Event) {
+        event.preventDefault();
+        for (let i=0;i<this.data.length;i++) {
```
whittet, line 46: If you check if each row === selected, like the main branch does, then changing the value of selected will cause all rows to need to be checked by change detection. When using @for and track by item.id, then changing the id of the two rows that you are changing here should be a 500 times faster solution. For more detail read on. Best wishes and very impressive benchmark. Although this one number appears wrong for angular.

Each iteration of the loop is checking length, but it can be set once.
Only one item in the list has clazz = 'danger', so looping all items is not the optimal solution. 
Simply store the index of the selected item in a class property; use this value to find one row to set to ''.
  this.selectedRowIndex = this.data.indexOf(item);

PR #804, opened 2020-10-07
Rename yarte_wasm_app to delorean and update

botika:
I'm doing the native version and I have separated it from `yarte`. I have updated to be safe with future implementations of wasm threads.


krausest:
Thanks - results are updated: https://krausest.github.io/js-framework-benchmark/current.html

PR #805, opened 2020-10-12
Update LS-Element version

lsegurado:
With this update the performance should be better. Also i fixed frameworkVersionFromPackage field on package.json.

Cheers!


krausest:
Thanks. Here are the results:

PR #806, opened 2020-10-13
Add alpine keyed

ryansolid:
Given it's popularity over the past year I figured this was a good inclusion. https://github.com/alpinejs/alpine

Unfortunately the runtime isn't tree-shakeable (built primarily for CDN) so it doesn't tell it's size story as well more optimized build processes I think it's fair.

I've been having some issues running the tests locally though and I want to know if you experience the same. It seems that after the first click things stop tracking. It's strange since the library is pretty simplistic API-wise so not sure about the cause of this behavior.

I will make an issue on their repo too to see if they have any ideas. But the library looks really straight forward to use.

### DIFF
--- a/frameworks/keyed/alpine/index.html
+++ b/frameworks/keyed/alpine/index.html
@@ -1,0 +1,111 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Alpine-keyed</title>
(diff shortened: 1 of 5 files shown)


krausest:
Thanks. Here are the results:

This implementation uncovered an issue in the benchmark driver: https://github.com/krausest/js-framework-benchmark/commit/09fec7d33bf27060a697e9521b86d72d8b8b1a59

PR #807, opened 2020-10-13
Update to fntags 0.3.2

narcolepticsnowman:
This ran a little faster than the current in the bench on my machine...

I do believe that this change fixes the #722 issue of not being data driven, but I'm not sure how to remove the label... Please correct me if I'm wrong about that.

I also removed the node cloning stuff. It's a little faster but it's so ugly I decided to take the hit :D


frameworks/keyed/fntags/src/Main.js
```diff
@@ -29,50 +29,58 @@ const Button = ( id, title, onclick ) =>
...
-    const update = () => {
-        tr.className = item().selected ? 'danger' : ''
-        label.innerText = item().label
-        id.innerText = item().id
-    }
+        if( currentLabel != item().label ) {
+            label.innerText = item().label
+            currentLabel = item().label
+        }
+    } )
```
ryansolid, line 81: The implementation definitely looks cleaner. But this is the part that is going to draw heat. It is data-driven in the sense that you are subscribing to the model. But I guess what I'd be wondering is does the HyperScript not have a way to subscribe to these changes without escaping into this imperative chunk of direct DOM manipulation?

I looked at the library to see if I could find a way of representing what I mean but I think the answer is no.  Like a few other libraries this falls into a place where it is intended use. It's a tricky one since in one sense it is what any reactive library does under the hood and could do in the open. Writing it this way leaves it easy to optimize for the specific benchmark. It isn't unlike what Solid and Surplus were doing previously, but it's also exactly what categorized them under 772.

narcolepticsnowman: No, it doesn't. That's essentially the point of this library, provide reactive updates in an obvious way. The current voodoo magic methods are so non-obvious that it makes the learning curve more like a learning cliff for a lot of these frameworks.

IMO, it looks like a duck, it quacks like a duck, it's a duck. 

All I really wanna know is if it's faster or not.

narcolepticsnowman: I'd also like to do a marketing flip on this statement "leaves it easy to optimize for the specific benchmark" and change it to more accurately (and less pessimistically) say "leaves it easy to optimize for any specific use case".

krausest:
Thanks. Here are the results:

I think issue 772 is still valid for the implementation even though it's implemented as intended.

PR #808, opened 2020-10-14
Add BDC

bwhmather:
This adds keyed and non-keyed implementations for the [BDC](https://github.com/bwhmather/bdc-js) rendering library.

BDC should be an interesting library to include as it presents a vdom style interface but diffs directly against the DOM.
As would be expected, it is quite significantly slower than most of the more traditional vdom libraries in benchmarks that only update a bit of the DOM but slightly faster when everything needs to change.  Due to the small size of the library and and implementation it fairs rather well in the lighthouse and memory tests.

The code is fairly idiomatic, not least because the library doesn't really provide any escape hatches to enable cheating.  Hope you like it.

Screen Shot 2020-10-14 at 21 24 21

### DIFF
--- a/frameworks/keyed/bdc/index.html
+++ b/frameworks/keyed/bdc/index.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8"/>
+  <title>bdc</title>
+  <link href="/css/currentStyle.css" rel="stylesheet"/>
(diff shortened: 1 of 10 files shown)


krausest:
Thanks. Here are the resutls:

bwhmather:
Fourth slowest in the big table. :)

Thank you for setting everything up.  Adding an implementation has been a lot of fun.

PR #809, opened 2020-10-19
update to fntags 0.3.5

narcolepticsnowman:
Added functionality to perform select via events.

Row mapping now uses "bindAs" function to automatically subscribe. 

These two things possibly solve the 722 issue, but I am probably still misunderstanding that issue.


krausest:
Thanks! [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated.
#772 is IMO still correct since select and deselect set directly a class name on a DOM node.

narcolepticsnowman:
@krausest Ah, ok. That makes sense. I'll see if I can think up a way  to do that and not be super slow and terrible. :D Thanks again for maintaining this project.

PR #797, opened 2020-09-26
Trying blazor-wasm .Net 5

aminya:
I have heard that a lot of work has been done to improve the performance of Blazor. I want to see if that makes a difference in this benchmark.

Requires sdk-installer from: https://github.com/dotnet/core/blob/master/release-notes/5.0/preview/5.0.0-rc.2.md

PR #810, opened 2020-10-19
Fix blazor wasm version

aminya:
I forgot to change this version in #797. It only affects the title of the result page.


aminya:
@krausest This is a patch for #797

krausest:
Thanks.[Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated (despite it's only the label for the version that has changed).

PR #811, opened 2020-10-20
Dojo: Fix key issue for swapping nodes

agubler:


krausest:
Thanks - good to see that swap nodes is correctly keyed now. [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated.

PR #812, opened 2020-10-23
Upgrade dependencies on react contributions

josepot:
Hi and thanks a lot for creating and maintaining this project!

I'm about to make a contribution for `react-rxjs` and I thought that maybe it would be a good idea to first upgrade all react-based contributions, so that they all use the latest version of both React and their own library.

Regarding `react-mobx` though, I have been able to upgrade React to version 17.0.1 without any issues. However, I haven't been able to upgrade `mobx` and `mobx-react`, because the latest versions that are not backwards compatible with the current implementation.

If you rather handle these upgrades yourself, no worries, I totally get it. If that's the case, when I send the `react-rxjs` contribution I will be using the same version of React 16 that all the other ones are using and I will upgrade when the other ones upgrade. Thanks!

### DIFF
--- a/frameworks/keyed/react-easy-state/package-lock.json
+++ b/frameworks/keyed/react-easy-state/package-lock.json
@@ -13,35 +13,43 @@
 				"@babel/highlight": "^7.10.4"
 			}
 		},
+		"@babel/compat-data": {
+			"version": "7.12.1",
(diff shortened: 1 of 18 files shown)


krausest:
Thanks a lot! [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated.

PR #815, opened 2020-10-27
Update to fntags 0.4.0

narcolepticsnowman:
Minor performance improvements, and minor memory usage improvements.

Fixed the 772 issue by replacing the select events that used direct dom manipulation with a bindSelectAttr function that allows to do data driven binding to attributes. If I am still misunderstanding this, please correct me.


ryansolid:
Yeah this does it I think!

I mean basically the only difference between what you are doing now and what I do in Solid's HyperScript implementation or the way Sinuous works is that our `h` function detects based on input (ie.. if you pass a function) whether to call the helpers instead of just directly writing to the DOM without any further end user intervention

Although that is just syntax, and doesn't change functionally how it works. In my opinion I think #772 can be removed unless I missed something.

narcolepticsnowman:
@ryansolid Thanks for the input on this, I really like the way it ended up here and wouldn't have pushed it if it weren't for this awesome project and you fine people's input.

krausest:
Thanks. Results are updated:

PR #818, opened 2020-10-30
update Mikado to latest + fix issues

ts-thomas:
Passed Issues Keyed:

- [x] 634 The HTML structure for the implementation is not fully correct.
- [x] 694 Keyed implementations must move the DOM nodes for swap rows
- [x] 772 Implementation uses manual DOM manipulations
- [x] 796 Implementation uses explicit requestAnimationFrame calls
- [x] 800 View state on the model
- [x] 801 Implementation uses manual event delegation

Passed Issues Non-Keyed:

- [x] 634 The HTML structure for the implementation is not fully correct.
- [x] 694 Keyed implementations must move the DOM nodes for swap rows
- [x] 772 Implementation uses manual DOM manipulations
- [x] 796 Implementation uses explicit requestAnimationFrame calls
- [x] 800 View state on the model
- [x] 801 Implementation uses manual event delegation

### DIFF
--- a/frameworks/keyed/mikado/src/main.js
+++ b/frameworks/keyed/mikado/src/main.js
@@ -1,14 +1,13 @@
 import Mikado from "../node_modules/mikado/src/mikado.js";
 import Array from "../node_modules/mikado/src/array.js";
-import app from "./template/app.es6.js";
-import item from "./template/item.es6.js";
-import { buildData } from "./data.js";
+import tpl_app from "./template/app.es6.js";
+import tpl_item from "./template/item.es6.js";
+import buildData from "./data.js";
 
-Mikado.once(document.getElementById("main"), app);
+Mikado.once(document.getElementById("main"), tpl_app);
 
-let selected = 0;
 const store = new Array();
-const view = new Mikado(document.getElementById("tbody"), item, {
+const view = new Mikado(document.getElementById("tbody"), tpl_item, {
     "reuse": false, "store": store
 })
 .route("run", () => store.set(buildData(1000)))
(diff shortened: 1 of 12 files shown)


frameworks/keyed/mikado/src/main.js
```diff
@@ -25,8 +24,4 @@ const view = new Mikado(document.getElementById("tbody"), item, {
     store[1] = tmp;
 })
 .route("remove", target => store.splice(view.index(target), 1))
-.route("select", target => {
-    store[selected]["selected"] = "";
-    store[selected = view.index(target)]["selected"] = "danger";
-})
-.listen("click");
+.route("select", target => view.refresh(view.selected = view.index(target)));
```
ryansolid, line 27: Implementation looks pretty clean just was trying to understand this one. It's only update where action on the view happens instead of the store. What does that mean? Where is selected state stored?

ts-thomas: It is stored to the view instance, so different views with the same data can handle its own select state.

krausest:
Nice - results are updated:

frameworks/keyed/mikado/src/main.js
ryansolid: Gotcha so it's a bit more like a React having it's own `useState` in the child component. Love to dig in to understand how rows get unselected at some point but cool stuff.

ryansolid:
And I gotta say that partial update performance is impressive. Might have been the run but otherwise it looks like it is clearly faster than the vanilla implementation which is pretty impressive (and probably deserves some exploration to see if we can improve Vanilla's version).

PR #821, opened 2020-11-01
bugfix: ember issue #800

lifeart:
related to https://github.com/krausest/js-framework-benchmark/pull/814


krausest:
Thanks. Ember and glimmer take a bad hit from removing #800 

much more than react or angular https://github.com/krausest/js-framework-benchmark/issues/772#issuecomment-699624765

PR #822, opened 2020-11-02
Upgrade to fntags 0.4.7

narcolepticsnowman:
This fixes a memory leak during the create/replace. 

May be a little faster too.


krausest:

PR #823, opened 2020-11-06
add react-rxjs

josepot:
Hi @krausest,

Would it be possible to add react-rxjs?

Thanks,

### DIFF
--- a/frameworks/keyed/react-rxjs/index.html
+++ b/frameworks/keyed/react-rxjs/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <title>React + React-RxJS</title>
+  <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+  <div id='main'></div>
+  <script src='dist/main.js'></script>
+</body>
+</html>
--- a/frameworks/keyed/react-rxjs/package-lock.json
+++ b/frameworks/keyed/react-rxjs/package-lock.json
@@ -1,0 +1,5422 @@
+{
+  "name": "js-framework-benchmark-react-rxjs",
+  "version": "1.1.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.10.4",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
+      "integrity": "sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "^7.10.4"
+      }
+    },
+    "@babel/compat-data": {
+      "version": "7.12.1",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.12.1.tgz",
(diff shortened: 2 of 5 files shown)


krausest:
Thanks, quite a good react version! How come this version beats the react-hooks and react implementation?

ryansolid:
I've wondered this too.. React Redux Hooks also has this characteristic of scoring better on the selection test. I can think of how it would be possible but I'm not sure I see it here. So definitely interesting.

josepot:
Hi @krausest and @ryansolid 

Sorry for the late response. I wanted to give you this explanation a few days ago, and then I forgot.

> How come this version beats the react-hooks and react implementation?

First of I want to say that I'm also a bit surprised. However, after having had a look at the react-hooks implementation, I think that I understand why this is happening.

I think that the key reason is that thanks to react-rxjs we can separate the producer from the consumer. 

What I mean by this is that in the React implementation both `state` and `dispatch` are created together, but for instance the `Jumbotron` component only cares about `dispatch`, right? But now every time that the state changes we will create a new `Jumbotron` element and react will have to compare it with the previous one in order to keep the old reference... That comparison is beyond-cheap, of course, but what's even cheaper (and cleaner) is not having to worry about it, like in the react-rxjs implementation 🙂 .

Although, of course, the `Jumbotron` itself won't make a difference. However, the fact that we can define the "producer" functions outside of React (`onAdd`, `onUpdate`, `onClear`, etc) means that we won't need to re-create them after every update. And maybe that makes a small difference? Because if we look at the `Row` component of the `react-hooks` implementation we can see [`useCallback` being used twice](https://github.com/krausest/js-framework-benchmark/blob/18d79b286b19f687d72f78ee6ce76c5e7df41463/frameworks/keyed/react-hooks/src/main.jsx#L59-L60) in order to try to keep those functions stable across the different updates. If you ask me, I think that's a mistake, I'm pretty sure that's detrimental for performance. At the end of the day, the only time that those functions would be re-created will be when the `selected` property changes, which is not very often... But even when that happens, what's the point on keeping the functions stable? What component is benefitting from those functions remaining stable? I'm 99% sure that those usages of `useCallback` are bogus and that the performance of the react-hooks implementation would improve if they got removed... 🤷 

> React Redux Hooks also has this characteristic of scoring better on the selection test. I can think of how it would be possible but I'm not sure I see it here. So definitely interesting.

The only thing that IMO can make the row-selection a little bit faster on the react-rxjs implementation (when compared to the react-hooks implementation) is the fact that the shallow comparison of each row that react performs due to `React.memo` has to compare 3 properties with the react-hooks implementation vs 2 properties in the react-rxjs implementation (because with react-rxjs we don't have to pass `dispatch`). That and the fact that maybe is slightly faster to create a react element with one property less? I don't know.

It's worth pointing out that I could have optimized the react-rxjs implementation so that it scored better on the row-selection implementation, by means of having one Observable per row and only triggering an update for the row that has to be updated... However, that would have come at the cost of making the render function of each row slower, so that optimization would have penalized all the other metrics and consumed a lot more memory, so it was not worth it.

I know why the react-redux-hooks performs so well in this benchmark. However, I'm of the opinion that there should be one of those warnings for react-redux-hooks implementation, because they are kinda "cheating" in the sense that:

1) Using `react-redux` hooks can be pretty problematic due to [Stale Props and "Zombie Children" problems.](https://react-redux.js.org/api/hooks#stale-props-and-zombie-children)
2) The current implementation of `useSelector` of react-hooks is using references in a way that breaks the react contract of always returning the same given the same react-state, react-properties and react-context.

In other words, IMO the react-redux hooks implementation is sacrificing correctness in favor of performance.

React-RxJS, on the other hand, tries its best to always behave correctly... Even under [Concurrent Mode](https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-mode#results).

PR #824, opened 2020-11-08
Ember 800 improvements

lifeart:
related https://github.com/krausest/js-framework-benchmark/pull/821


krausest:
Thanks. New results are on the left and old results on the right side.
Indeed faster! (But if you want my opinion it's quite a lot of user code for a basic functionality but that's just my 5 cents).

PR #825, opened 2020-11-10
Add Callbag-JSX

loreanvictor:


krausest:
Thanks! Results are on the left side (the right side shows old results which is pretty useless for callbag-jsx :smile: )

loreanvictor:
Thank you for the elegant benchmark tool (like this is a pretty handy tool for analyzing behaviors of such libraries and working on them)! And yep seems about right (its strange that sometimes `callbag-jsx-list` can overtake `callbag-jsx`, though it has also happened on local testing on and off).

PR #827, opened 2020-11-12
Update marionette implementations

paulfalgout:
For version 5 Marionette is making backbone an optional dependency.

This PR updates the straight marionette option to avoid issue 772 as removing the Bb dependency allows for updating the className as part of a templating solution.

Removing Bb as dependency for views also means that jquery is essentially out of the picture by default, so marionette-backbone represents the mirgration path for current projects that have a deeper dependency on backbone as marionette-jquery did for v4.


krausest:
Thanks! Updated results are on the left side and old results on the right side.

PR #828, opened 2020-11-12
add react redux + immutable

TheHaff:
Was just curious to see how redux + immutable.js performs


krausest:
Thanks! Pretty interesting to see the impact for immutable, now I think we should add immer.js in this mix :smile: 
Maybe I'll remove this implementation at some point in time again when the redux version evolves to limit the number of react versions somewhat. For the time being I think it's an interesting variation.
Updated results are on the left side.

Issue #826, opened 2020-11-11
Discrepancy between Chrome's performance measurements and benchmark tests


andyhasit:
I'm comparing my framework (not ready, not submitted) against Svelte and getting very different stories from Chrome's devtools (just the record button) compared to the benchmark tests. 

For example, running "Add 10,000 rows" I get:

| Framework        | DevTools           |  Benchmark  |
| -------------------- | ----------------------| ------------------|
| Svelte               |      avg 8500 ms |      2,196 ms |
| Mine                 |      avg 3700 ms |      2,226 ms |

I've run all of these dozens of times now. 

I am not concerned about the difference between Svelte and my framework, or between Chrome+devtools and Chromium. What concerns me is how it is possible for devtools to clock Svelte as being twice as slow as my framework, when the benchmark clearly puts it ahead.

Surely devtools is not _that_ unreliable?

Screenshot from local results table:

![image](https://user-images.githubusercontent.com/2077696/98866573-1060ac80-2465-11eb-97c9-f7e4f1170540.png)

Screenshot of my framework add 10,000 rows:

![image](https://user-images.githubusercontent.com/2077696/98866400-c4ae0300-2464-11eb-8354-4fadf6ac6126.png)

Screenshot of Svelte add 10,000 rows:

![image](https://user-images.githubusercontent.com/2077696/98866446-db545a00-2464-11eb-8fa2-b3a7c0f80056.png)

leeoniya:
have you tried `Disable JavaScript samples`?

![image](https://user-images.githubusercontent.com/43234/98872608-7b49bd80-243c-11eb-83d2-231a189d4f34.png)

andyhasit:
@leeoniya that does speed up the times by a bit:

![image](https://user-images.githubusercontent.com/2077696/98874744-c1217880-2472-11eb-9b40-f6d55609d83b.png)

But I am still seeing Svelte as roughly twice as slow on Chrome. 

It hadn't occurred to me to manually test in Chromium, so I did that, and here are roughly representative samples of what I'm getting (with samples disabled):

Svelte on Chromium:

![image](https://user-images.githubusercontent.com/2077696/98875322-e793e380-2473-11eb-81a6-475f489bd99b.png)

My framework on Chromium:

![image](https://user-images.githubusercontent.com/2077696/98875380-085c3900-2474-11eb-9440-301dbcb0aee6.png)

[...]

PR #830, opened 2020-11-15
Added RedRunner 0.8.1

andyhasit:
Hi Stefan,

Here's is a PR for you to verify #826 

I forgot that I was experimenting with a different strategy for rows, which means this implementation is not keyed. I'm going to leave it as is to see if you observe similar figures to me.

I don't think this detracts from figuring out the above issue, but if you want to make this implementation keyed, simply change line 10 in **src/views.js** from this:

`<tbody id="tbody" :items="*|.rows|Row"></tbody>`

To this:

`<tbody id="tbody" :items="*|.rows|Row|item.id"></tbody>`

And it will pass the check.

### DIFF
--- a/frameworks/keyed/redrunner/babel.config.json
+++ b/frameworks/keyed/redrunner/babel.config.json
@@ -1,0 +1,7 @@
+{
+  "presets": ["@babel/preset-env"],
+  "plugins": [
+      "babel-plugin-redrunner",
+      "@babel/plugin-proposal-class-properties"
+  ]
+}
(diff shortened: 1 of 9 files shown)


krausest:
Are you going to update this PR or do you prefer to submit a new PR when redrunner is ready?

andyhasit:
I will update, but won't have time to do anything until after the weekend. Feel free to keep open or close and I'll open a new, as you please.

andyhasit:
I'll just close this for now as I've made some significant changes to RedRunner, and need a bit of time to test them in the wild.

PR #831, opened 2020-11-18
remove bogus useCallback from react-hooks implementation

josepot:
As I mentioned in [this comment](https://github.com/krausest/js-framework-benchmark/pull/823#issuecomment-729622921): I'm quite confident that using `useCallback` inside that `Row` component can only be detrimental for performance.


krausest:
Thanks! Here are the results - old results are left, new to the right.

PR #832, opened 2020-11-18
Vue.next updated to 3.0.2

husayt:
Existing tests were based beta version and since then Vue 3 has been released, with many improvements. This patch upgrades yo latest Vue3


krausest:
Thanks. Results were [updated](https://krausest.github.io/js-framework-benchmark/current.html).

Issue #829, opened 2020-11-14
ws prefers gzipped bundle if there is one


andyhasit:
I occasionally gzip the production bundle of the framework I am developing, as well as those of several other frameworks, to see their sizes gzipped:

    cd dist
    gzip -fk main.js
    ls -lh

That command leaves two files in the dist directory:

    dist/main.js
    dist/main.js.gz

I don't usually bother deleting the gzipped file as there is no need.

But I just discovered that the webserver (`ws`) chooses `main.js.gz` over `main.js` if it exists.

Unless this is a very recent change, this means:
 
1. Every time I updated the repo, installed the new versions of a frameworks and run tests, I have in fact been re-testing whatever version I had at the time I last gzipped the bundle for that framework, which could be months ago.
2. Every time I changed my own code and re-run tests (or even just loaded the page in the browser) I have actually been re-testing whatever went into the last bundle I gzipped, which could be anything between a few minutes or days ago.

Given a lot of people interested in performance are also interested in bundle sizes, and might be tempted to do the same, I think we should change this behaviour. If we can't, then we should warn about it as it invalidates your local tests, and given you expect variance anyway, it's very hard to spot.

I couldn't find anything about gzip in [here](https://github.com/websockets/ws/search?q=gz).

andyhasit:
I had a big dig over at [websockets/ws](https://github.com/websockets/ws) and couldn't determine where or how it decides to serve a `main.js.gz` instead of `main.js`, nor can I see anything which does that in this repo, which makes me think it's a dependency of [websockets/ws](https://github.com/websockets/ws) which causes this behaviour...

I know how to enable this behaviour on nginx, and I know that some node servers (like [Koa](https://github.com/koajs/static)) have a flag for this. But as far as I can tell from it's [websocket-server.js](https://github.com/websockets/ws/blob/master/lib/websocket-server.js) file, it uses the [http](https://nodejs.org/api/http.html) module for static content, and [http](https://nodejs.org/api/http.html) definitely doesn't do that by default (I tested that, partly to see if it might be Chrome headers).

I am wondering if the simpler solution would be to warn (or abort?) the benchmark if a `.gz` file is found in the `/dist` directory? This would protect us should we ever switch to a different server.

If you think this is a good solution I can do a PR for that, and would added it to the isKeyed check too. And let's hope no existing webpack actually creates a gzipped output.

[...]

PR #833, opened 2020-11-26
Gzip check

andyhasit:
I included a check for the gzipped files as per #829. 

The check returns a `FrameworkVersionInformationError` which is not really the right name for it, but it's what the function returns if it doesn't find a `package.json` file so I figure it is the correct thing to return, but perhaps it should be renamed to something which captures the purpose better, like `Framework[Implementation/Installation/Configuration]Error` or something?

I wanted to mention gzip files in the README, but got a bit carried away and made these changes:

1. Put headings to correct sizes relative to others
2. Grouped all the "how to use this tool" sections together, and all the "how to contribute a framework" together - as they were previously a bit scattered.
4. Fixed some obvious typos, and added missing `:` before some code blocks

But that's all I changed in the README. I haven't reworded anything inside the sections or removed content, except for mentioning gzipped files in two places.

PS: I highly recommend using [typora](https://typora.io/) for README files. It highlights spelling mistakes, and you can see at a glance if your heading sizes are consistent.

Here's how the headings look now:

image

How they were before:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,11 @@ The current snapshot that may not have the same quality (i.e.
 results might be for mixed browser versions, number of runs per benchmark may vary) can be seen [here](https://rawgit.com/krausest/js-framework-benchmark/master/webdriver-ts-results/table.html)
 [![Results](images/results.png?raw=true "Results")](https://krausest.github.io/js-framework-benchmark/current.html)
 
-## How to get started - building and running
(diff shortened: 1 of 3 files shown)


andyhasit:
Sorry, I did modify some content on the README, regarding the rebuilding of the index file. Please check if that is good.

PR #834, opened 2020-12-05
Add fractal implementation

iminside:
Hi @krausest )
Would it be possible to add my framework?
Thanks


krausest:
JFI: I'll take a look at it this weekend.

krausest:
Thanks. The non-keyed version was reported by the isKeyed test to be actually keyed. So I removed the non-keyed version (who needs non-keyed anyways?). 

iminside:
Ok. 
Let only the version with keys remain.
Thanks!

PR #835, opened 2020-12-06
Update Riot.js version

GianlucaGuarini:
Update to the latest Riot.js version that should be more performant than previous one


krausest:

PR #836, opened 2020-12-08
Update keyed Mimbl implementation; remove non-keyed Mimbl implementation.

mmichlin66:


krausest:
Great to see a removal of a non-keyed version!
[Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated. Looks like a nice improvement!

mmichlin66:
Hi Stefan,

Thank you for your work and for the kind words!

I am actively working on my library (after some pause) and might re-introduced the non-keyed version again – it is supposed to be so much more efficient 😊

How often do you run the tests and update the results? Do you have a schedule?

Michael Michlin
https://mmichlin66.github.io/

PR #840, opened 2020-12-27
update react-tracked v1.6.0

dai-shi:
also follows the change in #831.


krausest:

PR #841, opened 2020-12-27
add valtio

dai-shi:
https://github.com/pmndrs/valtio


krausest:
Thanks - and sorry for the long delay. Here's a screenshot of the results (left new, right old - so for you only the left is relevant ;-):

Issue #796, opened 2020-09-26
Note: Implementation uses explicit requestAnimationFrame calls


krausest:
(Severity: Code smell / cheating)
There were some discussions whether using requestAnimationFrame in client code is okay, a code smell or cheating.
Applying RAF for selected benchmark operations should be avoided and is considered cheating:
https://github.com/krausest/js-framework-benchmark/issues/166 
https://github.com/krausest/js-framework-benchmark/issues/430

Using RAF for all operations in client code should be okay (that doesn't answer the question whether is's nice) since frameworks are allowed to use RAF under the covers too.

PR #842, opened 2020-12-28
Update skruv

SahAssar:
This updates skruv to 0.1.0 and also removes the RAF #796 implementation issue.


krausest:

PR #848, opened 2021-01-03
Fix explicit use of requestAnimationFrame in BDC

bwhmather:
There was no good excuse for having this in before.


krausest:

bwhmather:
You are still much faster than this submission...
More seriously, no hurry.  Best to batch things up.  Thank you for taking any time at all to keep this little competition running.

PR #849, opened 2021-01-05
Update ScarletsFrame

StefansArya:
Hi, I have updated the framework to v0.33.1 for keyed and non-keyed.
The non-keyed version will still have state on the model #800 because I want to check how is the impact of having the state on the model for this framework on the future. Thanks!


krausest:

StefansArya:
It's okay, I just waited not long ago.
Looks like the select row was 17ms slower and now it catches my eye because it has turned yellow.

Anyway, thank you so much for maintaining this project.

PR #852, opened 2021-01-14
Use mimbl@0.6.2 for keyed and non-keyed implementations

mmichlin66:


krausest:
Thanks - [results](https://krausest.github.io/js-framework-benchmark/current.html) are update and it looks like mimbl improved a bit!

mmichlin66:
Thank you!

PR #854, opened 2021-01-22
Update seed to v0.8.0

flosse:
@MartinKavik would you mind to review this update?


MartinKavik (approved):
I don't see any problems, good job!

flosse:
> I don't see any problems, good job!

thank you!

krausest:
Thanks. [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated. But I'm afraid to say that I don't see an improvement for v0.8.0.

frameworks/non-keyed/seed/src/lib.rs
```diff
@@ -119,37 +120,41 @@ fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg>) {
...
             }
-        },
+        }
         Msg::Select(id) => model.selected = Some(id),
         Msg::Remove(id) => {
             model.rows.remove(Row::position(id, &model.rows));
-        },
+        }
     }
     orders.force_render_now();
```
flosse, line 154: @MartinKavik Could you explain why it is required here to force the rendering?

flosse:
> Thanks. [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated.

Thanks a lot!

> But I'm afraid to say that I don't see an improvement for v0.8.0.

That's a pity but hopefully a motivation to make seed faster ;-)

MartinKavik:
@flosse @krausest 

While I was trying to improve Seed performance, I fell into the rabbit hole and the result is the experimental framework Zoon (a part of [MoonZoon](https://github.com/MoonZoon/MoonZoon); it's in the design phase).
In Zoon, I try to integrate principles / features like: reactivity, light-weight VDOM, `no_std`, incremental computing, stateful functions, Rust/Cargo `features`, etc. In the case of success (fast, small bundle size, good API), the Zoon will be used as an inspiration for Seed improvements.

Back to Seed. Seed 0.8.0 has some tools to improve performance - especially `Node::NoChange` - but you should use it only when the slow rendering is reducing UX. So I'm not sure if we want to try to integrate it to the benchmarked example. Without `Node::NoChange`s or `el_key`s, the entire `view` is always recomputed and diffed and it's a relatively slow process.

The bundle size - I don't know the exact reason, but in most cases the reason is either a too large macro definition or a new / updated dependency. So we would need to check/refactor macros or try to hide dependencies behind features or make sure linked dependencies don't use a big transitive dependency like the `url` crate.

frameworks/non-keyed/seed/src/lib.rs
MartinKavik: If I remember correctly it was a way to skip waiting for the browser render loop (aka `requestAnimationFrame`). Yew doesn't / didn't use it at all so I was thinking it makes the results more comparable.
But I'm not sure if we need / want it. Feel free to remove it.

PR #855, opened 2021-01-24
docs(README): suggest running `ci` instead of `install`

meeroslav:
running `npm install` potentially overrides the `package-lock` file causing unplanned changes.
`npm ci` instead leverages the `package-lock` file to create exactly the same node modules setup.


krausest:
I finally switched to npm ci for rebuilding all frameworks and created a new task docker-rebuild-ci for rebuilding a single framework. I'll check how it works out.

PR #856, opened 2021-02-04
Update ScarletsFrame

StefansArya:
Hi!
I have updated the framework to v0.34.6 and changed the bundler from Webpack to Rollup.

Because "scarletsframe" name was overlapping it's own box on the [table](https://krausest.github.io/js-framework-benchmark/current.html), I temporary renamed it into "scarlets-frame".


krausest:
Thanks. Results are updated (but they're almost unchanged).
I think we can consider this benchmark to be a thing when we see renames to "re-act" and "an-gular". vue might be short enough to fit in the table cell :smile:

StefansArya:
Thanks! Optimizing the performance is getting harder now 😅. 
But now it got reduced the `total KB weight` and optimized the memory consumption (but just 0.1MB 😢).

---
About the name on the table, I have tried to use `word-break: break-word`, but it makes `marionette-backbone` become like this.

OilfAVzPSu
And using `word-break: break-all` make the table head even worse.

So I decided to split the phrase of `scarletsframe` only on this benchmark, and I think it looks more better than got sliced..
I'm sorry if it make a little confusion 😅

PR #857, opened 2021-02-18
add s2

gr0uch:
https://github.com/gr0uch/s2


gr0uch:
this is my second contribution to the benchmarks :)

ryansolid:
This one falls under #800. But it is also non-keyed so maybe no one cares.

gr0uch:
> This one falls under #800. But it is also non-keyed so maybe no one cares.

Seems that is correct, view state is on the model. But I am actually not sure if this qualifies as keyed or non-keyed, since it internally uses a WeakMap to associate a model to a DOM node, so actually one could mess with the DOM externally and this would continue to work as expected.

ryansolid:
That sounds like keyed then. If the data stays attached to the same DOM node for its lifetime. There is an isKeyed test.

gr0uch:
@ryansolid thanks for the advice!

I did run the isKeyed test with `npm run isKeyed -- --framework s2` and got:
```
s2-vinvalid (no package-lock)-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as keyed in the results
```
so it is indeed keyed.

krausest:
Results are updated:

Nice to see that it turned out to be keyed. Performance looks pretty good!

PR #859, opened 2021-03-03
feat: introduced rematch framework

semoal:
Hi! 👋 ,
I'm Sergio Moreno, lead maintainer of [Rematch](https://rematchjs.org), I've been working on a v2.0.0 version with a lot of new improvements and wanted to checked that the difference with Redux in terms of performance is ridiculous but the maintainability and the steep curve was reduced.

I've correctly run the suite on my local machine:
Screenshot 2021-03-03 at 12 41 32

Feel free rquest any changes =)


krausest:

Issue #801, opened 2020-09-28
Note: Implementation uses explicit event delegation


krausest:
Using explicit event delegation is often discussed and (almost as often) considered a hack to boot performance.
Please note that frameworks are allowed to use any event delegation model they choose to, but it's considered fair play to stay clear of user code that uses explicit event delegation.

brainkim:
Any progress on this issue? I’m realizing that assigning DOM `onclick` properties is eating a significant chunk of time on tests like rowlots. Is there a sliding scale? Am I allowed to read the row id off the DOM via ev.target? Will I be penalized if I have one event listener on the table vs one for each row? My framework doesn’t implement any sort of framework provided event delegation so I would argue it is idiomatic to use some kind of event delegation mechanism when you have a parent with many children.

Also, does anyone have any microbenchmarks for whether `onevent` properties are faster or slower than addEventListener calls? It would be a shame to have to switch for performance, I think virtual DOM libraries are actually the perfect use-case for `onevent` props.

ryansolid:
We don't seem to be marking libraries for this currently. So I don't believe there is any repercussion to doing so? Atleast today. #772 is more about manipulating the selection class directly.

I've argued against this being penalized since it is the defacto way to solve events in a large table. Either the framework does it, or you do it. Some libraries actually have this in the documentation as being the way to do it. Once you get to a certain point on the table everyone is using event delegation. It isn't like it is fundamentally changing the nature of the test.

I think this is sort of gray area in that it makes the implementation look less clean so it's discouraged. It breaks the abstraction wall. But it's hard to penalize someone for something that almost every other library is doing. I do see the argument that if we are critical on libraries using imperative escape hatches to directly manipulate DOM elements in ways the library already has means to but are being avoided purely for performance reasons, that being fine reading from them is a very subtle distinction to make when the library already has a way to attach events. To be fair explicit event delegation still generally uses the libraries event attaching mechanism, it just involves manual traversal of DOM trees.

But I think we need to decide if this is something worth penalizing on. And if not, leave to the implementor. I once made an event delegated version of Svelte but I wasn't comfortable merging it without Svelte community or Rich's blessing. I never got it so I retracted my PR.

brainkim:
[...]

PR #861, opened 2021-03-04
updated to remove 772 error and pass the iskeyed test

yysun:


krausest:
Thanks! I think it's a great step for AppRun to get rid of #772. 
But I think I must mark it for using explicit event delegation #801 in the click function (which is a much minor issue than #772) . 
Would you concur?

yysun:
I agree with you and thank you very much.

krausest:

yysun:
Awesome, thanks!

PR #862, opened 2021-03-07
Add Jotai implementation

Thisen:
This adds an implementation for the library [Jotai](https://github.com/pmndrs/jotai).


krausest:

PR #863, opened 2021-03-16
Add react-tagged-state framework

oleggrishechkin:


krausest:
Thanks, [results](https://krausest.github.io/js-framework-benchmark/current.html) are updated. Here's a screenshot from the latest run including some react variants (and older version for the other updated frameworks).

PR #864, opened 2021-03-17
Update MobX JSX, VueRX JSX, and KO JSX

ryansolid:
This PR adds Solid's optimization around selection to the other reactive libraries based on my DOM Expressions renderer. While it isn't quite the same mechanism I'm hoping it restores these libraries to some of their former glory. There is a steeper memory cost here but I'm hoping performance is worth it.


krausest:
Thanks, and yes - performance looks pretty solid :smiley: 
[Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated. Here's a screenshot from the latest run.

PR #865, opened 2021-03-17
Add 1more implementation

Freak613:
Retire `stage0` and `domc` and replace it with something less awkward, which is [1more](https://github.com/Freak613/1more)
Should not mark any warning flags.


krausest:
Nice! [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated. Here's a screenshot from the latest run.

Freak613:
Thank you!
Last Chrome update brought some unexpected dips, but still pretty good for such boring test implementation.
I think just by looking at the code, nobody could expect that it would work the way it is. That is fun part)

PR #866, opened 2021-03-22
chore: update valtio

dai-shi:
to v1.0.0.


krausest:
Thanks. Here's a screenshot from the result (left new, right old). Looks a little bit slower, but could just be a sampling error:

PR #868, opened 2021-03-22
Doo html ver 0.20.8

hman61:
Initial submission to js-framework-benchmark


krausest:
Thanks. It took some work to adjust the benchmark driver to supporting shadow dom roots with arbitrary names and having the buttons outside of the shadow dom, but now the benchmark driver is flexible enough for it.

I have one strange observation that I don't understand. The isKeyed check reports that swap rows is not keyed. But when I test it manually and e.g. attach background-colors to the TR or TD that is swapped it behaved keyed and the color also changes the position.
All other correct keyed frameworks create no new tr elements for the swap operation. React adds and removes 997 rows, 1more adds and removes 2 rows. The isKeyed check reports 2 added, 2 removed and 2 new nodes for doohtml. Can you make sense of it? Are you cloning the rows for swap?

Anyways - here's a screenshot from the result (left new, right old):

hman61:
Thank you for merging it.

I will make adjustments to the swap row method.

Also, I have not been able to stand it up in Docker locally so that I can
run the tests.
I will send you the docker errors that I'm getting later this week when I
have more time to work on it.

Henrik Javen
310-562-3076

hman61:
I sent up a new version of DooHTML - v0.30.4 for you to pull

Do you have a place where I could download a fully working Docker Image?

I'm still not able to get it to work in Docker.

I would like to be able to be able to test the benchmarks before I submit.

Thanks

Henrik Javen

krausest:
I'm afraid there's no published docker image yet. 
Did you take the steps mentioned [here](https://github.com/krausest/js-framework-benchmark/wiki/Build-all-frameworks-with-docker)? I rebuild the whole docker container and ran all tests on Fedora 34 yesterday and it worked almost perfectly (actually building literaljs currently breaks when removing the package-lock.json).
What error message do you get and what host system are you using?

hman61:
My last try was on Docker Desktop for Windows, but is using a Linux Sub
System so it should be similar.

Docker image started fine but it could not find a  /build directory.

I looked inside the container and the build directory is missing.

I will try one more time, to make sure I didn't miss a step and send you
the exact logs

krausest:
Thanks. You should start the container via [`npm run docker-start`](https://github.com/krausest/js-framework-benchmark/wiki/Build-all-frameworks-with-docker). This command binds the js-framework-benchmark volume to the /build folder. If the volume doesn't exist it should be created automatically.

krausest:
There was indeed a file that I forgot to check in (which I added now)

Issue #838, opened 2020-12-23
forceUpdate in react not a valid comparison?


greggman:
I noticed in react there's this code

```
swapRows = () => {
    const data = this.state.data;
    if (data.length > 998) {
      let temp = data[1];
      data[1] = data[998];
      data[998] = temp;
    }
    this.forceUpdate();
  }
```

Why is that not the same as react hooks? as in

```
  swapRows = () => {
    const data = this.state.data;
    if (data.length > 998) {
      this.setState({ data: [data[0], data[998], ...data.slice(2, 998), data[1], data[999]] })
    }
  }
```

Same for `update` and `remove`

greggman:
Trying it the results seems to show not using `forceUpdate` is better (except for remove)

<img width="284" alt="Screen Shot 2020-12-23 at 13 28 13" src="https://user-images.githubusercontent.com/234804/102959602-50668500-4523-11eb-8695-4e67be5303da.png">

here's the updated code

```
  update = () => {
    const data = this.state.data;
    const newData = data.slice(0);
    for (let i = 0; i < newData.length; i += 10) {
      const r = newData[i];
      newData[i] = { id: r.id, label: r.label + " !!!" };
    }
    this.setState({data: newData});
  }

  remove = (item) => {
    const data = this.state.data;
    const idx = data.indexOf(item);
    this.setState({ data: [...data.slice(0, idx), ...data.slice(idx + 1)]});
  }

  swapRows = () => {
    const data = this.state.data;
    if (data.length > 998) {
      this.setState({ data: [data[0], data[998], ...data.slice(2, 998), data[1], data[999]] });
    }
  }
```

ryansolid:
That's awesome. I think that implementation has been that way for years.  I'm positive that for the longest `forceUpdate` was faster for partial updates but really great to see that it isn't since it feels hacky. Other implementations like Redux and Hooks removed that hackiness. To be fair there are VDOM libraries that benefit from that sort of approach but so much better if we can get rid of it.

krausest:
Yes, it was changed in Oct 2018. 
Would anyone bother if we just removed the class based implementation and just go with the hooks implementation (now that it's even faster)?

ryansolid:
It's a fair question. I still cite this often enough. Out in the community anti-hooks sentiment likes to point to performance and I've needed to answer that question, even wrote an article on it. So I kind of like leaving it around. But maybe we should just make the hooks the official one.. and call out classes instead.

[...]

PR #869, opened 2021-03-23
Update react and react-hooks

oleggrishechkin:
it PR slightly increase react-hooks performance (useCallback is missing in current implementation), and rewrite react with classes - now react-hooks and react implementations are same - first use memo, useReducer and useCallback, second use PureComponent, state and methods (one dispatch instead of one method per action)

react-hooks performance are great, and react with classes so slow. hooks:classes - 1:0

### DIFF
--- a/frameworks/keyed/react-hooks/src/main.jsx
+++ b/frameworks/keyed/react-hooks/src/main.jsx
@@ -1,7 +1,7 @@
-import React, { memo, useReducer, useCallback } from 'react';
-import ReactDOM from 'react-dom';
+import { memo, useReducer } from 'react';
+import { render } from 'react-dom';
 
-function random(max) { return Math.round(Math.random() * 1000) % max; }
+const random = (max) => Math.round(Math.random() * 1000) % max;
 
 const A = ["pretty", "large", "big", "small", "tall", "short", "long", "handsome", "plain", "quaint", "clean",
   "elegant", "easy", "angry", "crazy", "helpful", "mushy", "odd", "unsightly", "adorable", "important", "inexpensive",
@@ -12,19 +12,24 @@ const N = ["table", "chair", "house", "bbq", "desk", "car", "pony", "cookie", "s
 
 let nextId = 1;
 
-function buildData(count) {
+const buildData = (count) => {
   const data = new Array(count);
+
   for (let i = 0; i < count; i++) {
     data[i] = {
       id: nextId++,
       label: `${A[random(A.length)]} ${C[random(C.length)]} ${N[random(N.length)]}`,
     };
   }
-  return data;
-}
 
-function listReducer(state, action) {
+  return data;
+};
+
+const initialState = { data: [], selected: 0 };
+
+const listReducer = (state, action) => {
   const { data, selected } = state;
+
   switch (action.type) {
     case 'RUN':
       return { data: buildData(1000), selected: 0 };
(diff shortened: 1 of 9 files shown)


ryansolid:
Hmm.. a lot of changes here. And the React classes implementation was definitely in need of an overhaul. On the use of useCallback there was this previous PR #831 . Im concerned with all the changes whether we can no for sure in isolation which way is actually faster. The original implementation had them and then they were intentional removed.

oleggrishechkin:
@ryansolid 
I know how improve performance without useCallback - all row data (id, label, etc) should be moved in othe memoized component like RowData, but i think it's cheating too.

I post results in [issue](https://github.com/krausest/js-framework-benchmark/issues/838) with optimized react implementation with mutations, global state and direct subscribe to updated selection, label and data array. It's close to react-rxjs, but i get not so good results for react-rxjs on my machine

frameworks/keyed/react-hooks/src/main.jsx
```diff
@@ -34,35 +39,47 @@ function listReducer(state, action) {
...
+
         newData[i] = { id: r.id, label: r.label + " !!!" };
       }
       return { data: newData, selected };
     case 'CLEAR':
       return { data: [], selected: 0 };
     case 'SWAP_ROWS':
-      debugger;
-      return (data.length > 998) ? { data: [data[0], data[998], ...data.slice(2, 998), data[1], data[999]] } : state;
+      return { data: [data[0], data[998], ...data.slice(2, 998), data[1], data[999]] };
```
ryansolid, line 52: Probably shouldn't remove the length check here. While unnecessary for the test runs it is necessary for the operation not to die horribly after a couple of rows have been removed.

ryansolid:
@oleggrishechkin Yeah looking closer the Hooks differences are actually probably mostly Prettier. It's the removal of the length check on SWAP_ROWS.. the removal of the return `true` function on memo (I'm gathering it is unnecessary?) and the re-introduction of the useCallback, hoisting initial state, and some stylish stuff to use more destructuring and less property access. Am I missing anything? I'm just unclear where the perf gain is coming from. Because looks like mostly just undoes the last PR which was shown to be faster?

To be fair I never tested that myself, and I wrote the original hooks implementation with `useCallback`. I understand what you mean through different structure we could remove any need for `useCallback`. I'm just trying to understand this sort of back and forth tug of war with `useCallback`. Is there some subtle difference here that makes it different than it was before #831 in terms of performance?

oleggrishechkin:
@ryansolid, i think i concentrate on select row performance when say "performace increased"

Yer, mostly differences is Prettier and react with classes refactor - i use PureComponent, remove forceUpdate, create dispatch method like in hooks.

I added useCallback, because it was imported, but not used, and i think that somebody want add useCallback but forget about it. Also it's good practice to prevent unnecessary re-rendering - in classes you should use methods, in hooks - useCallback. But in this benchmark useCallbacks in each row is decrease rows creating performance, but in Real World example you prefer use virtualization/windowing instead of create so much rows

oleggrishechkin:
@ryansolid i added length check for swap, removed `useCallbacks`, and added `shouldComponentUpdate`, also i added second parameter to `memo` in react-hooks, it's should exclude `dispatch` while checking props eq and (i beat react-rxjs with it in my tests)

ryansolid:
Nice. Yeah I'm not here to state I know which way is faster. I will take whatever averages out faster. I just was unclear what that is when presented with seemingly contradictory information. If you test both and one churns out better scores I think we should go for it.

oleggrishechkin:
I did some tests across react-redux-hooks, react-hooks, react-rxjs (fastest in react case). I tested "with `useCallback`", with `memo` second parameter, with components for row label, row remove cross, with "`useSelector` on each row" and without. 

Fastest in all cases except select row - with memo second parameter (for react-hooks), pure row function component (without `useSelector` and `useCallback`). Other combinations was slower. So, that's why react-rxjs implementation fastest for now - it's use global states, and you no need provide dispatch prop.

In this PR react-hooks performance should be closer to react-rxjs

krausest:
Thanks. Here's a screenshot from the result (left new, right old). Maybe it's a little bit faster for create row and select row.

PR #870, opened 2021-03-25
1more - Immutable implementation

Freak613:
Removed observable overhead and switched to immutable state approach.
From my local testing, it has same or better results, which means that observables don't provide significant improvements anymore. Select case will be slower, but it's covered by better render results, at the same time significantly improving memory consumption.
So the only optimization test currently has is automatic sCU based on props reference equality.
It will make comparison even more interesting, without questioning library rendering efficiency.


krausest:
Thanks. Here's a screenshot from the result (left new, right old). Pretty close...

Freak613:
@krausest Thank you for your time and effort.
Do you accept donations? It feels unimaginable to run all the tests for 12hrs each time, so donate option could be good way to say thank you and compensate for your time.

krausest:
@Freak613 Thanks a lot for asking, but I'm already very happy when this benchmark gets positive feedback (which is really hard for any kind of benchmark). I think there are quite a few other open source projects that really need donations and I feel very privileged that I can make a good living from my job and run this project in my spare time.
Thanks to all the contributors (most of all @leeoniya and @ryansolid ) without whom this project wouldn't be worth a cent :smile:

leeoniya:
@krausest gives me way too much credit for sure.

> without whom this project wouldn't be worth a cent 😄

i've only ever put in my $0.02 🤣

PR #871, opened 2021-04-01
Add Maple framework

lukechu10:
This adds an implementation for the library [Maple](https://github.com/lukechu10/maple). It is written in Rust and runs in WASM.


lukechu10:
Haha, it's no where near the performance of [Solid](https://github.com/ryansolid/solid) right now but it's faster than quite a few WASM libraries.

krausest:

PR #872, opened 2021-04-05
Improved speed v0.30.4

hman61:


krausest:
I really want to avoid loading JS files from a CDN since it might influence the startup measurement. 
Can you please provide doohtml via npm or include it as the local doo.html.js?

hman61:
Sure, I will make that change, so don't merge yet

Also. I saw this on the master, and that is probably on reason that the
docker build is not working as expected

    "docker-build-frameworks": "npm run docker-sync && docker exec -it
js-framework-benchmark npm install && docker exec -it
js-framework-benchmark node docker-rebuil-all.js",

No "d"  in rebuild and I don't see a docker-rebuil-all.js on the project
root

PR #873, opened 2021-04-12
Update react-tagged-state

oleggrishechkin:
I'm update react-tagged-state example

Now it should be close to react-rxjs example (or even faster than react-rxjs) with the same state structure (two separate states for data and selected).

Both implementations (react-rxjs and react-tagged-state) shows how global state improve performance for React


krausest:

Nice improvement!

oleggrishechkin:
Thanks, but I don't understand how startup metrics works and why bootup time is not 16ms

PR #876, opened 2021-04-20
Doo html v0.30.5

hman61:


krausest:

Nice performance, but I had to add #772 for the implementation for using dom access.

hman61:
When will the official results get updated?

I still see the previous results for DooHTML on the official site.

Issue #860, opened 2021-03-04
Consider adding Stencil to the benchmark


LifeIsStrange:
I'm aware that it was mentioned in that closed issue:
https://github.com/krausest/js-framework-benchmark/issues/441
And I understand that you expect external contributors for adding frameworks as there are so many!

However, while (as you know) there is a HUGE number of js "frameworks" out there, I have taken a look at all the missing frameworks referenced in that issue and Stencil *really* stands out:
Here's my analysis of why:
From that list there are three frameworks that have a decent number of commits (and therefore expected maturity):
Omi:
https://github.com/Tencent/omi
But it has a very low number of users ~100
Skatejs:
https://github.com/skatejs/skatejs
But it's development has stopped in 2019
And
Stencil:
https://stenciljs.com
Which has ~10000 users on github which for reference is 3 times as much as infernojs.
I would say that it is the 5th most popular vdom framework.
It also is used by Apple and Microsoft!
And it has a huge downstream user: the Ionic framework (which is the Electron of smartphones). Stencil has a similar value proposition than Svelte: it compiles to web-components and therefore allow Ionic to implement once a widget and make it usable in any framework. In addition to the interoperability, this compilation ability usually allow next gen performance (Svelte is the fastest of the 4th popular Vdoms for this reason) and benchmarking Stencil would be very interesting as it could be even faster than svelte and therefore push the boundaries of what a production ready js framework can do! 
I also think that it would increase its popularity because having the performance of a framework as an unknown is a major risk.

Hence why I hope that you agree that Stencil support should be prioritarized and that letting this issue open would help give it visibility.

ryansolid:
Stencil does really standout as being absent. The thing I've found though is it takes a certain expertise in the framework to be able to write these benchmarks in a way that is representative. Everyone that I've added I've pretty much needed to talk to the author, core maintainer about details to make sure I was doing the right thing.

People write these issues, but very rarely does it result in new submissions. I've had the best luck going to the source repo and asking there. This is where I'd be putting my focus: https://github.com/ionic-team/stencil/issues/2765

I did see this: https://github.com/ionic-team/stencil/issues/2716. But I disagree strongly with the response. The problem is while I don't want to put this on the maintainers, it's the like RealWorld demo or TodoMVC except more stringent as we are showcasing performance.

[...]

PR #879, opened 2021-04-26
Add Stencil-Keyed

ryansolid:
I got impatient after waiting on this for a couple of years now and made it.

I believe this is the best I can do here, as Stencil does not support Native Built-Ins and only has Web Component components. This forces me into a single component implementation that usually is not beneficial for VDOM libraries. The strict DOM structure means this is what I'm stuck with.

Roughly speaking Vue 3 < Stencil < React Hooks.

So hopefully this makes someone following #860 's day.


krausest:
Thanks. Results are updated:

PR #880, opened 2021-04-26
Add hydro-js implementation

Krutsch:
I hope this is correct, let me know otherwise! 😊


frameworks/keyed/hydro-js/src/app.ts
```diff
@@ -0,0 +1,179 @@
...
+  for (let i = 0; i < oldData.length && newData.length; i++) {
+    oldData[i].id = newData[i].id;
+    oldData[i].label = newData[i].label;
+    newData[i] = oldData[i];
+  }
+
+  for (let i = oldData.length; i < newData.length; i++) {
+    renderItem(newData[i], i);
+  }
+}
```
ryansolid, line 161: I suspect this implementation is non-keyed. Can't know for certain without running it but it looks like replace at index and render any additional rows.

Krutsch: Thanks for having a look. I think it is keyed. A change in the data will automatically change the table. You will see this even better on the clear function. However I do not have a functionality that would add new DOM elements if an item is being inserted. That's why I have to observe the data and add new elements to the table. Would that make it non-keyed?

L152 - L156 could be deleted, they are used as optimization. I am not sure if this is allowed: so if we have 1000 elements already and click on add 1000 elements, it will reuse the existing DOM elements and override the data without destroying the key connection.

ryansolid: I see, the `merge` function raised immediate flags with me as it is basically how you would write a non-keyed(index based) reconciler so I thought that was how it was updating. 152-156 is fine assuming it is the same data (same key). Like if I swap 2 rows, with keyed I want to see 2 DOM elements moved. Those TRs literally change locations in the table, not just have the text nodes values change (which is what this assignment looked like).

Other than that without a templating language it is easy to be classified as #772. On one hand you only use DOM commands on creation. The "template" code more or less looks like the compiled output of a reactive friendly view library. On the other hand, the mapping to the DOM nodes is in end user land (ie there is no list helper) which means from an implementation standpoint it is DOM manipulation almost by definition. We've classified this as #772 in the past.

Krutsch: Thanks again. I will use a templating language to fix #772. 

I do not have the logic to move elements. If there is a swap, the nodes would change their values, just as said. If this is a requirement for a keyed implementation, then I would move it to non-keyed for now and come up with a solution in the future. Otherwise I probably have to move the elements manually..

ryansolid: Yeah I believe this makes it non-keyed. Here is an old article talking about it in more detail: https://www.stefankrause.net/wp/?p=342#:~:text=In%20%E2%80%9Ckeyed%E2%80%9D%20mode%20the%20row,the%20last%20row%20is%20deleted.

Krutsch: I moved it to non-keyed and marked it as 772. I would have liked to use the templating language, but for some reason appending the node then will increase the time from 0.22ms to 6ms. The same node, no changes.. Could't figure out why 😔

Krutsch:
#772 has been fixed. This can be merged. I will create another PR for a keyed version at some point :)

krausest:
I'm getting the following error. Can you take a look at it please?
```
root@df3f610a6606:/build/frameworks/non-keyed/hydro-js# npm run build-prod

> js-framework-benchmark-hydro-js@1.3.4 build-prod /build/frameworks/non-keyed/hydro-js
> html-bundle

file:///build/frameworks/non-keyed/hydro-js/node_modules/html-bundle/dist/bundle.js:16
import { serialize, parse, parseFragment } from "parse5";
                    ^^^^^
SyntaxError: The requested module 'parse5' is expected to be of type CommonJS, which does not support named exports. CommonJS modules can be imported by importing the default export.
For example:
import pkg from 'parse5';
const { serialize, parse, parseFragment } = pkg;
    at ModuleJob._instantiate (internal/modules/esm/module_job.js:97:21)
    at async ModuleJob.run (internal/modules/esm/module_job.js:135:5)
    at async Loader.import (internal/modules/esm/loader.js:178:24)

```

Krutsch:
Thanks, I will check! Which node version are you using?

Krutsch:
Should work now in node v10+

krausest:
Thanks. That worked fine.

PR #881, opened 2021-05-02
DooHTML v0.30.6

hman61:


krausest:
Thanks. Results are updated. Looks like a major regression for create many rows:

PR #882, opened 2021-05-02
DooHTML-v0.30.7

hman61:


hman61:
Trying another version of runLots. I still getting errors in my local Docker so it is hard to measure performance accurately

krausest:
Looks good (pretty good indead)! Results are not yet updated, I must get rid of the -random versions first.


What's the error you're getting for the docker build?

hman61:
I was wondering if you would consider allowing multiple <tbody> in the
result. Any test for <tr> count would still return the correct number, but
I don't know if the current tests are counting table rows or childNodes of
#tbody

krausest:
I'm afraid this won't work. Both the test driver and the isKeyed test use xpath expressions like //tbody/tr[991] to check the DOM structure.

PR #886, opened 2021-05-08
Remove Lit polyfill

AndrewJakubowicz:
PR to resolve this issue: https://github.com/lit/lit/issues/1855
Removed the polyfill.

Is there any reason we'd want to keep the polyfill?
 - In [many cases](https://lit.dev/docs/tools/requirements/#modern-browser-breakdown) it shouldn't be required.

### Testing

I ran `npm run check keyed/lit` and `npm run check non-keyed/lit`. I didn't benchmark the non-keyed/lit, but did benchmark keyed/lit.

### Screenshot

image


krausest:

PR #887, opened 2021-05-08
Doo html v0.30.8 non keyed

hman61:


krausest:
Thanks. The isKeyed test thinks the non-keyed version is actuallly rather keyed. What does it make non-keyed to you? 
```
Non-keyed test for swap failed. Expected than no TRs are added or removed, but there were 2 added and 2 removed
Non-keyed test for create rows failed. Expected that TRs should be recycled, but there were 1000 added TRs and 1000 were removed
Non-keyed test for remove failed. Expected that the dom node for the 2nd row would NOT be removed, but it was.
doohtml-vdoohtml-non-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as keyed in the results
```

krausest:
P.S: I'd rather add a non-keyed version only if the framework doesn't support keyed updates. (But in that case I would sort that framework out :smile: )

PR #889, opened 2021-05-13
React + @grammarly/focal@0.8.5 benchmark

KatSick:
Here is a benchmark for [@grammarly/focal library](https://github.com/grammarly/focal).
All local tests are passed 
image

```bash
~/Projects/js-framework-benchmark/webdriver-ts % npm run check keyed/react-focal

> webdriver-ts@1.0.0 check /Users/ostap.chervak/Projects/js-framework-benchmark/webdriver-ts
> cross-env LANG="en_US.UTF-8" node dist/isKeyed.js "keyed/react-focal"

args.framework undefined true
http://localhost:8080/frameworks/keyed/react-focal/package-lock.json
Frameworks that will be checked react-focal-v17.0.1 + 0.8.5-keyed
react-focal-v17.0.1 + 0.8.5-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as keyed in the results
```

### DIFF
--- a/frameworks/keyed/react-focal/.prettierrc
+++ b/frameworks/keyed/react-focal/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "semi": false,
+  "singleQuote": true,
+  "trailingComma": "none",
+  "arrowParens": "avoid",
+  "printWidth": 100
+}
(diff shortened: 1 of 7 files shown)


krausest:
Thanks. [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated. Seems a bit slower than the other react variants though.

PR #894, opened 2021-05-28
Add "Old Scull Framework" benchmark

alexeyoganezov:


krausest:

PR #895, opened 2021-05-28
DooHTML-non-keyed-v0.40.2

hman61:


krausest:
Thanks. It's included in the "dark" side of the [table](https://krausest.github.io/js-framework-benchmark/current.html) :smile:

PR #897, opened 2021-06-05
Update valtio

dai-shi:
and avoid copying data when building.


krausest:
Thanks. Here a screenshot fro the results, seems pretty much unchanged:

dai-shi:
Thanks. Hm, okay.

PR #898, opened 2021-06-10
update Ember version from 3.22 to 3.27

lifeart:
prev mr: https://github.com/krausest/js-framework-benchmark/pull/814


krausest:
Thanks. Results are updated. Ember looks a bit slower, glimmer got a bit faster.

PR #901, opened 2021-06-14
Update Alpine to V3

ryansolid:
Alpine now uses Vue 3 reactivity. Seems to be way faster.


krausest:
Thanks. Results are updated. A big step for alpine!

ryansolid:
Thank you

PR #902, opened 2021-06-15
Jotai: Update to 1.0.0

Thisen:
Updates Jotai to 1.0.0.
This PR also replaces rollup build with Webpack build. This will align the build tooling with most of the comparable libs to Jotai.


krausest:
Thanks. Results are updated. A significant improvement!

PR #903, opened 2021-06-18
Update uhtml to v2.7.5

WebReflection:
The following MR provides:
  * both delegate and tr based version of the keyed version
  * keyed benchmark uses the delegate, as discussed in https://github.com/krausest/js-framework-benchmark/pull/878
  * all dependencies and dev dependencies have been updated
  * new syntax showcase has been added
  * changed the ugly uHTML for non keyed, made it uhtml

Thanks for considering this update.

### DIFF
--- a/frameworks/keyed/uhtml/package-lock.json
+++ b/frameworks/keyed/uhtml/package-lock.json
@@ -1,44 +1,839 @@
 {
   "name": "js-framework-benchmark-uhtml",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "js-framework-benchmark-uhtml",
+      "version": "1.0.0",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "js-framework-benchmark-utils": "^0.2.5",
+        "uhtml": "^2.7.5"
+      },
+      "devDependencies": {
+        "@rollup/plugin-node-resolve": "^13.0.0",
+        "@ungap/degap": "^0.2.6",
+        "rollup": "^2.52.1",
+        "rollup-plugin-includepaths": "^0.2.4",
+        "rollup-plugin-minify-html-literals": "^1.2.6",
+        "rollup-plugin-terser": "^7.0.2"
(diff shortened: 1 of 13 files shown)


WebReflection:
P.S. looks like various projects here are taking shortcuts for the select specific use case, assuming there will be only one selected node per time. I could cheat on *uhtml* code too with an early return, but that would not showcase real-world code (current MR).

**Example**
```js
let selectedRow = null;
let selectedId = -1;

export default (state) => {
  const {data, selected} = state;
  if (selected !== selectedId) {
    if (selectedRow)
      selectedRow.classList.toggle('danger', false);
    selectedRow = document.getElementById(selectedId = selected);
    if (selectedRow) {
      selectedId = selected;
      selectedRow.classList.toggle('danger', true);
      return;
    }
  }
  return html.for(state)`
    <table class="table table-hover table-striped test-data"
      @click=${handler} .state=${state}>
      <tbody>${
      data.map(item => {
        const {id, label} = item;
        return html.for(data, id)`
        <tr id=${id}>
          <td class="col-md-1">${id}</td>
          <td class="col-md-4">
            <a data-action="select">${label}</a>
          </td>
          <td class="col-md-1">
            <a data-action="remove">
              <span class="glyphicon glyphicon-remove" aria-hidden="true" />
            </a>
          </td>
          <td class="col-md-6" />
        </tr>`;
      })
      }</tbody>
    </table>
  `;
};
```

This is whre *uhtml* would sit, if I used above code instead.

Screenshot from 2021-06-19 08-06-57

I am not asking for actions, just pointing out that these "*issues*" could be improved. The select case, if the node is retained and no diff is performed, or no idiomatic code is around that specific case, should be, imho, flagged too.

krausest:
Thanks! [Results](https://krausest.github.io/js-framework-benchmark/current.html?t=12380132) are updated.

Toggling the css class directly should be flagged #772 (see also https://github.com/krausest/js-framework-benchmark/pull/878#issuecomment-864369900).

WebReflection:
Thank you ❤ ... and yes, [I did notice the updated results](https://twitter.com/WebReflection/status/1406245821306318850) 🥳

PR #907, opened 2021-06-20
Update Solid to 1.0.0

ryansolid:
I've been doing some last min pre-release performance checking for Solid's 1.0 release. This is likely one of the last Release Candidates so it's a good time to verify any regressions etc, although admittedly as soon as 1.0 comes out I'm probably back here again.

That being said I think I've made performance improvements. It's pretty hard to gain ground here but running on the Macbook Air with no slowdowns showed me slightly more accented bottlenecks that I was able to address on Remove Row and Clear Rows. Whether that translates elsewhere remains to be seen. I might be reaching the limits of improvement here in terms of approach.

Reducing the gap with Mikado has been a subgoal, but a general-purpose subscription system is going to have difficulty breaking through. I've concluded:
* Creation speed should be similar but memory overhead for data structures don't scale as well as direct proxies. This may not always matter but not much trimming to be done here.
* Updates on direct proxies are going to be unmatched. Where setting a property directly updates the DOM in Mikado, in Solid it gets queued and executed against a list of subscribers. In this sense, the data is independent of the view and can be applied against multiple views at the same time, but is a more complicated execution to account for more than 1 to 1 scenarios.
* Cleanup has a similar issue since with direct proxies you don't have to clean them up as simply losing the reference when the DOM node goes away will allow GC. General-purpose has the leaky observer problem where data can persist beyond the view and needs to be collected to be manually cleaned up. 

The biggest places left to gain ground here at least when running on my Macbook Air was around scheduling when creation and cleanup runs. Big creates and cleans on the same RAF cycles are slower. We saw hints of this in the old RAF issues but I saw similar effect simply on managing the timing of library internal data structure cleanup. I managed to figure out a way to handle internal disposal scheduling without resorting to DOM RAF stuff which can cause broken behavior. But I made no ground on replace rows even when reducing the operation to a clear and create (knowledge the benchmark should never have) to remove diffing from the equation. So just hitting the boundaries of memory I imagine.

### DIFF
--- a/frameworks/keyed/solid-state/package-lock.json
+++ b/frameworks/keyed/solid-state/package-lock.json
@@ -1,239 +1,1143 @@
 {
   "name": "js-framework-benchmark-solid-state",
-  "version": "0.20.0",
-  "lockfileVersion": 1,
+  "version": "1.0.0",
+  "lockfileVersion": 2,
   "requires": true,
-  "dependencies": {
-    "@babel/code-frame": {
-      "version": "7.12.11",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.11.tgz",
-      "integrity": "sha512-Zt1yodBx1UcyiePMSkWnU4hPqhwq7hGi2nFL1LeA3EUl+q2LQx16MISgJ0+z7dnmgvP9QtIleuETGOiOH1RcIw==",
-      "dev": true,
-      "requires": {
-        "@babel/highlight": "^7.10.4"
+  "packages": {
+    "": {
+      "name": "js-framework-benchmark-solid-state",
+      "version": "1.0.0",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "solid-js": "1.0.0-rc.7"
+      },
+      "devDependencies": {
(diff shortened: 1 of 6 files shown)


krausest:
Results for solid are updated. I'll try to re-run all frameworks this night.

krausest:
Just to make sure the security update for chrome and the necessary switch from intel to modesetting doesn't mix results up I ran the benchmark for all frameworks. Results are updated.

leeoniya:
hmm, wonder when this happened :thinking: 

Screenshot_20210621-063459_Opera_beta

krausest:
Strange, but in the oder chrome 91 results it already looked similar:
image


It was much lower for chrome 90:
image

ryansolid:
Hmm.. yeah that is almost always telltale sign of memory leak when those 2 light up. But for the browser version to be causing it... That's sort of crazy. A bug in Chrome perhaps.

@krausest thanks for the updates. Clear rows improved but remove not so much. It looks like Solid's (proper not state) position might be slightly favorable averagely but it looks like every framework has widened their gap from Vanilla with the recent run. It is all within the same range though.

PR #909, opened 2021-07-01
update fntags to 0.1.7

snowbldr:
I was using setTimeout to schedule an async task instead of creating a Promise because I'm a noob. I'm expecting this version to be a bit quicker than the last one.


krausest:
Yep, a little bit faster :open_mouth: 
Screenshot from 2021-07-03 21-32-04
Not bad! Did you expect that?

snowbldr:
Yep, that's pretty much what I was expecting.

PR #911, opened 2021-07-07
Update maple to sycamore v0.5.1

lukechu10:
I made a lot of architectural changes so performance is a bit different. Based on my own measurements, some stats went up while others went down.

For some reason, create 10k rows still performs horribly. Based on chrome's profiling tool, most of the time is taken up by "Major GC". Still need to do some more debugging to find out what is causing all the garbage.


krausest:
Thanks. I added the bundled-dist files. 
Results are updated - looks good except for the "create 10,000 rows" benchmark.

PR #915, opened 2021-07-25
Bump Alpine, Hyperapp, Preact, Solid, Svelte, Vue version

alpacone:


krausest:
Thanks a lot. Here's a comparison:
Screenshot from 2021-07-25 21-12-17

PR #917, opened 2021-07-29
add lui

L3P3:
https://github.com/L3P3/lui

Following my code style used across all lui projects including lui itself.

Since lui is so tiny, I dared adding the build artifact directly. This is how I do it usually but in the real world, I load lui via jsdelivr. I am fine with a download size of 10k for now.

`npm run check` ran through without issues. :grin:

I am really gespannt for your feedback. :sweat_smile:


L3P3:
@krausest could you give me feedback please? Is this acceptable? Do I need to change it? If you have a problem with having lui.js in the repo, I may add a build script that downloads the lui.js from jsdelivr instead... :grimacing:

krausest:
Sorry, I was just short of time. Lui is now included in the results:
Screenshot from 2021-08-10 22-01-59

PR #918, opened 2021-08-04
Bump 1more, alpinejs, attodom, lit, redom, riot, solid, stencil, svelte

alpacone:


krausest:
Thanks. Here's a comparison:
Screenshot from 2021-08-10 22-01-59

PR #921, opened 2021-08-09
Update vue to 3.2

yyx990803:
- Update Vue to 3.2.1
- Now uses `<script setup>` - this is just syntax simplification, implementation logic remains the same
- Added use of [`v-memo`](https://v3.vuejs.org/api/directives.html#v-memo) on the `v-for` list (similar memo optimizations are present in almost all other vdom-based implementations, in the form of `shouldComponentUpdate`, `React.memo` etc.)
- Updated `webpack` version in Vue benchmarks


krausest:
Thanks! Here are the results from my machine:

Very nice improvement for vue, even though it didn't beat svelte on my machine.

Issue #916, opened 2021-07-29
deterministic memory metrics


leeoniya:
i've been investigating whether domvm has a real memory leak (in https://github.com/domvm/domvm/issues/229) or if this was simply a recent change in chrome's GC behavior. right now it's looking like the latter (see profiles in that issue).

wonder if chromedriver or lighthouse has an explicit api to collect garbage, like puppeteer [1] which can be invoked prior to taking mem measurements. would be curious to see if this makes a difference for some libs.

cc @krausest @ryansolid 

[1] https://github.com/puppeteer/puppeteer/issues/1394#issuecomment-355722418

krausest:
The memory benchmark (tries to) run the GC after initBenchmark and each runBenchmark loop: https://github.com/krausest/js-framework-benchmark/blob/master/webdriver-ts/src/forkedBenchmarkRunner.ts#L271
chrome has to be started with a special flag to accept window.gc calls:
https://github.com/krausest/js-framework-benchmark/blob/master/webdriver-ts/src/webdriverAccess.ts#L239
So there should be a garbage collection between all runs.

leeoniya:
that mem metric was in line with other adjacent libs for a long time and changed suddenly without any lib or bench impl updates, so idk 🤷

leeoniya:
here are 3 cycles of "create 1k" x 5, with a few seconds in between, then a forced GC from devtools at about 17500ms:

![image](https://user-images.githubusercontent.com/43234/127559880-f5c4fa41-84f3-4900-b805-9f04c1c2f815.png)

you can see the GC tends to collect under pressure (e.g. on the first or middle run of the _next_ 5x set); simply being idle after each 5x set is not always sufficient, even though the memory can technically be reclaimed.

same story for create/clear:

![image](https://user-images.githubusercontent.com/43234/127560489-8637509f-5c76-40ff-9789-66fab916c097.png)

leeoniya:
below are similar runs for ivi. (certainly, domvm is doing something that makes the GC quite a bit lazier during idle periods, i havent had much luck with tracking this down)

![image](https://user-images.githubusercontent.com/43234/127561254-5f5506aa-3006-4561-8d09-c5e2535932c3.png)

![image](https://user-images.githubusercontent.com/43234/127561464-2baeedcd-8508-4860-aac6-a036cfb3a311.png)

leeoniya:
finally, react-hooks:

[...]

PR #922, opened 2021-08-10
Arrange elements in dropdowns in vertical order instead of horizontal

ishitatsuyuki:
The framework list is currently arranged like:

A B C
D E F

This patch changes it to:

A C E
B D F

This makes it easier to scan through the list in alphabetical order when
looking for a certain name.

The implementation is done with CSS multi-column layouts. There seems to be
some caveats in the browser support table [1], but this patch only uses the
basic part, so it should be OK.

[1] https://caniuse.com/multicolumn

### DIFF
--- a/webdriver-ts-results/src/App.css
+++ b/webdriver-ts-results/src/App.css
@@ -94,11 +94,7 @@ input + label {
 }
 
 .grid {
-    display: flex;
-    flex-wrap: wrap;
-}
-.col-3 {
-    width: 33%;
+    column-count: 3;
 }
 
 .hspan {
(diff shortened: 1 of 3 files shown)


ishitatsuyuki:
Here's a screenshot, to show it visually.

Before:

before


After:

after

ishitatsuyuki:
@krausest what do you think of this?

krausest:
@ishitatsuyuki I think I'll switch to vertical layout, but I'd rather do it in code and not via css.
But I haven't taken time to do that yet since I'm currently working on #916 and #947 which really make me worry.

ishitatsuyuki:
I don't understand. This can be done with just 10 lines of CSS change. Why do you think doing it in JavaScript is better?

PR #924, opened 2021-08-12
Add ef.js

ClassicOldSong:
Add [ef.js](https://ef.js.org)

https://github.com/TheNeuronProject/ef.js

As ef.js relies heavly on optimization at component creation,  the results on creating may look bad. But considering it's updating performance, it's still pretty impressive.


krausest:
Thanks for the PR. It's now included in the [results](https://krausest.github.io/js-framework-benchmark/current.html).

PR #929, opened 2021-08-24
keyed/wasm-bindgen: update all dependencies and rebuild with latest Rust

utkarshkukreti:
I updated all Rust and NPM dependencies and rebuilt with Rust 1.54.0.

I ran the benchmarks a few times. There seems to be a tiny improvement in all my runs, roughly ~1.07 -> ~1.05 on average. The bigger difference is the wasm size, which is now 8.4kb smaller, from 44.4kb -> 36kb.

No changes were made to the actual source code.


krausest:
Thanks - I updated the [results](https://krausest.github.io/js-framework-benchmark/current.html). CPU results are pretty close to the old version on my machine.

PR #933, opened 2021-08-29
add fre

yisar:


krausest:
Good news. It builds fine and the test runs successfully:

As expected select rows returns a bad result (will take a look into soon)
Memory results look a bit odd. Please note that due to #916 a few frameworks report unexpected high memory usage.


BTW: If you change the following line in package.json  to `"frameworkVersionFromPackage": "fre"` it'll report the correct fre version.

krausest:
Here a few screenshots from the chrome timeline for select rows. It shows that indeed fre is performing much more work for select row than the other frameworks (here without any CPU slowdown):
vanilla.js (2.61 msecs)
Screenshot from 2021-08-29 21-23-21
vue.js (2.61 msecs)
Screenshot from 2021-08-29 21-20-09
react-hooks (4.71 msecs)
Screenshot from 2021-08-29 21-24-44
fre (40.57 secs):
Screenshot from 2021-08-29 21-26-05

If you want I can merge this PR. Should I do that?

yisar:
@krausest Yes, you can do it.
Because fre does not implement `fre.memo`, the current performance is poor, and CPU slowdown magnifies this defect.
But it’s okay, I will improve it in the next version.

PR #935, opened 2021-08-30
Bump `compostate-jsx`

LXSMNSYC:
- Minor core improvements (difference may not be noticeable).
- Dependencies of the example app has been updated.


krausest:
Thanks. Results are updated:

PR #937, opened 2021-09-02
keyed/svelte: optimize hot loop by eliminating extra whitespace nodes

utkarshkukreti:
Svelte does not currently [1] remove purely whitespace text nodes like
most other frameworks do, which causes noticeable performance impact
when benchmarking creating thousands of rows.

This commit manually fixes that so that the generated HTML is closer to
the competing implementations, using a workaround mentioned in [1].

On my system, all the benchmarks involving creation of rows run ~5%
faster with this change and the overall geometric mean improves from
around 1.28 to 1.25.

[1]: https://github.com/sveltejs/svelte/issues/6540

### DIFF
--- a/frameworks/keyed/svelte/src/Main.svelte
+++ b/frameworks/keyed/svelte/src/Main.svelte
@@ -31,15 +31,16 @@
 </div>
 <table class="table table-hover table-striped test-data">
   <tbody>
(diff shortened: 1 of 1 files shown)


krausest:
Thanks. Results are updated:

PR #939, opened 2021-09-09
Update keyed/jotai

dai-shi:
package update and slight change (atomic updating) in code.
I don't see significant difference (a little slower) when I run it locally.


krausest:
Thanks. Due my attempts to fix #916 I had to re-run all frameworks with the new chrome 94 version so this time I can't offer a before and after comparison for the framework version bump.

PR #941, opened 2021-09-12
Update sycamore to v0.6.0

lukechu10:
This PR updates Sycamore from 0.5.1 to 0.6.0

This new release should have some minor performance improvements across the board.


krausest:
Thanks. Due my attempts to fix #916 I had to re-run all frameworks with the new chrome 94 version so this time I can't offer a before and after comparison for the framework version bump.

PR #945, opened 2021-10-03
Doo html keyed 0.60.2

hman61:


krausest:
Thanks. Looks like creating rows got way slower in the new version ...

hman61:
Thanks for letting me know, would you mind reverting pull request #945

Issue #936, opened 2021-08-30
Improve Non-Keyed Vanilla Implementations


ryansolid:
I tend to treat that table like it doesn't exist so I haven't looked at it for months (years?), but noticed Remove Rows in the vanilla implementation seems to be slower than a handful of implementations. We should look at improving it.

If anyone in the community familiar with Non-Keyed wants to contribute I think this is a worthwhile effort. Vanilla should be the fastest and without it being so it compromises the value of the particular tests.

LXSMNSYC:
I'm interested on improving this, although I was curious as to why the `Remove Row` seems to be slower than the keyed version

LXSMNSYC:
I had written my own take on the vanillajs implementation
![image](https://user-images.githubusercontent.com/4783372/131333010-d1847d47-2e8b-419f-89b7-3396cc2db3f8.png)

Tho it seems to me that both implementations are just patching elements and not actually replacing them. Is that a valid approach?

LXSMNSYC:
![image](https://user-images.githubusercontent.com/4783372/131340357-02bc70ff-b4a9-4e77-9b3c-b42abad840eb.png)

Anyways, I didn't saw the note above so I just reworked it a bit.

krausest:
@LXSMNSYC Would you like to submit a PR for that?

LXSMNSYC:
> @LXSMNSYC Would you like to submit a PR for that?

It's actually my own take. I didn't try to update both vanillajs and vanillajs-1 due to their select implementation. In a nutshell, my implementation assumes that data and view are both in sync.

krausest:
That would be fine. You could just submit a new vanillajs implementation and I'd remove the others when I checked that it's the fastest.

PR #946, opened 2021-10-04
New non-keyed VanillaJS implementation

lxsmnsyc:
Mentioned at https://github.com/krausest/js-framework-benchmark/issues/936#issuecomment-918080302

# Implementation details
- Unlike `vanillajs` and `vanillajs-1`, this implementation assumes that the store and the UI are synchronized, instead of deriving UI from the store. This makes sure that it can skip a lot of unnecessary runs that were happening from the previous implementations (for instance, `delete row`)
- There could be some potential optimizations. In most parts it can be slower than `vanillajs`, it's just that the average score of the implementation is, most of the time, a lot better than the predecessors.


frameworks/non-keyed/vanillajs-2/src/Main.js
```diff
@@ -0,0 +1,186 @@
...
+            }
+        } else if (rowL < dataL) {
+            for (let i = rowL; i < dataL; i++) {
+                const node = createRow(data[i].id, data[i].label);
+                tbody.appendChild(node);
+                currentData[i] = data[i];
+                rows[i] = node;
+            }
+            for (let i = 0; i < rowL; i++) {
+                patchRow(rows[i], data[i].id, data[i].label);
```
xania, line 82: suggestion: inline the body of patchRow here to have better perf

xania: nevermind, you would end up with code duplication so ignore my suggestion

PR #958, opened 2021-11-10
Update fre 2.4.3

yisar:


krausest:
Thanks. Here are the results:

The only framework in this batch with a clear improvement :+1:

PR #960, opened 2021-11-14
Update blazor-wasm to .NET 6

ronnygunawan:
Might need to run `dotnet workload install wasm-tools` before running the project.


serge-wq:
I'm very interested in seeing how this update affect the results.

krausest:
Thanks. Here are the results:

Actually I hoped for a real improvement, but I'm afraid to say there isn't much improvement.

I changed the blazor-wasm build due to its heavy dependencies and bundeled the (former) dist folder. If you want to install or build you'd have to choose the install-force and build-prod-force script targets.

PR #961, opened 2021-11-16
Update `yew` to `0.18`

a1phyr:
Additionally:
- Update other dependencies
- Switch to edition 2021 (requires `rustc` 1.56)
- Optimize `RowData::new` to make a single allocation


krausest:
Thanks. Here are the results:

krausest:
Thanks. Here are the results:

Actually I hoped for a real improvement, but I'm afraid to say there isn't much improvement.

I changed the blazor-wasm build due to its heavy dependencies and bundeled the (former) dist folder. If you want to install or build you'd have to choose the install-force and build-prod-force script targets.

PR #972, opened 2021-11-27
add keyed react-lab

guojin-long:
"[React Lab](https://github.com/guojin-long/react-lab)" is an experimental branch of React that focuses on performance optimization.


krausest:
Thanks. Actually I hoped it would be even a bit faster..
[results](https://krausest.github.io/js-framework-benchmark/current.html) are updated.

guojin-long:
The current pull request only contains optimizations related to reconcile children.

PR #973, opened 2021-12-20
Add Zustand

ImedAdel:
This PR adds [Zustand](https://github.com/pmndrs/zustand). It still generates some timeout errors that I can't seem to solve though.

Also, I'm not sure if I should also edit some other files. Tell me if something's off :)


krausest:
Thanks. The add operation had an error and I made a small fix for it.

ImedAdel:
Thank you! Dunno how I missed that :sweat_smile: 

It was really disappointing seeing Zustand come out _that_ slow...

oleggrishechkin:
@ImedAdel, you used slightly more useStore() than needed especially in Row component. I think that in row you should select only isSelected and in Main you should use array of data directly, not ids. If you do that, result will be better

PR #977, opened 2021-12-26
Add Malina.js keyed

lega911:
Implementation for [Malina.js](https://github.com/malinajs/malinajs)


krausest:
Thanks. Nice Results!

PR #979, opened 2021-12-29
Update react-tagged-state

oleggrishechkin:


krausest:
Thanks. I did an update to my linux system but didn't notice that it would update to chrome 97. So I had no chance to compare it against the old chrome 96 results. So now it's included in the chrome 97 results: https://krausest.github.io/js-framework-benchmark/2022/table_chrome_97.0.4692.71.html

PR #980, opened 2021-12-29
Update Zustand implementation

ImedAdel:
This PR updates the Zustand implementation following the recommendations by @oleggrishechkin (https://github.com/krausest/js-framework-benchmark/pull/973#issuecomment-1002744463)

### DIFF
--- a/frameworks/keyed/react-zustand/src/main.jsx
+++ b/frameworks/keyed/react-zustand/src/main.jsx
@@ -218,7 +218,7 @@ const Main = () => {
 			<Jumbotron />
 			<table className="table table-hover table-striped test-data">
 				<tbody>
-					{items.map((item, i) => (
+					{items.map((item) => (
 						<Row key={item.id} item={item} />
 					))}
 				</tbody>
--- a/frameworks/keyed/react-zustand/src/main.jsx
+++ b/frameworks/keyed/react-zustand/src/main.jsx
@@ -120,27 +120,21 @@ const getSelect = (state) => state.select
 const getRemove = (state) => state.remove
 
 const Row = memo(
-	({ itemId }) => {
+	({ item }) => {
 		const select = useStore(getSelect)
 		const remove = useStore(getRemove)
-		const itemLabel = useStore(
-			useCallback(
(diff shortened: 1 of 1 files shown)


frameworks/keyed/react-zustand/src/main.jsx
```diff
@@ -211,21 +205,21 @@ const Jumbotron = memo(
...
 
-const getIds = (state) => state.data.map((el) => el.id)
+const getItems = (state) => state.data
 
 const Main = () => {
 	/**
 	 * @type {string[]}
 	 */
-	const ids = useStore(getIds, shallow)
+	const items = useStore(getItems, shallow)
```
oleggrishechkin, line 214: shallow is really needed here? looks like you can omit it 

it compare each item, right?

ImedAdel: Hmm. Wouldn't removing it trigger a re-render any time anything in the store changes (since we're selecting a list of objects)?

oleggrishechkin: state.data === state.data will be faster

So when data changed - re-render list

When selected changed - re-render only selected

ImedAdel: Okay, let me quickly test that before changing this :)

ImedAdel: Update: removing `shallow` drops the geometric mean from 1.74 to 1.85

ImedAdel: Yeah, no other changes I tried gave me similar results to the one with `shallow`, so I think we should keep it. Still, not much performance difference between Hooks and Zustand. I guess the latter only shines with transient updates.

krausest:
Thanks. I did an update to my linux system but didn't notice that it would update to chrome 97. So I had no chance to compare it against the old chrome 96 results. So now it's included in the chrome 97 results: https://krausest.github.io/js-framework-benchmark/2022/table_chrome_97.0.4692.71.html
Looks much better now.

PR #985, opened 2022-01-06
Add yew framework written with hooks

voidpumpkin:
Yew.rs has two flavors to write web apps - Struct components and Hooks. (Just like react)
This pr adds the hooks version.


frameworks/keyed/yew/src/lib.rs
```diff
@@ -201,7 +201,7 @@ impl Component for Jumbotron {
             <div class="jumbotron">
                 <div class="row">
                     <div class="col-md-6">
-                        <h1>{ "Yew" }</h1>
+                        <h1>{ "Yew-Hooks" }</h1>
```
lukechu10, line 204: I believe this should be in the `yew-hooks` implementation, not the `yew` implementation

krausest:
Thanks. yew-hooks is now included:
Screenshot from 2022-01-15 22-34-59

PR #992, opened 2022-01-30
Upgrade Sinuous v0.28.0

luwes:
This upgrades Sinuous to the latest version.

- Reverts to the old map diff algo which performed a lot better
  https://github.com/luwes/sinuous/pull/192


krausest:
I suppose this is not what you expected, results are almost unchanged (ecept for clear rows):

luwes:
thanks! yes the results are not what I expected, strange what could have caused this slow down. will have to dig 😅

PR #997, opened 2022-02-05
Master puppeteer windows

puckowski:
* Add common path for Chrome on Windows for Puppeteer. At least try launching Chrome instead of throwing an error on Windows.


krausest:
Thanks. I merged the windows path change manually.

PR #999, opened 2022-02-09
Update react-tagged-state

oleggrishechkin:


krausest:
Thanks - results are updated, but look pretty much unchanged.

oleggrishechkin:
Sorry, I expected better results especially in "select row". (I forceUpdate one (or two) rows directly, so it should be faster than other react examples and it is)

PR #1000, opened 2022-02-10
Upgrade ember

NullVoxPopuli:
cc @lifeart


lifeart (approved):

NullVoxPopuli:
I couldn't run the benchmark tool locally. Ewen after installing java.
Wound be great if there was a docker image we could use or something

krausest:
So far I didn't publish a docker image, but you can compile all frameworks in a docker container as described here:
https://github.com/krausest/js-framework-benchmark/wiki/Build-all-frameworks-with-docker

Currently I'm only using this container and I'm regularly rebuilding all frameworks in that container (at least for each chrome update).
The workflow looks like that (after the initial setup): `npm run docker-start` 
Change some source code or pull PRs, then: `npm run docker-rebuild keyed/ember` or `npm run docker-rebuild-ci keyed/ember` (for former runs npm install, the latter npm ci). This commands performs a clean up, fresh build and runs some sanity checks.
(The clean build might be a bit slow when optimizing a single framework, but it works nice for merging the PRs.) 
Does this approach work for you?

krausest:
Thanks - results are updated, but look pretty much unchanged.

NullVoxPopuli:
All good -- we're planning some rendering engine updates soon, and this makes _that_ update a bit easier. thanks!

PR #1003, opened 2022-02-13
Upgrade Sinuous 0.28.1

luwes:
This upgrades Sinuous to the latest version with a perf fix 🤞

https://github.com/luwes/sinuous/commit/2a4345ccf878a90648447d05318154fb4ddb321e


krausest:
Here we go (maybe a little bit faster):
Screenshot from 2022-02-14 21-40-42

PR #1008, opened 2022-02-26
Implemented Voby-keyed

fabiospampinato:
Hey there, I just published a new little framework, [`voby`](https://github.com/fabiospampinato/voby).

Here's the keyed implementation for it.

I'm very curious to see how it compares with the rest in the benchmark!

### DIFF
--- a/frameworks/keyed/voby/index.html
+++ b/frameworks/keyed/voby/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
+    <title>Voby-keyed</title>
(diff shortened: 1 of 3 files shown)


krausest:
I changed the row numbers to start with 1 instead of 0, which makes all tests pass.
Looks quite good!
Screenshot from 2022-03-04 21-47-12

fabiospampinato:
@krausest Awesome, thank you!

Quick question: will I see the framework on the official results page before Chrome 100 is released? Or is that updated only with each release of Chrome?

krausest:
Yes, the next official run is scheduled for chrome 100 (unless something other forces me to perform an official run like a major OS update)

PR #1009, opened 2022-03-03
Use mimbl@0.8.1.

mmichlin66:
Upgrading to Mimbl version 0.8.1


krausest:
Thanks, results are updated.
Screenshot from 2022-03-04 21-47-12

krausest:
(If you wonder why vanillajs got faster: I'm asking that myself too. Some wayland packages got an update today, maybe that update caused the difference. Doesn't look like the noise I'd expect.)

PR #1010, opened 2022-03-03
Update to Sling.js version 14.3.0

puckowski:
* Update to Sling.js version 14.3.0.


krausest:
Thanks, [results](https://rawgit.com/krausest/js-framework-benchmark/master/webdriver-ts-results/table.html) are updated.

PR #1011, opened 2022-03-06
Updated Voby to 0.2.0

fabiospampinato:
I implemented a few optimizations here and there, maybe they'll make a measurable difference in the benchmark 🤔

### DIFF
--- a/frameworks/keyed/voby/src/main.tsx
+++ b/frameworks/keyed/voby/src/main.tsx
@@ -145,7 +145,7 @@ const RowDynamic = ({ id, label, className, onSelect, onRemove }: { id: Observab
   </tr>
 );
 
-const RowTemplate = template ( RowDynamic );
+const RowTemplate = template ( RowDynamic, { recycle: true } );
 
 const App = (): JSX.Element => {
 
--- a/frameworks/keyed/voby/src/main.tsx
+++ b/frameworks/keyed/voby/src/main.tsx
@@ -175,8 +175,8 @@ const App = (): JSX.Element => {
           <For values={$data}>
             {( $datum: IDatum ) => {
               const {id, label, className} = $datum ();
-              const onSelect = () => select ( id );
-              const onRemove = () => remove ( id );
+              const onSelect = select.bind ( undefined, id );
+              const onRemove = remove.bind ( undefined, id );
               const props = {id, label, className, onSelect, onRemove};
               return RowTemplate ( props );
               // return RowDynamic ( props );
--- a/frameworks/keyed/voby/src/main.tsx
+++ b/frameworks/keyed/voby/src/main.tsx
@@ -133,7 +133,7 @@ const Button = ({ id, text, onClick }: { id: string, text: string, onClick: (( e
 );
 
 const RowDynamic = ({ id, label, className, onSelect, onRemove }: { id: ObservableMaybe<string>, label: ObservableMaybe<string>, className: ObservableMaybe<string>, onSelect: ObservableMaybe<(( event: MouseEvent ) => any)>, onRemove: ObservableMaybe<(( event: MouseEvent ) => any)> }): JSX.Element => (
-  <tr class={className}>
+  <tr className={className}>
     <td class="col-md-1">{id}</td>
     <td class="col-md-4">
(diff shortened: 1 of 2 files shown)


fabiospampinato:
v0.3.0 has a pretty amazing optimization, I've very excited to see that in the benchmark!

fabiospampinato:
From running the benchmark locally it looks like the benchmark doesn't quite measure this area much, so the optimization doesn't make much of a difference in the benchmark 😢 It'd be great if a "replace all rows" test but with 10k rows were added, it seems to be missing currently.

fabiospampinato:
Shipped another batch of tweaks and optimizations.

@krausest if possible it'd be awesome to get this merged before the results table is refreshed for Chrome 100 🙏

fabiospampinato:
Expanding on the issue of the benchmark being somewhat creation-biased:

- The "append" test is also mainly measuring creation at the end of the day, and that's much more expensive than the diffing.
- The "clear rows" test is kind of giving bad results if nodes are reused, because if they can be reused there's a bit more work to do to recycle them, while if you can just throw them away it would be faster, the thing is recycling nodes ends up giving you much better performance at runtime when it matters, but that's almost entirely not measured by this benchmark, except for the "replace 1000 rows" test.

Basically it'd be nice if the benchmark measured more things similar to runtime performance, like complex edits that happen without refreshing the browser in between.

fabiospampinato:
I'm getting some promising results locally with the last build. There were a handful of missing optimizations and some useless work that the previous builds were doing.

krausest:
Screenshot from 2022-03-13 12-55-12
A nice improvement!

krausest:
There one catch though. I think voby keeps the selected state on the row and not on the table. Thus I had to add note #800 .
Could you try to store the selected row in the model and then set the danger-class for the row that's equal to the selected row in the row render loop?

fabiospampinato:
Nice! Thanks for merging. 

There's at least one more little optimization that would be worth implementing. Other than that one I'm running out of ideas though.

> Could you try to store the selected row in the model and then set the danger-class for the row that's equal to the selected row in the row render loop?

Sure, I can do that 👍

frameworks/keyed/voby/src/main.tsx
```diff
@@ -100,25 +99,26 @@ const Model = (() => {
 
   const remove = ( id: string ): void => {
     const data = $data ();
-    const index = data.findIndex ( datum => datum.sample ().id === id );
-    $data ( [...data.slice ( 0, index ), ...data.slice ( index + 1 )] );
+    const index = data.findIndex ( datum => datum.id === id );
+    if ( index === -1 ) return;
+    $data ( data.slice ( 0, index ).concat ( data.slice ( index + 1 ) ) );
```
leeoniya, line 104: @fabiospampinato would the shorter `.splice()` be any slower?

fabiospampinato: @leeoniya I'm not sure, it might depend on where the deletion happens.

I went with slice/concat because I need to make a new array to have the observable detect it as a change so that it can notify all its subscribers. But now that I think about it it would probably be better to expose the method for notifying subscribers, so that I can just edit the array in place and notify them manually.

leeoniya: not sure the notification route would have any measurable effect in this specific case, but good to have as option in voby.

for this bench i'd probably leave it written as-is.

fabiospampinato: It might shave 1ms or so, but more importantly it would make the code cleaner.

leeoniya: > but more importantly it would make the code cleaner.

unless it's mutate & notify in some places, and shallow copy in other places :)

fabiospampinato: > unless it's mutate & notify in some places, and shallow copy in other places :)

A little messy, but cleaner than what the code looks like now. For the cleanest code when working with non-trivial data structures I think we ideally want to wrap the data structure in a proxy and extract observables out of it automatically, or something like that. I have another library for that but I haven't wired it with this one, it should be doable 🤔

PR #1012, opened 2022-03-14
Updated Voby to v0.7.0

fabiospampinato:
I added some more optimizations, cleaned up the code a bit, and now I'm storing the "selected" state in the model. 

I think performance might be slightly better, as N-1 no longer necessary observables have been deleted in the process, but the noise of measuring is probably higher than that anyway.

Fingers crossed now, unless there are some issues or decent optimizations missing, the framework is "done". I'll probably tweak it a little bit once I start using it and know where the pain points are though.

### DIFF
--- a/frameworks/keyed/voby/package.json
+++ b/frameworks/keyed/voby/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-voby",
-  "version": "0.6.0",
+  "version": "0.7.0",
   "main": "dist/main.js",
   "js-framework-benchmark": {
     "frameworkVersionFromPackage": "voby",
@@ -17,7 +17,7 @@
     "url": "https://github.com/krausest/js-framework-benchmark.git"
   },
   "dependencies": {
-    "voby": "0.6.0"
+    "voby": "0.7.0"
   },
   "devDependencies": {
     "esbuild": "0.14.23"
--- a/frameworks/keyed/voby/package.json
+++ b/frameworks/keyed/voby/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-voby",
-  "version": "0.5.0",
+  "version": "0.6.0",
   "main": "dist/main.js",
   "js-framework-benchmark": {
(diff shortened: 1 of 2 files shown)


krausest:
Thanks, especially for removing #800.


Both vanillajs and voby appear to be a bit slower than in the last run. Hard to tell if it's just noise or maybe just today's OS updates.
Maybe I find the time to do a full run for all frameworks.

leeoniya:
@fabiospampinato the memory metrics for Creating/Clearing 1k rows seem to be a bit of an outlier among the fast libs :thinking: 

image

fabiospampinato:
@krausest thanks for merging and the updated results. The numbers seem about right to me, at the end of the day there are too many factors to measures things super accurately. Maybe I have made things a bit slower with the new way the selected row is handled, could be.

@leeoniya yeah nodes are [recycled](https://github.com/fabiospampinato/js-framework-benchmark/blob/b45e99b29d167701f65efa8f0c84d59d2008f6c2/frameworks/keyed/voby/src/main.tsx#L132], that's a pretty awesome optimization but it benchmarks poorly here (a bit more time spent when making rows, and that's something the benchmark measures a lot, and a bit more time spent when removing nodes), on the other hand the amount of time it takes to replace rows is super close to the vanilla implementation, despite all the extra stuff that voby does, I don't know I think it's worth it as an optimization for what I need, I could try turning it off and see what the impact on the benchmark would be. Maybe you could try that.

fabiospampinato:
@leeoniya Just ran the benchmark locally with recycling on and off, with recycling off the memory usage that you see after clearing rows is more reasonable. Other than that the rest seems amount the same, but replacing rows is a bit slower.

Screen Shot 2022-03-15 at 13 41 11

leeoniya:
@fabiospampinato nice.

in real apps i would probably keep recycling off if the difference is this small. i would rather keep the GC happy and not have a lot of retained nodes for it to track that may never be re-used for the lifecycle of a long-running app. in extreme cases it may even resemble a memory leak unless there is some kind of max lifetime retention policy.

PR #1013, opened 2022-03-18
Updated Voby to v0.11.4

fabiospampinato:
I added a handful of optimizations, fixed a few things, and turned off recycling of rows since what that enables isn't something this benchmark measures much. I'd expect performance to be more or less the same, like nothing too drastic should be changing except for the "replace all rows" test and the "create/clear" memory test.


krausest:
Thanks - results are updated.  
Screenshot from 2022-03-22 20-19-03

PR #1014, opened 2022-03-24
Update react-tagged-state

oleggrishechkin:
just update to latest version.

performance should be slightly slower, I think


krausest:
Thanks - yes maybe a bit slower: 
Screenshot from 2022-03-26 11-48-37

PR #1016, opened 2022-03-28
implement benchmark for ui5-webcomponents

pskelin:
This PR adds a benchmark implementation with the UI5 Webcomponents framework
https://github.com/SAP/ui5-webcomponents


krausest:
Thanks. Seems like select row is a bit weak point of the framework (maybe due to the onBeforeRendering callback?) and it demonstrates that this benchmark doesn't handle frameworks with lengthy names well :-)

pskelin:
Thanks for the inclusion. I was surprised that out of all existing frameworks, this was the first with a long name, but the data in the table looks better with narrow columns.

I saw that select row was slower and was able to trace it to the usage of the static html function of lit-html, which is used for scoping (adding a tag name suffijx) and getting multiple instances of the same component running with different tag names.
`import { html } from 'lit/static-html.js';`

We'll optimize this in a future version and update the benchmark.

krausest:
github pages didn't update yesterday, but now ui5-webcomponents is included in https://krausest.github.io/js-framework-benchmark/current.html .
I added some css magic for automatic hyphenation (doesn't work on safari it seems).

PR #1017, opened 2022-03-28
Updated Voby to v0.14.0

fabiospampinato:
- v0.12 had faster clearing but it left some work to do in the future potentially, now clearing may be a bit slower but there's no leftover work or risks of memory leaks, though I had to resort to some manual disposition to gain some of that performance back.
- A fairly important bug got fixed, it just made things less efficient though, the framework still worked correctly as far as I can tell.
- The cache behind the crucial For component is more optimized.
- The framework is now tree shakeable, and its core is mangled. 

Basically the framework should do better on memory usage and startup tests, but it may be a bit slower.

Apparently today/tomorrow Chrome 100 will be released, I hope I submitted this in time for that.

### DIFF
--- a/frameworks/keyed/voby/package.json
+++ b/frameworks/keyed/voby/package.json
@@ -6,7 +6,7 @@
     "frameworkVersionFromPackage": "voby"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


krausest:


Now I'll try to move all frameworks to chrome 100!

krausest:
JFI: Chrome 100 is somehow unexpected:
https://github.com/krausest/js-framework-benchmark/issues/1020

PR #1018, opened 2022-03-29
lui: update to 1.2.3

L3P3:
It has changed a lot, I want to see how it performs now. :wink:
The main list bottleneck is not fixed but somewhat patched over. :see_no_evil:

I hope `script` `defer` attribute is allowed.


krausest:

L3P3:
Thanks. The swap rows number hurts but the older implementation was wrong and only worked out of coincidence for this benchmark — so I am happy for now! :D

PR #1019, opened 2022-03-30
Updated Voby to v0.16

fabiospampinato:
A few more optimizations here and there, but most importantly I've replaced the model implementation with a much cleaner one, it's basically copied from Solid.

I think it would be interesting to see both implementations in the benchmark potentially, the clean one and the hand-optimized one that still doesn't mess with the DOM manually or do other flagged stuff. @krausest if you think it'd be appropriate to have two implementations for this framework, a clean one and one that squeezes every possible drop of performance out of it, then I'll submit both, otherwise it's probably better to keep the clean one only in here.

Sorry for the many PRs in close succession I've submitted so far, this might be the last for a while.

### DIFF
--- a/frameworks/keyed/voby/package.json
+++ b/frameworks/keyed/voby/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-voby",
-  "version": "0.14.0",
+  "version": "0.16.0",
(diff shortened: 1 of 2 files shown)


krausest:
Thanks results look like that:

PR #1021, opened 2022-03-30
Update react to 18

oleggrishechkin:
In my local tests pure react18 solution faster than react17 and event faster than other react implementations,
React with some external state management slower (clear rows much slower, maybe useEffect affected in new react18, useSyncExternalStore has the same performance as forceUpdate with direct subscribing in useEffect)


krausest:
It seems like clear rows got slower (all measurements are above 100 msecs):

PR #1022, opened 2022-03-30
Added support for ForgoJS

spiffytech:
This PR adds support for [ForgoJS](https://forgojs.org/) at version 2.2.3.


krausest:
Thanks. Here are the results:

PR #1030, opened 2022-04-30
Update uhtml to version 3

WebReflection:
[µhtml](https://github.com/WebReflection/uhtml#readme) has landed with smaller size, modern code, and overall better performance, so I'd love to have it updated in here too, thank you 🙏

### DIFF
--- a/frameworks/keyed/uhtml/package-lock.json
+++ b/frameworks/keyed/uhtml/package-lock.json
@@ -10,45 +10,45 @@
       "license": "Apache-2.0",
       "dependencies": {
         "js-framework-benchmark-utils": "^0.3.2",
-        "uhtml": "^2.7.6"
+        "uhtml": "^3.0.0"
       },
       "devDependencies": {
-        "@rollup/plugin-node-resolve": "^13.0.0",
-        "@ungap/degap": "^0.2.7",
-        "rollup": "^2.52.6",
+        "@rollup/plugin-node-resolve": "^13.2.1",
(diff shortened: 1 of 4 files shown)


krausest:
Thanks for the PR. Here's a comparison (both runs from today with playwright) - and I'm afraid to say the benchmark shows (almost?) no difference:
Screenshot from 2022-04-30 21-00-21

WebReflection:
@krausest thanks for merging! The difference is bigger "*on my machine*" but the most noticeable is the startup metrics which, still "*on my machine*" is faster than solidjs. The size and logic has been simplified so it's faster overall, not just in this bench, yet it's weird your results are somehow different fro mine ... I'll post a screenshot tomorrow just for record sake.

WebReflection:
This is what I see via Chromium 101.0.4951.41 and the **chrome100** branch.

It's not like day and light, but the difference is definitively more relevant than your results.
Also visible the startup metrics improvements ... but the test itself feels visibly faster with v3.

uhtml3

krausest:
FWIW here's a screenshot from my MacBook Air:
Bildschirmfoto 2022-05-03 um 21 13 02
Bildschirmfoto 2022-05-03 um 21 15 43
Bildschirmfoto 2022-05-03 um 21 15 50

WebReflection:
well, startup metrics score same as vanilla + smallest amount of bytes is something 🥳

thanks for the heads-up, I'm not even disappointed by these results 👍

PR #1034, opened 2022-05-02
Updated Voby to v0.18

fabiospampinato:
A huuuuge amount of improvements have been made:

- The underlying algorithm for the reactivity system is much more similar to MobX's, which improves performance significantly under real scenarios.
- The exposed interface is significantly cleaner, a lot of redundant or not useful enough stuff got deleted.
- A basic Suspense implementation got implemented.
- A more optimized clearing algorithm got implemented.
- A decent amount of internal bugs has been fixed.

Unfortunately it seems that at least locally some of these optimizations and changes worsened performance in the benchmark, maybe the added overhead can be trimmed away, I haven't looked into that yet.

### DIFF
--- a/frameworks/keyed/voby/package-lock.json
+++ b/frameworks/keyed/voby/package-lock.json
@@ -1,25 +1,19 @@
 {
   "name": "js-framework-benchmark-voby",
-  "version": "0.16.0",
+  "version": "0.18.0",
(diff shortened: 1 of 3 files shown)


krausest:
Thanks. Here's a screenshot. Looks indeed a bit slower in **this** benchmark.

PR #1036, opened 2022-05-03
Update react-redux to 8.0.1

oleggrishechkin:
I just updated deps, but I think we should use @reduxjs/toolkit, because it's official "true" way of using redux. What do you think?


oleggrishechkin:
Thanks for your fast reaction and merge!

PR #1038, opened 2022-05-04
Updated Voby to v0.18.1

fabiospampinato:
With this update the framework should gain back some of the performance that was lost with v0.18.

Locally I'm seeing 1.11 -> 1.09 for the templated version (this version), and 1.25 -> 1.17 for the version that doesn't use `template` which seems pretty good to me.

### DIFF
--- a/frameworks/keyed/voby/package-lock.json
+++ b/frameworks/keyed/voby/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-voby",
-  "version": "0.18.0",
+  "version": "0.18.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "js-framework-benchmark-voby",
-      "version": "0.18.0",
+      "version": "0.18.1",
       "license": "MIT",
       "dependencies": {
-        "voby": "0.18.0"
+        "voby": "0.18.1"
       },
       "devDependencies": {
         "esbuild": "0.14.28"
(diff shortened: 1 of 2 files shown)


krausest:
Thanks. Looks good:

fabiospampinato:
Nice, thanks! That "clear rows" result looks suspiciously good compared to v0.18 🤔 I don't know if it's a lucky run or it was an unlucky run before. Anyway, thanks for merging!

krausest:
Hard to tell. Of course there were many OS updates and I think at least one chrome patch in between the two runs. So I ran it three times again:
````
result voby-v0.18.1-keyed_09_clear1k_x8.json min 51.685 max 59.93 mean 56.8636 median 58.0085 stddev 2.7964911029518547
result voby-v0.18.1-keyed_09_clear1k_x8.json min 51.1 max 60.121 mean 55.3722 median 55.0905 stddev 3.126652409349157
result voby-v0.18.1-keyed_09_clear1k_x8.json min 51.116 max 59.731 mean 57.05799999999999 median 59.0055 stddev 3.444565962014179

result voby-v0.18.0-keyed_09_clear1k_x8.json min 52.707 max 62.114 mean 58.50440000000001 median 58.972 stddev 3.253729456758472
result voby-v0.18.0-keyed_09_clear1k_x8.json min 54.01 max 62.052 mean 58.6212 median 60.2545 stddev 3.1351688170034975
result voby-v0.18.0-keyed_09_clear1k_x8.json min 54.353 max 60.221 mean 58.18849999999999 median 58.3455 stddev 1.6578949336767719
```
Maybe it really got a bit faster (esp. when looking at the min value + mean, median isn't that clear).

fabiospampinato:
Interesting, thanks, maybe it got a little faster indeed. Looking back at the commits I pushed maybe this one makes it so that a few effects are skipped so they don't need to be cleaned up, I'm not sure https://github.com/vobyjs/voby/commit/a24d90bb762c347eee866f431e90b634a924e8a9

PR #1040, opened 2022-05-17
update ui5-webcomponents to 1.3.1

pskelin:
Hi,
the original benchmark result had a sub optimal result on select row as discussed in #1016 

ui5-webcomponents provides an updated version with a fix which no longer uses lit-html static by default,
only when scoping enabled.

This PR is using the latest version of ui5-webcomponents.


krausest:
Thanks - nice improvement!

PR #1042, opened 2022-05-21
Update karyon to v1.0.10

syduki:
An update to the latest version of [karyon](https://www.npmjs.com/package/karyon) library.


krausest:
Thanks. Maybe a little bit faster:

PR #1043, opened 2022-05-23
Updated Voby to v0.21.1

fabiospampinato:
Some important internal changes and fixes. Nothing that seems to have any impact on the benchmark though. Though I'm seeing weird memory usage numbers being reported locally.

### DIFF
--- a/frameworks/keyed/voby/package-lock.json
+++ b/frameworks/keyed/voby/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-voby",
-  "version": "0.18.1",
+  "version": "0.21.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "js-framework-benchmark-voby",
-      "version": "0.18.1",
+      "version": "0.21.1",
       "license": "MIT",
       "dependencies": {
-        "voby": "0.18.1"
+        "voby": "0.21.1"
       },
       "devDependencies": {
         "esbuild": "0.14.28"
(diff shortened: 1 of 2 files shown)


krausest:
Thanks. Yes, results are unchanged (which is a good thing when a framework is that fast).

fabiospampinato:
I need to look again at that memory usage though:

Screen Shot 2022-05-24 at 20 33 53

This doesn't look right 🤔

fabiospampinato:
FWIW locally I'm seeing the same numbers reported by the devtools for memory usage both for v0.18.1 and v0.21.1, so I suppose memory how this is measured changed in the benchmark recently perhaps? Otherwise I'm not sure how else to explain the wild difference.

krausest:
Hi @fabiospampinato 
I went back to voby 0.16.0 and ran benchmark 25 again. Results look like that:
voby-v0.21.1-keyed_25_run-clear-memory 4.775737762451172
voby-v0.18.1-keyed_25_run-clear-memory 5.140129089355469 
voby-v0.18.0-keyed_25_run-clear-memory 5.457756042480469 
voby-v0.16.0-keyed_25_run-clear-memory 1.6423683166503906 

In the devtools I'm seeing a memory usage that's indeed higher than that of e.g. doohtml (reports ~2.2 MB)
Screenshot from 2022-05-26 09-08-20

I can't tell for sure that it's not a problem in the benchmark driver but currently it looks to me more like voby is really using  more memory since 0.18.0.

fabiospampinato:
"clear-memory" is the "clear rows" test, right?

I'm seeing this locally with the latest build (with no CSS):

Screen Shot 2022-05-26 at 12 14 26

- Snapshot 1 is a snapshot at startup, 1.1MB looks just normal to me.
- Snapshot 2  is a snapshot after created/clearing 5 times, 1.5MB seems ok to me, looking at the allocated objects between the two snapshots it's all stuff internal tot he browser basically.

Basically there's nothing to optimize here that I can see. The only explanation that I can think about is that either the framework is measuring the right thing but wrongly, somehow, (maybe no GC is triggered?), or it's just measuring the wrong thing (maybe peak memory usage or something?). I just see nothing here to optimize locally.

fabiospampinato:
It'd be useful to run the benchmark again and dump the heap after measuring memory usage, if memory usage says like 5MB but the heap only lists like 3MB of stuff in it then the measure is not right.

krausest:
I'm currently running chrome 102 on my linux machine, so I could only take a look at it on my MacBook.

However I'm getting a different picture for the scenario above (profile after startup, second profile after 5 cycles of create and clear):
Bildschirmfoto 2022-05-26 um 22 33 09

This is for voby 0.21.1. It shows lots of detached HTMLTableRowElements. I attached the heap dump 
[Heap-20220526T223315.heapsnapshot.zip](https://github.com/krausest/js-framework-benchmark/files/8781961/Heap-20220526T223315.heapsnapshot.zip) 

That's still far of the ~5 MB as reported by the benchmark driver, however I don't understand chrome here:
If I just open the memory tab, it displays after loading the page:
Bildschirmfoto 2022-05-26 um 22 37 52


Then clicking 5 times on create and the clear and running GC gives me:
Bildschirmfoto 2022-05-26 um 22 38 18

This is close to be what the benchmark driver reports too.

fabiospampinato:
It seems to me like Chrome 102 is buggy then, if it's consuming 4~10x the amount of memory that I'm seeing on Chrome 101, executing the same code 🤔  I've never seen those `Detached *` things either, maybe it's keeping them in memory for a while in case it will need them again in the future? It'd be interesting to see what the impact of this is on the other frameworks too once the table gets refreshed.

PR #1047, opened 2022-05-28
Updated Voby to v0.25

fabiospampinato:
Some internal fixes and new APIs, the main thing relevant to the benchmark I guess is that now the framework tree-shakes a little better. Nothing drastic though.

I hope this will be the last PR for a while, the future main ones should be about supporting stores, which may have a positive impact on clearing actually (and maybe a negative one on everything else), and about supporting automatic application of the "template" function via a bundler plugin, which could be kind of a big improvement, but irrelevant for the benchmark I think.

### DIFF
--- a/frameworks/keyed/voby/package.json
+++ b/frameworks/keyed/voby/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-voby",
-  "version": "0.27.2",
+  "version": "0.27.3",
   "main": "dist/main.js",
   "js-framework-benchmark": {
     "frameworkVersionFromPackage": "voby"
@@ -16,7 +16,7 @@
     "url": "https://github.com/krausest/js-framework-benchmark.git"
   },
   "dependencies": {
-    "voby": "0.27.2"
+    "voby": "0.27.3"
   },
   "devDependencies": {
     "esbuild": "0.14.39"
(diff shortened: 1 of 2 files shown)


fabiospampinato:
Just shipped some memory usage optimizations, they seem to have shaved ~5% with 10k rows on the page.

fabiospampinato:
I think I'll push another update to this tomorrow, a few more optimizations that I missed may be somewhat interesting.

fabiospampinato:
Just pushed another bunch of memory optimizations, they should be fairly significant all things considered. I'd be nice to see the updated table with v0.27.2 in it.

fabiospampinato:
Last (🤞) batch of changes. They don't seem to have much of an impact in the benchmark though.

fabiospampinato:
Just deleted **two** functions per row 😂 It's crazy, since the last PR with 10k rows on the page there are now 50k fewer functions in the heap (out of 80k).

fabiospampinato:
Tied up some loose ends, that'll be it it, for a while.

krausest:
Looks good!

Usually I forget to take a look at the memory results, but not this time. And it looks fine too.

fabiospampinato:
Nice! The cleanup time looks like it was a bit unlucky, locally I'm getting numbers close to Solid, but on average it seems about what I'm seeing locally.

PR #1049, opened 2022-06-02
Update to the same implementation as keyed/react-hooks

guojin-long:
image


guojin-long:
This update is for comparison with React 18.

TODO:
1. Optimize the insertion of dom nodes
2. Optimize the deletion of dom nodes

PR #1051, opened 2022-06-11
update keyed/jotai and keyed/valtio

dai-shi:
Should I update react/react-dom too?


krausest:
Thanks. As expected for small version bumps results didn't change much:
Screenshot from 2022-06-12 21-00-33
A PR for react would be welcome (but I'm on holidays for a week from tomorrow on, so it'll take a bit until it's merged).

PR #1052, opened 2022-06-11
Update alpine, hyperapp, preact, redom, solid, svelte, vue

alpacone:


krausest:
Thanks. As expected for small version bumps results didn't change much:
Screenshot from 2022-06-12 21-00-33

PR #1055, opened 2022-06-20
Updated uhtml to the patcehd 3.0.1

WebReflection:
The 3.0.0 version of *uhtml* had an [issue with partial content](https://github.com/WebReflection/uhtml/issues/64) so that every node, even single nodes, where created as document fragment and retained as such.

Version 3.0.1 patched that behavior, and it's been rock-solid/robust for long time now so that I think it's fair to have the latest, not leak and not overhead prone version of this module.

Thank you.

### DIFF
--- a/frameworks/keyed/uhtml/package-lock.json
+++ b/frameworks/keyed/uhtml/package-lock.json
@@ -10,12 +10,12 @@
       "license": "Apache-2.0",
       "dependencies": {
         "js-framework-benchmark-utils": "^0.3.2",
-        "uhtml": "^3.0.0"
+        "uhtml": "^3.0.1"
       },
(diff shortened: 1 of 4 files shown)


WebReflection:
Uhm ... indeed if the version is not exact any library patch would land next round of benchmark, is that correct? Not sure otherwise I fully understood what didn't work as intended 🤔

thanks for the merge though

krausest:
I suppose *if* I ran npm install with `"uhtml": "^3.0.0"` again that would happen. Before committing package-lock.json and switching to `npm ci` I used to remove the caret in all version strings in package.json to prevent that.
Unfortunately the update to the last node lts version required a `npm install`.

PR #1059, opened 2022-06-25
Update karyon to v1.1.3

syduki:
An update to the latest version of [karyon](https://www.npmjs.com/package/karyon) library.


krausest:
Thanks - it's included in the results now:

PR #1060, opened 2022-06-25
Add mahal framework into list

ujjwalguptaofficial:
Add [mahal framework](https://github.com/ujjwalguptaofficial/mahal) into list


krausest:
Thanks, but I'm afraid to say that mahal is not fully keyed.
```
npm run isKeyed -- --headless true keyed/mahal
mahal-v1.3.0-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
```

Can you take a look at that please?

ujjwalguptaofficial:
yeah ok, will have a look. Thanks.

ujjwalguptaofficial:
hey @krausest 

I am trying to understand - why swap rows need to removed one rows and removed one rows. Could you please help me to understand this ?

Right now, the rows are just updated with each other without removing or adding row.

ryansolid:
Keyed behavior would have the rows swap with each other. The actual `<tr>`s swap positions in the table. It is keyed in the sense that once realized the data is associated with its DOM nodes, and only those nodes as long as that portion of the template exists.

Non-keyed would just move the data to different DOM nodes if the data changed positions. This can have side effects with any persistent state in the DOM, including but not limited to CSS transitions, and web components. This article has more information: https://www.stefankrause.net/wp/?p=342

ujjwalguptaofficial:
@krausest i have fixed the swap rows issues and also tested locally by runing the command - `iskeyed`.

krausest:
Thanks. Looks good (I fixed one div that had a className instead of class)
Results are updated:
Bildschirmfoto 2022-07-03 um 18 12 08

PR #1061, opened 2022-06-28
Add dioxus implementation

erichulburd:
Hi, I've worked with some Dioxus contributors to implement the benchmark and it's ready for review.

I've run the suite locally, but didn't update the results, as I imagine there is a formal process for doing that on consistent hardware.

Do you prefer the commits be squashed before merge?

TODO:

* [x] remove errant results updates
* [x] remove non-keyed implementation


krausest:
Thanks. Results are updated. Here's a screenshot:
Bildschirmfoto 2022-07-03 um 18 12 08

PR #1063, opened 2022-07-13
Update react-rxjs to 18.2.0+0.9.8

voliva:
Upgrading react-rxjs from 0.6.0 to 0.9.8. A few things have been changed since then, and it should yield some better results.


krausest:
Thanks and sorry, results are a bit slower than for the old version:
Bildschirmfoto 2022-07-16 um 20 15 48

PR #1064, opened 2022-07-15
Add sauron

flosse:


frameworks/keyed/sauron/index.html
```diff
@@ -0,0 +1,16 @@
...
+    <meta charset="UTF-8">
+    <title>Sauron</title>
+    <link href="/css/currentStyle.css" rel="stylesheet"/>
+    <base href="bundled-dist/"></base>
+  </head>
+  <body>
+    <div id='main'></div>
+    <script type="module">
+      import init from './js-framework-benchmark-sauron.js';
+      init('./js-framework-benchmark-sauron_bg.wasm');
```
ivanceras, line 13: This could just be written as
`await init().catch(console.err)`

ivanceras (approved):
Thank you for doing this work. I always wanted to do it, but I was caught up with a lot of work involving rust code, and I had a hard time trying to get into typescript/javascript.

krausest:
Thanks - there seems to be an error in the implementation.
Click "create 1,000 rows", remove the first row, remove the then first row (id 2). The latter row is not removed. Can you please take a look at it?

ivanceras:
I tried to log the generated patches by enabling debug in sauron.

Add this into `Cargo.toml`
```toml
sauron = { version = "0.49.4", features = ["with-debug"] }
log = "0.4.17"
console_log = "0.2.0"
console_error_panic_hook = "0.1"
```

and in `src/lib.rs`
```rust
#[wasm_bindgen(start)]
pub fn start() {
    console_log::init_with_level(log::Level::Trace).unwrap();
    console_error_panic_hook::set_once();
    log::info!("loaded...");
    let document = window().unwrap().document().unwrap();
    let mount_el = document.query_selector("#main").unwrap().unwrap();
    Program::append_to_mount(App::new(), &mount_el);
}

```

I only see one big AppendChildren patch attached

[debug-patch.log](https://github.com/krausest/js-framework-benchmark/files/9125404/debug-patch.log)

ivanceras:
Ah sorry, I misread it. I miss to click on the remove button. I will update shortly

frameworks/keyed/sauron/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+        let noun = *NOUNS.choose(rng).unwrap();
+
+        let label = [adjective, colour, noun].join(" ");
+
+        Self { id, label }
+    }
+    fn view(&self, selected: bool) -> Node<Msg> {
+        let id = self.id;
+        node! {
+            <tr class={if selected { "danger" } else  { "" }}>
```
ivanceras, line 63: I noticed that the `key` attribute is not set in the code.
If there is no key specified, sauron will aggressively reuse similar nodes.
If there is a key specified, if the key differ the node will be always discarded.

I still have a hard time running the js-framework-benchmark myself, but I would say adding a key here would most likely solve the issue.

```html
            <tr class={if selected { "danger" } else  { "" }} key=id >
```

ivanceras (changes requested):

frameworks/keyed/sauron/src/lib.rs
krausest: Please let me know if I can help running the benchmark on your machine. Where did you get stucked?

flosse: thanks, it looks like this was the trick ... but I also updated sauron to `0.50.1`.. so I don't know what causes it ;-)

flosse: > Please let me know if I can help running the benchmark on your machine. Where did you get stucked?

I chrome is not installed on a usual place on my machine (NixOS) so this is the error:

```
Executing frameworks/keyed/sauron and benchmark 01_run1k failed: browserType.launch: Failed to launch chromium because executable doesn't exist at /usr/bin/google-chrome
```
So how can I set the path to a custom location?

ivanceras: The update to 0.50.1 should have no effect to the code. 
> `Executing frameworks/keyed/sauron and benchmark 01_run1k failed: browserType.launch: Failed to launch chromium because executable doesn't exist at /usr/bin/google-chrome`

It looks like you need to have google-chrome installed

krausest: There's an option for that 😄 :
`npm run bench -- --chromeBinary /bla/bla keyed/vanillajs` and `npm run isKeyed -- --chromeBinary /bla/bla keyed/vanillajs` (in directory webdriver-ts) should both work with custom chrome locations. Can you try that?

flosse: @krausest thanks a lot, now my CPU is sweating :laughing:

flosse: @ivanceras 

The `isKeyed` test fails with this:
```
Keyed test for swap failed. Swap must add the TRs that it removed, but there were 997 new nodes
sauron-v0.50.1-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and non-keyed for 'swap rows benchmark'
```

ivanceras: @flosse  I'm not sure what `isKeyed` test does? but what should be the expected behaviour?

ivanceras: I'm able to run the benchmark now, I'm taking a look.

krausest: The isKeyed test checks whether the implementation behaves keyed. As such replacing rows must create new dom nodes (and not reuse existing ones), removing a row must remove the row you clicked on (and not e.g. the last row) and swap must actually move the dom nodes (and not create new nodes). 

Seems like swapping rows in sauron added 997 new tr rows. You can read here a bit more about that: https://github.com/krausest/js-framework-benchmark/pull/1060#issuecomment-1168247794

ivanceras: keyed diffing in sauron is only designed from top to bottom matching, that's why this fails. It may take me awhile to try to include heuristic to start from bottom to top in the existing code.
Meanwhile @flosse, maybe we can move sauron to the non-keyed frameworks for now.

flosse: > Meanwhile @flosse, maybe we can move sauron to the non-keyed frameworks for now.

done!

krausest:
Thanks. Results are updated https://krausest.github.io/js-framework-benchmark/current.html

flosse:
Cool, thanks a lot!

ivanceras:
aw, sauron has a lot of red scores. However, this would serve good baseline for future optimizations.

PR #1065, opened 2022-07-17
update whatsup version & implementation

iminside:
Hi @krausest
Please take a look at this when you have time.


krausest:
Thanks. Here are the results:
Looks a bit slower, especially for clear rows, but maybe OS 12.5 or a recent chrome update caused a bit performance degradation since vanilla-js also ran a bit slower.

PR #1067, opened 2022-07-24
Update Sycamore - 0.7.1

birkskyum:


krausest:
Thanks. Here are the results:
Looks a bit slower, but maybe OS 12.5 or a recent chrome update caused a bit performance degradation since vanillajs also ran a bit slower.

PR #1068, opened 2022-07-27
expose webdriver-ts-results folder as static directory

ujjwalguptaofficial:
This PR add `webdriver-ts-results ` directory into list static directory in express server. Due to this changes not present - result page was not visible.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -7,6 +7,7 @@ const app = express()
 const port = 8080
 
 let frameworkDirectory  = path.join(__dirname, "..", "frameworks");
+let webDriverResultDirectory  = path.join(__dirname, "..", "webdriver-ts-results");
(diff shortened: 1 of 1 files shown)


krausest:
Thanks. This was an oversight when changing the driver, but I know why it happened.
I switched my workflow to build the frameworks on one machine (on linux it was a docker container) and run the benchmark from my laptop. Currently I'm using a hetzner VPS to build the frameworks (since not all frameworks build always on arm64 OSX and I'm not sure whether it's security-wise a good idea to install all those frameworks on my main working machine) and run the benchmarks from my laptop. I'm using a ssh tunnel for port 8080 to access the server. In such a setup I must open the results from the local directory and not from the web server.

PR #1069, opened 2022-07-27
Reflex benchmark

zouloux:
Hello :)
I made some benchmark for Reflex, a reactive lib with factory function I created recently.
Tell me if you have any question.


krausest:
Thanks for your PR. But I'm afraid the commit has 1,724 file changes, which seems a bit odd so can't merge it. Can you please make a PR for reflex only?

zouloux:
How crap it added another commit that I did after the PR ... Patching this out, thanks !

PR #1073, opened 2022-08-02
React 18 + Starbeam

NullVoxPopuli:
Demos React with a universal reactivity implementation, Starbeam. https://github.com/starbeamjs [docs](https://starbeamjs.com/)

Preliminary results with Starbeam 0.5.8:

image

### DIFF
--- a/frameworks/keyed/react-starbeam/.babelrc
+++ b/frameworks/keyed/react-starbeam/.babelrc
@@ -1,0 +1,21 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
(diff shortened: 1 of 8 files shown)


krausest:
There appears to be an issue with the implementation. If you load the page and click just "append 1,000 rows" nothing happens, but of course there should be 1,000 rows displayed exactly as for create 1,000 rows. The isKeyed check uses exactly that button and thus currently fails. Can you take a look at it?

NullVoxPopuli:
fixed! thanks!

krausest:
That worked thanks. I just read chrome 104 came out so I'm updating the results for chrome 104.

PR #1075, opened 2022-08-02
Add clarification about the state of starbeam

NullVoxPopuli:
Starbeam isn't quite ready for production yet, and I want to make that clear in the demo.

There is a bunch of debug tooling still in the build.


krausest:
Is included and results are published:
https://krausest.github.io/js-framework-benchmark/2022/table_chrome_104.0.5112.79.html

PR #1082, opened 2022-08-12
Upgrade starbeam in react starbeam

NullVoxPopuli:
- Upgrades starbeam deps
- uses more idiomatic React


krausest:
Bildschirmfoto 2022-08-13 um 22 14 52

PR #1086, opened 2022-08-26
update mahaljs to 1.4.3

ujjwalguptaofficial:
* update mahaljs to 1.4.3


krausest:
Thank you. With the updated config the index.html page tried to load http://localhost:8080/js/runtime.8bf36938d9012d6ced34.bundle.js?b83370a6094779c319a5 which isn't the correct path if the server is run from the js-framework-benchmark root dir. I tweaked the config such that the framework path is included. Maybe you find a better solution. Results will be updated soon.

krausest:
Thanks. Results are updated and are almost unchanged:

PR #1087, opened 2022-08-29
Update react-tagged-state

oleggrishechkin:


krausest:
Thanks. Results are updated and are almost unchanged:

PR #1088, opened 2022-08-29
add Diagon to keyed/frameworks

akutruff:
Adding Diagon state management for React.


krausest:
Thanks. Results are updated. Seems like Diagon performs similar to zustand:

akutruff:
Yep. Concurs with my results

PR #1091, opened 2022-08-29
upgrade Recoil to latest and React 18

akutruff:


krausest:
Thanks. Results are updated and it seems that the react update to v18 made is slower (see #1021)

PR #1094, opened 2022-09-01
Add legend-state benchmark

jmeistrich:
This adds benchmarks for legend-state. See https://legendapp.com/dev/state for more info.

One thing that's unusual about this library: when an array is updated and the length did not change, it short-circuits React and updates the existing elements in place. So although the arrays are very much keyed, the "swap" benchmark says it's not keyed because the list does not re-render and elements are updated in place. So I'm not sure whether you want to consider it keyed overall.

### DIFF
--- a/frameworks/non-keyed/legend-state/index.html
+++ b/frameworks/non-keyed/legend-state/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <title>Legend-State</title>
+  <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+  <div id='main'></div>
+  <script src='dist/main.js'></script>
+</body>
+</html>
--- a/frameworks/non-keyed/legend-state/package-lock.json
+++ b/frameworks/non-keyed/legend-state/package-lock.json
@@ -1,0 +1,12479 @@
+{
+  "name": "js-framework-benchmark-legend-state",
+  "version": "0.14.0",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "js-framework-benchmark-legend-state",
+      "version": "0.14.0",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@legendapp/state": "0.14.0",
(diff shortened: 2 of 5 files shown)


fabiospampinato:
I think according to this benchmark if the node didn't change in that test then the implementation is not considered keyed. There are some nuanced issues with non-keyed behavior, for instance the old node might be in the midst of a CSS transition, or it could be focused, or it could have been mutated outside of the framework, or other things. If the framework doesn't clean things up properly that can be an issue, though if it does that it's a pretty nice optimization where applicable. For the purposes of the benchmark that's just considered an invalid implementation though.

jmeistrich:
It's still re-rendering the node with React as usual, so I don't think any of those issues would apply. But the item component is re-rendering itself rather than the parent re-rendering the whole list, so the node does change as would be expected. But it looks like `isKeyedSwapRow` checks for elements being added/removed, which is not happening in this case because the Row component updates itself.

krausest:
Thanks. Results are updated. 
I'm sorry: Swapping rows must move the actual dom nodes for a keyed implementation (otherwise state on the dom node wouldn't be swapped). 
Non-keyed implementations are allowed to "optimize" the process of updating the list (I'm using quotations marks since most keyed implementations can behave non-keyed by using the index as the key. The opposite is often not true).

jmeistrich:
Thanks! Can I just ask what you mean that state on the dom node wouldn't be swapped? It's not doing any manipulation outside of React - it still re-renders the two swapped elements normally with React, but it just doesn't need to re-render the parent List to do it. So it wouldn't lose any dom state unless the dom was modified outside of React.

I totally understand if it needs to be considered non-keyed - just trying to understand what the metric is, and maybe we can improve the implementation.

ryansolid:
The concern in this article and linked article would still be present: https://www.stefankrause.net/wp/?p=342

The reason you want non-keyed is so that state that is stored in the DOM stays with the model. Now we try our best not to make the DOM stateful but it is sometimes, there are more examples of this than just the ones in the article, especially considering things like web components. But just look at the animation example and think about what would happen.

krausest:
Yes, it's fine to submit both a keyed and a non-keyed implementation.

jmeistrich:
Great, thanks @krausest :)

PR #1095, opened 2022-09-05
React 18 with concurrent features

MrFlashAccount:
Hi! In this PR, I implemented time slicing for `React` by wrapping dispatch in `startTransition`.

### DIFF
--- a/frameworks/keyed/react-hooks/src/main.jsx
+++ b/frameworks/keyed/react-hooks/src/main.jsx
@@ -1,4 +1,4 @@
-import { memo, useReducer } from 'react';
+import { memo, useCallback, useReducer, useTransition } from 'react';
 import { createRoot } from 'react-dom/client';
 
 const random = (max) => Math.round(Math.random() * 1000) % max;
@@ -107,14 +107,16 @@ const Jumbotron = memo(({ dispatch }) => (
 
 const Main = () => {
   const [{ data, selected }, dispatch] = useReducer(listReducer, initialState);
+  const [, startTransition] = useTransition();
+  const transitionedDispatch = useCallback((...args) => startTransition(() => dispatch(...args)), []);
 
   return (<div className="container">
-    <Jumbotron dispatch={dispatch} />
+    <Jumbotron dispatch={transitionedDispatch} />
     <table className="table table-hover table-striped test-data">
       <tbody>
-        {data.map(item => (
-          <Row key={item.id} item={item} selected={selected === item.id} dispatch={dispatch} />
-        ))}
+      {data.map(item => (
+        <Row key={item.id} item={item} selected={selected === item.id} dispatch={transitionedDispatch} />
+      ))}
       </tbody>
     </table>
     <span className="preloadicon glyphicon glyphicon-remove" aria-hidden="true" />


krausest:
Thanks. Looks like concurrent mode is slower than the standard implementation. Thus I reverted the react-hooks folder and created a new folder react-hooks-use-transition (and maybe react-hooks-concurrent would have been a better name):
Bildschirmfoto 2022-09-12 um 21 59 48

MrFlashAccount:
@krausest by the total time - for sure, but by TBT the implementation takes 15-20% overhead compared with solid.js and the UI remains responsive after the click. So I doubt these numbers could represent what I described above.

fabiospampinato:
> but by TBT the implementation takes 15-20% overhead compared with solid.js

How much total blocking time does the regular React implementation have?

Btw how are you measuring that?

> and the UI remains responsive after the click

Do you see it being smoother than Solid? If I recall correctly even making 10k rows takes like 90ms on my computer, most of the time is spent doing layout/style recalculations, which concurrent features shouldn't have much to say about I think 🤔

MrFlashAccount:
Yep, since React commits the changes in another frame, the ui is responsive between the ‘beginning’ and ‘committing’ phases. To be clear: in this test (creating 10k rows) we’re deeply limited by the browser. Still we can compare things to each other and I’m guessing that the current measurement (like total time) doesn't represent time slicing correctly. I think adding smth like ‘overall responsiveness rate’ would help.

MrFlashAccount:
Solid makes everything in the same frame, so the button stays pressed and the main frame locked during the operation

krausest:
Would be interesting to add, but how can we measure that? Is there something in the chrome timeline that can be interpreted?

MrFlashAccount:
It's a tough question 🥲

I think we can start with ‘TBT’.

Or

‘How long it takes to render the next frame after the interaction’ - I’m not sure we can extract this from the timeline itself, but we can measure it in userland.

fabiospampinato:
> Yep, since React commits the changes in another frame, the ui is responsive between the ‘beginning’ and ‘committing’ phases.

> Solid makes everything in the same frame, so the button stays pressed and the main frame locked during the operation

I see 🤔

It's interesting because as far as I can tell for every test other than the 10k rows one, accounting for the fact that some of these tests have artificial slowdowns, all the main-thread work should be doable within 50ms or so, sometimes much less than that, which should fall below the threshold where users start to perceive things as not immediate. 

So like there's necessarily less work happening, and the result is instant as far as the user is concerned, so I don't really see any benefit, at least for this benchmark, from concurrent features, if you have the resources to implement things more efficiently to begin with.

The 10k rows test may tell a different story, though in practice updating 10k rows is meaningless, they can't even all fit on the screen in a reasonable way, and I'm not sure if seeing the :active state on the button lifted but rows appear later on screen is necessarily a better experience. I should probably test what ReactConcurrent's approach _feels_ like for myself though, I've haven't yet seen a use case where given the time to optimize things ReactConcurrent seems the best approach.

ryansolid:
I want to add, Time Slicing helps zero with expensive DOM operations. So like if the expense is dropping 10k rows, both Solid and React are going to be creating all 10k rows and dropping them in. The time between the beginning and the commit phase in this benchmark is basically generating the VDOM nodes for those 10k rows. In Solid it is creating the effect per row. But almost all the time is spent actually creating the DOM nodes and inserting them during commit.

I'd be very surprised if we see any benefit to time slicing in any of these tests. The operation during the pure computation needs to be expensive and distributed. I think only partial update (maybe, select row) could theoretically fall into the category and they are already so quick. I doubt without artificial slowdown, and I don't mean lighthouse, I mean like locking the thread in a for loop you are going to see any benefit of time-slicing in a synthetic environment. The benefit comes here when the application code has grown to a complexity that you won't find in simple demos.

If we ever find that I'd love to enter Solid's version of time slicing in to see how it performs.

PR #1096, opened 2022-09-06
Update Solid to 1.5, rename to Solid Store

ryansolid:
I actually tried to do a little performance work for this release. A very small optimization for Select Rows. Although it looks like something is weird with remove rows in the current version of chromes results so I don't know if I'm just tempting fate here.

In any case also renamed Solid State -> Solid Store.. long overdue as we renamed that a year ago with 1.0 release.


krausest:
Thanks. I'd say performance is about the same as before:

ryansolid:
Something really strange is up with remove row in Chrome 105. I noticed this on the previous results as well and was hoping it was just a run blip. Solid store/state just use signals under the hood so here is no really tangible reason for these to differ. If anything Solid Store should be equal or slower than Solid in every benchmark. 

And it looks like many of the frameworks are affected. Inferno, Voby and others.

krausest:
Variance was bad for solid. I'll check if 2x/4x CPU throttling makes results more stable. 
Bildschirmfoto 2022-09-13 um 12 06 37

ryansolid:
If you compare Chrome 104 to 105 you will a number of frameworks that took a plummet randomly on remove Row.. Inferno, Voby, a number of others. Not really sure what is up.

krausest:
I got my new laptop (MBP 14, 32 GB, 8 CPU-Cores) and ran the benchmark on it:
Overall the new laptop is slightly slower 🙂, but it looks like variance is much better for remove rows. 
I'll try to run with 4x throttling this evening and then we can decide which result is more useful.

PR #1097, opened 2022-09-07
Preact Signals Implementation

ryansolid:
Still figuring out if this is the best way of accomplishing this. Tried a few different versions but they have different tradeoffs. I also think I may have stumbled on a memory leak. Wouldn't mind a review.

@developit @marvinhagemeister

Thanks,

### DIFF
--- a/frameworks/keyed/preact-signals/index.html
+++ b/frameworks/keyed/preact-signals/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>preact</title>
+  <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+  <div id='main'></div>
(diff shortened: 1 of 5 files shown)


krausest:
Thanks. Here are the results:

ryansolid:
I mean it's out there now. I marked it as a draft as I was hoping Jason would have time to address the memory leak first.

krausest:
Sorry, didn't see that. Seems like I was too much in a hurry. Please let me know if I should remove it from the master branch.

developit:
I'll try to get the fix PR'd today so we can update this benchmark. Would be nice to hold off on results and comparisons until that is done though, the current numbers are effectively meaningless.

krausest:
Sorry, my fault. I'll delete the implementation from master and update the results and you'd post a new PR when it's ready.

ryansolid:
The memory leak seems fixed in 1.0.4. Are we good @developit?

PR #1099, opened 2022-09-07
Update react-tagged-state to latest

oleggrishechkin:
I improved react integration is react-tagged-state.
Also, I swap to jotai implementation


krausest:
Thanks. Looks indeed a bit faster:

PR #1100, opened 2022-09-09
updated: Riot.js version

GianlucaGuarini:


krausest:
Thanks. Looks like riot 7 is a bit slower. I bumped the @riotjs/webpack-loader in the non-keyed implementation since it didn't build otherwise.

PR #1102, opened 2022-09-14
Update ef.js to v0.15.4

ClassicOldSong:
Performance boosted a bit on v0.15.2, so worth an update


krausest:

PR #1103, opened 2022-09-15
Updated Voby to v0.42.1

fabiospampinato:
Update after a few releases. Mainly there was a decent optimization I had missed.

### DIFF
--- a/frameworks/keyed/voby/package-lock.json
+++ b/frameworks/keyed/voby/package-lock.json
@@ -1,24 +1,24 @@
 {
   "name": "js-framework-benchmark-voby",
-  "version": "0.37.0",
+  "version": "0.42.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "js-framework-benchmark-voby",
-      "version": "0.37.0",
+      "version": "0.42.1",
       "license": "MIT",
       "dependencies": {
-        "voby": "0.37.0"
+        "voby": "0.42.1"
       },
       "devDependencies": {
-        "esbuild": "0.15.1"
+        "esbuild": "0.15.7"
       }
     },
     "node_modules/esbuild": {
-      "version": "0.15.1",
-      "resolved": "https://registry.npmjs.org/esbuild/-/esbuild-0.15.1.tgz",
-      "integrity": "sha512-zgxo2st9wSbdiR6rTo44l/L7ohttqdXFmhUi5tE6yWahgdBjCwZjBgIkm/gr/TcBTTIwyzd7em8WI37yZ+F2Mg==",
+      "version": "0.15.7",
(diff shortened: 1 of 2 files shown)


fabiospampinato:
I just added some last-minute optimizations. This new release overall might make a dent in the benchmark 🤔

krausest:
Thanks. I'm afraid the optimizations don't show in the results. Here's a screenshot with a comparison:

fabiospampinato:
Interesting, most numbers went down but vanilla's went down more 🤔 Did the computer or the version of the browser change?

In theory in v0.42 creating observables should be 20x faster and some splitting/mapping logic for the class should be skipped for every row basically 🤔 maybe there are other factors at play I'm not considering. Locally I got relatively different results on different runs. It'd be interesting to estimate how much results vary between runs.

fabiospampinato:
Like according to the old table Voby was 0.03 slower than Solid. In the new table they have the same mean value, but Solid didn't get refreshed. I'm not sure how to interpret these results, maybe they optimized something in the browser or something.

krausest:
The non official runs are always incremental (only the updated implementations + vanillajs as a check if the system looks ok). Thus it's possible that the minor chrome version or os version is different for those incremental runs.
Further if you select only some frameworks all factors are calculated only between the selected ones. Maybe I should stop doing it since vanillajs isn't fastest in all benchmarks and thus numbers in the complete table look different:
Are the comparison screenshots helpful or confusing?

fabiospampinato:
> Thus it's possible that the minor chrome version or os version is different for those incremental runs.

Makes sense 👍 

> Maybe I should stop doing it since vanillajs isn't fastest in all benchmarks and thus numbers in the complete table look different:

Yeah I don't know, that's a little bit confusing but probably fine. Maybe not doing that has the advantage that for example if I'm comparing like Svelte vs React the Svelte numbers would be largely useless to me, because I'd just see 1.00 in most of them, if instead I always saw numbers compared to the virtual best implementation the column for Svelte would give me useful information too. Maybe that's an argument for keeping those numbers consistent in filtered and unfiltered tables?

> Are the comparison screenshots helpful or confusing?

In general helpful I'd say, especially as not everyone runs the benchmark locally and it's cool to get a sort of personalized update, but sometimes it's hard to interpret them, especially if the percentage change in the framework is comparable to the percentage change that one may experience between runs naturally. I think in general if variance could be dropped significantly so that results are rock stable that would be nice.

krausest:
Official results for chrome 106 are out with the old logic (factors are computed between selected frameworks).
Additionally I created a new version where factors are computed against all other frameworks no matter if they are selected (though the frameworks are still filtered according to the issues checkboxes): https://krausest.github.io/js-framework-benchmark/current.html

PR #1107, opened 2022-09-21
Upgrade to latest version of Svelte

benmccann:
Major performance increase during SSR: https://github.com/sveltejs/svelte/blob/master/CHANGELOG.md#3490


krausest:
Thanks. Most of the time small version bumps show no differences like in this case. Here's a screenshot with a comparison:

PR #1110, opened 2022-09-22
Update karyon to v1.4.3

syduki:
Updating to the latest version of library, as some micro-optimizations were done since the last update. Also, a minor refactoring of implementation which could save a few bytes of memory on create rows.


krausest:

PR #1114, opened 2022-10-01
Doz v5 preview

fabioricali:
Upgrade from v3 to v5


krausest:
Thanks. I changed to version name to a fixed string, since the git url broke my naming scheme...
Here are the results:
Bildschirmfoto 2022-10-12 um 18 02 55

PR #1119, opened 2022-10-16
Correct mentions of sycamore in leptos directory and update leptos

gbj:
My original PR was based on the Sycamore directory structure and accidentally left a couple of mentions of Sycamore, including the module preloads. Since I also have a more recent version of Leptos out I think I'd correct those preloads and update at the same time.


krausest:
Here's the comparison:
Bildschirmfoto 2022-10-22 um 21 48 46

PR #1124, opened 2022-10-23
v0.80.1-keyed

hman61:
Improved add rows


krausest:
Thanks, looks like it's probably a bit faster. Here's the comparison:
Bildschirmfoto 2022-10-24 um 21 09 10

The non-keyed implementation returns an error for the isKeyed check since it's actually also keyed. I decided to merge it anyways (who cares about the correctness of non-keyed implementations 😄 ), but maybe we should consider removing the non-keyed implementation. Replace all rows and swap rows show the true nature and are slower than non-keyed implementations.

PR #1126, opened 2022-10-26
DooHTML-non-keyed-v0.60.11

hman61:


krausest:
Bildschirm­foto 2022-10-29 um 21 25 49
Please note that this was an incremental run, only updated frameworks and vanillajs were run. And at least vanillajs had a pretty bad run. It's currently hard to say if this is due to the recent chrome bugfix in 107 or just a bad run.

PR #1127, opened 2022-10-26
udomsay - a vDOM-less JSX & signals based library

WebReflection:
While I am still working on some possible performance improvements I am already pretty happy about the current [udomsay](https://github.com/WebReflection/udomsay#readme) result and I'd love to see it as part of the table, where the goal is to compete with other JSX based implementations in terms of memory consumption, code size and, after all, raw performance.

Thank you in advance for considering this addiction.

udomsay

### DIFF
--- a/frameworks/keyed/udomsay/index.html
+++ b/frameworks/keyed/udomsay/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
+    <title>udomsay-keyed</title>
+    <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+<div id='main'></div>
+<script src='dist/main.js'></script>
+</body>
+</html>
--- a/frameworks/keyed/udomsay/package-lock.json
+++ b/frameworks/keyed/udomsay/package-lock.json
@@ -1,0 +1,2313 @@
+{
+    "name": "js-framework-benchmark-udomsay",
+    "version": "1.0.0",
+    "lockfileVersion": 2,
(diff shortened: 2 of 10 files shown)


WebReflection:
P.S. if curious, this is the current result locally.

udomsay-v0 1 5

WebReflection:
P.S. I've made some change to allow non-keyed benchmark to run as well.

WebReflection:
FYI - I've flagged the module as 772 because I believe taking the signal off the whole App and using classList to add/remove danger is 772 ... even if the code is still idiomatic as it uses a signal to update the view (off main logic, like solid-js does) ... although it doesn't use an ad-hoc utility, it just effect a part from the `<App />`.

I am not planning to make any other change now so please consider running and merging this version.

Happy to change anything that needs changes though.

WebReflection:
updated the non-keyed too 😅

krausest:
Thanks. Here's the comparison and it looks much better for select row than yours (did you change something about it?):
Bildschirm­foto 2022-10-29 um 21 25 49
Please note that this was an incremental run, only updated frameworks and vanillajs were run. And at least vanillajs had a pretty bad run. It's currently hard to say if this is due to the recent chrome bugfix in 107 or just a bad run.

WebReflection:
@krausest I've epxlained in here: https://github.com/krausest/js-framework-benchmark/pull/1127#issuecomment-1293249167

I've change the logic and added 772 as *issue* because I felt stupid not using off-logic signal to handle the single selected row like solid-js does (or others) but there's no ad-hoc utility which is why 772 is flagged as issue ... it's all good, your run looks fine to me.

PR #1129, opened 2022-10-29
DooHTML-v0.80.2-keyed

hman61:
Minor improvement to add() method


krausest:
Hi @hman61, seems like swap rows doesn't work for the non-keyed implementation. There's an error on the console:
```
Main.class.js:129 Uncaught TypeError: this.updateRow is not a function
    at Main.swapRows (Main.class.js:129:10)
    at HTMLDivElement.<anonymous> (Main.class.js:156:11)
```
Can you take a look at it please?

hman61:
It should be ok now

krausest:
Thanks. Looks better - results will be updated tomorrow.

PR #1130, opened 2022-11-01
Updated Voby to v0.43.6

fabiospampinato:
Lots of internal tweaks, if I haven't messed anything up, which I'm not sure about, they might amount to something.

### DIFF
--- a/frameworks/keyed/voby/package-lock.json
+++ b/frameworks/keyed/voby/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-voby",
-  "version": "0.43.6",
+  "version": "0.43.7",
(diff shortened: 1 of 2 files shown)


fabiospampinato:
I'll push another update later today, I've found other optimizations 🤣

fabiospampinato:
I've added just one extra optimizations, others didn't make enough of a difference.

krausest:
Thanks, here's the comparison for the incremental update. Numbers are pretty close so it's hard to say whether it's actually faster:

fabiospampinato:
Thank you!

Yeah there isn't really that much room for improvement left for this benchmark anymore, there are only very minor tweaks left I think.

PR #1132, opened 2022-11-06
udomsay v0.3.4

WebReflection:
Well, this is my better take at this benchmark, after quite some work on internal logic for udomsay, where:

  * the current benchmark works but it's basically broken as the library wasn't effecting as expected with the partial rows update test, moving each time the whole App reload for no reason, and scoring accordingly (poorly)
  * instead of using hacks around `textContent` props, the JSX now uses the signal as is in the place it belongs to: the text part f the `<a />` element. The library now automatically creates "*atomic effects*" for signals used live on the DOM, granting updates in loco, resulting in more natural JSX representation without abusing props names to use as direct accessors
  * because usignal was broken before, the heap usage was also misleading ... now it's more aligned with real-world scenarios (it's higher, unfortunately, but it is what it is)

Thanks for considering this update so that the table will better reflect *udomsay* potentials out there.

### DIFF
--- a/frameworks/keyed/udomsay/package-lock.json
+++ b/frameworks/keyed/udomsay/package-lock.json
@@ -9,7 +9,7 @@
             "version": "1.0.0",
             "license": "ISC",
             "dependencies": {
-                "udomsay": "0.1.7"
+                "udomsay": "0.3.4"
             },
             "devDependencies": {
                 "@babel/cli": "^7.19.3",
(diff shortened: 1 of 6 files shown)


krausest:
Thanks, here's the comparison for the incremental update. The difference looks big enough such that I'd say it indeed got faster:

WebReflection:
red arrow shows where it was before, and where it is now: thanks for the update ❤️

Screenshot from 2022-11-07 22-50-24
Screenshot from 2022-11-07 22-50-54

PR #1134, opened 2022-11-10
add sledgehammer

Demonthos:
Adds [sledgehammer](https://github.com/Demonthos/sledgehammer/) to the benchmark. The library is most similar to wasm-bindgen so I added the 772 note to mark manual DOM manipulations.


krausest:
Thank - not too bad 😄, but that #772 is definitely correct. 
Bildschirm­foto 2022-11-13 um 21 08 18

PR #1136, opened 2022-11-14
Updated Voby to v0.43.8

fabiospampinato:
I found a couple of optimizations, I think they might have a small impact especially in the "append rows" test.


krausest:
I'm afraid the benchmark can't show the effect of the optimisations.
Bildschirm­foto 2022-11-20 um 22 32 38

fabiospampinato:
It's alright, these changes are too small to be measured reliably by the benchmark, like with each different run numbers change a little bit. It might have a small impact on average over the course of multiple runs.

Thanks for merging.

PR #1137, opened 2022-11-18
Update Riot.js to v7.1

GianlucaGuarini:
I thought that [`replaceChildren`](https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceChildren) was a better alternative to while loops. I was wrong :)


krausest:
What a difference!
Bildschirm­foto 2022-11-20 um 22 32 38

PR #1143, opened 2022-11-25
update `karyon` framework

syduki:
An update to the latest version of `karyon` module, including some revises of benchmark package.


krausest:
A little bit slower in this run, but very close to the precision this benchmark can offer.
Bildschirm­foto 2022-11-28 um 07 36 49

PR #1147, opened 2022-12-08
Update xania's implementation

xania:


xania:
@krausest I'v removed the flag 772 should not be applicable anymore

krausest:
Thanks. It's a bit hard to come to a conclusion. I ran the benchmark twice. Both times vanillajs was slower than in the last official run, while xania didn't change much. I think we have to wait for a full run to see the impact.

Bildschirm­foto 2022-12-19 um 22 44 50

xania:
I also noticed that vanillajs was slower on my machine most of the time so I am not surprised. I will look closely at the vanillajs implementations when official results are published.

"while xania didn't change much" is exactly the goal, optimization is ideally mostly a concern of the view library and less of the user code.
There is in fact a lot changed in the internals of _Xania_, the bit that pushed _Xania_ slightly higher than previous is that it access the dom less on the hot path.

xania:
I noticed on my machine that the results are not always the same even if I close everthing else when running the benchmarks, but I am not surprised to see _Xania_ has moved a bit closer to vanillajs :)

PR #1150, opened 2022-12-27
add maverick to keyed bench

mihar-22:
👋

This PR adds [Maverick](https://github.com/maverick-js/maverick) to the keyed bench. I've tested locally and the only issue is at the very end of the bench it logs some `package-lock` errors. Is that expected or am I missing something? Just noting that I can still successfully build the results table and view it.


krausest:
That package-lock message was due to a little typo.
Looks quite good:
Bildschirm­foto 2022-12-27 um 20 53 03

krausest:
I mistakenly made the first run without memory and startup results, thus I performed another run. So results are bit different compared to the screenshot above.

mihar-22:
Awesome, thanks for pulling that in :) 

Numbers seem okay for now. There's a lot of areas to still revise/improve so I think we'll see some jumps forward over time.

PR #1151, opened 2022-12-28
update maverick to 0.22.2

mihar-22:
Sorry to be a pain @krausest, just a quick bump as I've resolved some small issues with scheduling and the select row bench. Numbers should be stable now until I revise everything in a month or two.

Note: Turns out I was also creating thousand/s of effects for no reason -_- lol. Fixed that too.


krausest:
Thanks. Results are updated and here's the comparison:
Bildschirm­foto 2022-12-29 um 22 26 34

PR #1152, opened 2023-01-02
update maverick to 0.23.1

mihar-22:
Overall minor improvements across the board + big memory improvements.


krausest:
Thanks. Results are updated, a bit faster though pretty close to the old version. Here's the comparison:

Memory usage has gone done for run 10k has gone down significantly!
Bildschirm­foto 2023-01-14 um 12 02 29

PR #1156, opened 2023-01-07
update ember to v4.9.3

vlascik:


krausest:
Thanks. Results are updated, hard to say if it's faster or just noise in the measurement. Here's the comparison:

vlascik:
It's probably just noise, nothing changed in the framework's core.

However the current ember implementation is fairly bad, and can be made faster (2.26x vanilla -> 1.8x vanilla). I'm not sure though, if some of the changes (e.g. a single click event listener on table instead of event listeners on each row, or not using the separate Row component) would constitute "cheating" (some frameworks do it, some don't).

Should I create a PR for it, and discuss it there?

krausest:
That would be fine. Ember is big enough that I could justify it. But I'd prefer to have a second version with those optimisations (e.g. ember-optimised - and I'd add only a keyed version).
Those optimisations will cause discussions. It would be good if it still stayed close to everyday ember. We'll have to add the appropriate issue flags for the implementation (and I'd prefer not to have manual dom operations since we're then converging to vanillajs).

fabiospampinato:
> (e.g. a single click event listener on table instead of event listeners on each row, or not using a separate Row component)

That should barely move the needle, I think.

vlascik:
Hm, what I think would make even more sense here, would be to create separate benchmarking scenarios for these cases - "Create 1000 (10K) inline rows" and "Create 1000 (10K) Row components".

Component instantiation is quite a heavy operation for some frameworks, while having no cost in others, and not testing both approaches creates imprecise results - e.g. Angular bench uses inline rows, Ember uses Row component, and while it looks Angular is faster, in reality they are about the same when using the same approach. Or, another example, Ember with inline rows looks as fast as React with Row components, but in real apps, you might be forced into using Row components anyway, so the results saying Ember is as fast as React do not represent the reality.

Both approaches get used in real life, so it would make a lot of sense to show the differences in framework performance while doing both. And would put all frameworks on a level playing field...

PR #1158, opened 2023-01-10
Update Mimbl with version 0.10.1.

mmichlin66:


krausest:
Thanks. The incorrect #1139 flag was removed. Here's the comparison:
Seems like append to 10k rows actually got slower in this version.

mmichlin66:
Thank you so much!

Michael

PR #1166, opened 2023-01-19
Add Silkenweb framework

simon-bourne:
This adds the [Silkenweb](https://github.com/silkenweb/silkenweb) Rust/WASM framework.


krausest:
Thanks. You contribution looks good (passes isKeyed check etc.), but there's one thing that I must ask you to change: Could you please rename the dist folder to bundled-dist?
The dist folder gets deleted during the build process for all frameworks. Could you please change that folder name (I suppose this should be configured in Cargo.toml or so. Just renaming the directory is not sufficient).
Nevertheless here's a preview of the performance (which is quite good):

simon-bourne:
No problem. I've renamed that to "bundled-dist" and rebuilt. Thanks for the preview!

krausest:
Thanks. It's added now. Here's a comparison just to get a feeling how much noise to expect in the measurement:
Bildschirm­foto 2023-01-24 um 07 53 54

PR #1167, opened 2023-01-20
Update whatsup to v2.5.0

iminside:
Hi @krausest
Please take a look at this when you have time.


krausest:
Thanks. Looks better for most benchmarks except select row:

PR #1168, opened 2023-01-21
Update leptos to v0.1.1

flosse:


krausest:
Thanks. Here's the comparison. Seems to be slower for remove row and select row:

gbj:
@flosse @krausest I’m sorry I didn’t see this PR until I took a look at the new Chrome 110 results. I’d been holding off on updating the benchmark to 0.1.x while we finished some optimizations, and unfortunately this PR edits my implementation in particular by using a different and much slower primitive for “select row,” hence the large performance regression. I’ll edit and resubmit when I have some time.

PR #1171, opened 2023-02-02
Add mogwai

schell:
After some weekends spent profiling `mogwai` and `mogwai-dom` I have an initial offering :) 

This continued @flosse's work from #1157. 

It's possibly controversial that the application itself is [hosted on crates.io](https://crates.io/crates/mogwai-js-framework-benchmark) and used for both the keyed and non-keyed versions, distinction is done with cargo features.

I'm happy _enough_ with the performance, but there's still room for improvement and lots of room for improving the filesize of deliverables as well as the memory footprint. Those are all dragons for a later adventure, unfortunately.


krausest:
Thanks. Looks much better than the first PR. What did the trick?
Bildschirm­foto 2023-02-06 um 08 22 40

schell:
## tl;dr
I had been working for quite a while on a new release that really changes the API to make it ergonomic and fit a niche. I hadn't been paying attention to optimization and the lib really needed to be run on some benchmarks. Thanks for providing that!

* Removed a lot of unnecessary debugging traces 
*  Consolidated update streams (instead of spawning an async loop per update per element, they are now consolidated into one)
* The "create element" path is optimized

There was a bunch of other little things but I don't remember all of them.

Issue #1176, opened 2023-02-07
bobril element.click doesn't work


krausest:
@Bobris I noticed there's one strange issue with bobril: Calling click on a button doesn't trigger the click event.
To reproduce open http://localhost:8080/frameworks/keyed/bobril/index.html, open the browser's dev tools and select the Create 1,000 rows button. Verify that $0 points to the button and call $0.click(), which doesn't trigger the click event.

<img width="556" alt="Bildschirm­foto 2023-02-07 um 19 37 20" src="https://user-images.githubusercontent.com/3374055/217335538-5a64f60c-3db7-48b0-8841-b3143707a1d7.png">

Can you please take a look at that?

Bobris:
It worked after updating to latest version #1177. Huge jump in time from 8 to 20 major version.

PR #1177, opened 2023-02-07
Updated Bobril to latest version and fix for #1176

Bobris:
Updated to latest version of Bobril and Bobril-build. Hope it will work on your machine as bobril-build is actually bundler written in C# and is dynamically downloaded from github.


krausest:
Thanks, looks like swap rows took a bad hit.

Bobris:
Thanks, will try to profile it.

PR #1183, opened 2023-02-13
DooHTML-v0.90.6-beta (keyed)

hman61:


krausest:
Looks like create/replace rows is no longer keyed (I just confirmed that the test works correctly):
```
doohtml-keyed is non-keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
ERROR: Framework doohtml-keyed is not correctly categorized
rebuild-check-single failed for keyed/doohtml
```

hman61:
why is a clear a requirement for a keyed system?

If every node gets accurately updated with the correct new key and content
it is still keyed, I don't understand why a keyed implementation would have
to clear first in order to be considered keyed.

If I'm going from 10000 to 1000 it gets cleared or from append to 1000 it
also gets cleared.

Isn't the purpose of this benchmark to show the most effective way of
rendering?
as long as the endresulting nodes have a key

krausest:
Keyed implementations give a guarantee that a new row has a clean dom state (even if updated outside the framework). Just attach a background color from the dev tools to the first row and click on create 1,000 rows again. The background-color will stay there, which means the row has no clean dom state and thus isn't keyed.

The optimisation you suggest is valid, but only for non-keyed implementations.

hman61:
ok, I added back the clear()

hman61:
I still disagree with the clear() requirement.
The test requirement should be:
After the create data, *N* number of elements should exist and each element
should be in the *original state* of the row template. The row element
should include a key attribute OR a key property should exist on the
nodelist that produced the element.

The previous version I submitted did not keep the selected state
because I'm doing this.select(undefined)  prior to the rendering, which is
of course is a burden on the implementer to have to remember and if is not
being unset it will persist the selected state, like you mentioned.

Although, when I release the data store version DooHTML the selected state
will be 100% managed in the data set, and any update to an existing node
would then 100% produce a DOM in the correct state, because the template is
using a computed for the selected state, and the implementer would not have
to worry about a potential mutated state of any elements that are bound to
the store data.

I also think that you need to consider adding more columns to the table so
that the rendering time will be a little bit longer so that it the time
factor has more "girth". The computers have gotten so fast so the
rendering  time on such a narrow table set is being produced so quickly.
The length of the values should also be equal length within each columns in
order to measure identical rendering "volume", Maybe just produce random
numbers of a certain length in each column instead of the sentence that is
now being used.

PR #1189, opened 2023-02-19
Update wasm bindgen

aclueless:
Hi @krausest ,

I am trying to update wasm-bindgen, I am also trying to modify the build process so it can be executed simply with `trunk build --release` (without having to invoke via npm).

http://localhost:8080/frameworks/keyed/wasm-bindgen/ failed, but
http://localhost:8080/frameworks/keyed/wasm-bindgen/bundled-dist/ works

The same thing happened with `keyed/silkenweb` (works with `/bundled-dist`, but fails with just http://localhost:8080/frameworks/keyed/silkenweb/). But I have not touched silkenweb and it already works on your system, so I guess I have some problems in my system.

Does this PR work on your system?


krausest:
Looks like the benchmark ran fine. The version in package.json didn't change, such that the displayed version stayed the same.
bundled-dist is a very good place for rust wasm files, since the build directory is deleted (which makes sense for js frameworks that are built regularly.)
I hope I find some time the next few days to check the build process works on my linux VPS and I'll come back here.

PR #1190, opened 2023-02-22
Update to latest Bobril 20.4.1

Bobris:
Should fix performance regression in row swapping.


krausest:
Thanks. Nice update!

PR #1191, opened 2023-02-22
Update to latest Mimbl 0.10.4

mmichlin66:


krausest:
Thanks. I'd say performance is unchanged in comparison to 0.10.3.

PR #1193, opened 2023-02-26
Update to latest Leptos 0.2.1

gbj:


krausest:
Thanks. A little bit faster than 0.2.0, hard to say if that's more than just noise.

PR #1196, opened 2023-03-01
Doo html v0.91.1 beta keyed

hman61:


krausest:
Thanks. Maybe it's not just noise and really a bit faster than the old version:

krausest:
BTW that append to many rows looks interesting. What are you doing differently to all other implementations?

krausest:
Ok. I found it:
Bildschirm­foto 2023-03-04 um 23 14 26
The test driver only checks that a <tr> is found, it doesn't check the contents.
Can you fix that please?

hman61:
I will try to get that fixed ASAP, no wonder why I ended up #1 in that
category LOL

hman61:
I have a fix for the append issue in DooHTML - keyed

https://github.com/krausest/js-framework-benchmark/pull/1199

PR #1197, opened 2023-03-03
add missing end tag

0-v-0:


krausest:
Thanks. Looks like (as expected) no change in performance:

PR #1199, opened 2023-03-04
DooHTML-v091.2-beta-keyed

hman61:


krausest:
Append rows looks much better 😄 
Bildschirm­foto 2023-03-05 um 13 05 02

PR #1202, opened 2023-03-09
added anansi benchmark

saru-tora:
Hello, this is a benchmark for my full stack framework [Anansi](https://github.com/saru-tora/anansi). It seems to pass all the tests.


krausest:
Sorry, looks like there's a bug in the implementation.
Click "create 1,000 rows" and then click on "x" to remove the row. The first row has now id 2. Now click on "x" to remove the row with id 2. That row will not be deleted, but the row with id 3.
Can you please take a look at it?

PR #1204, opened 2023-03-10
Add `million@2.0.0` benchmark

aidenybai:
Removes `unkeyed` Million 1.x implementations. 

Adds Million 2.x `keyed` implementation with (#801)


frameworks/keyed/million/src/main.js
```diff
@@ -0,0 +1,336 @@
...
+    );
+  };
+  return (props, key) => {
+    return new Block(root, edits, props, key, shouldUpdate);
+  };
+})();
+
+(() => {
+  new Block(
+    stringToDOM(
```
aidenybai, line 216: Does this count as too much optimization here?

fabiospampinato: Is this code generated by a compiler? (Asking since it's under /src)

aidenybai: No, it's not. I can rewrite it such that it uses the canonical API instead, do you think that's more aligned with this bench?

fabiospampinato: Yeah that would probably be better, I can't imagine people in practice will write code like that.

aidenybai: Got it, thanks!

krausest:
No doubt, it's pretty fast, but looking at the code I'm sure we'll have lots of discussions what flags to add (I propose #772). 
Bildschirm­foto 2023-03-15 um 22 15 00

ryansolid:
I see.. hmm... @aidenybai you can correct me, but it looks like on select you look up the VDOM nodes and then patch them directly. I agree this verges on #772. Instead of writing code to describe the selection in the view you do an imperative lookup and mutation.

aidenybai:
Sounds good, I agree with the #772 flag. Is the best way to add a new flag to open up a PR?

PR #1210, opened 2023-03-12
Doo html dom v0.91.5 beta

hman61:
First version of DooHTML-dom not running in the shadow dom


krausest:
The isKeyed test fails due to the hidden first tr. Is there a way to do without that hidden row in the dom?

hman61:
I just committed on update that I think will fix it, Can you retry

krausest:
Thanks, that works much better! 

PR #1213, opened 2023-03-14
ivi v2.0.0

localvoid:
- Pure `f(state) => UI` implementation
- Complete rerender on each change (doesn't use any `shouldComponentUpdate()` optimizations)
- Native events (doesn't use any type of event delegation)


krausest:
Nice!
Bildschirm­foto 2023-03-15 um 22 15 00

localvoid:
Thanks.

Oops, sorry, it seems that I've forgot to remove `sCU`[1] before the latest force push when I've tested it locally. It will be slightly slower, but I don't care. Still naively hope for a future when "signal"-based implementations rewrite their implementations in a declarative way, so that they derive "selected" value, but they are probably going to be slower than full diffing in this test case :) Maybe someday framework authors are finally start to optimize their libraries, instead of creating benchmark specific workarounds, because there are good general purpose solutions to similar problems outside of a webdev bubble.

I'll remove sCU with the next update, don't want to waste a lot of your time with updates. 

1. https://github.com/krausest/js-framework-benchmark/blob/660c725cc6e47d558f8fc67d1da282982bc1a7cc/frameworks/keyed/ivi/src/main.ts#L84

PR #1216, opened 2023-03-16
fix: Million.js #772 issue

aidenybai:
Fixes issues laid out in #772, specifically with the manual patching in select and remove benchmarks.


krausest (commented):
Looks good to me as far as #772 is concerned. 
Not necessary this time, but it would be cool if main.js would be moved into a dist folder (and  removes from git).
I'm not sure whether I should approve or you want to approve the PR. Please remove the draft state if ready.

aidenybai:
Got it. Let me finish some things up for v2.1.1 so the release + pr is ready to go. Thanks for checking in

aidenybai:
Going to reopen a new PR with the fix

PR #1217, opened 2023-03-17
Fixed anansi remove test

saru-tora:
This should fix the problem of the wrong row being removed.


krausest:

PR #1221, opened 2023-03-26
`karyon` update

syduki:
An update to the latest version of `karyon` module, also a bit of refactoring of the benchmark implementation which may yield a few femtoseconds of performance gain, hopefully.


krausest:

PR #1223, opened 2023-03-31
Fix build for dojo

### LINKED ISSUE #1212: Update dojo to allow building with node 18.15.0
I can't build dojo with the latest LTS node. It fails with the following error:
```
> dojo-non-keyed@1.0.0 build-prod
> ./node_modules/.bin/dojo build

Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
Commands are not available:
 Failed to load module /home/stefan/js-framework-benchmark/frameworks/non-keyed/dojo/node_modules/@dojo/cli-build-app

Nested error:
 Package subpath './lib/input' is not defined by "exports" in /home/stefan/js-framework-benchmark/frameworks/non-keyed/dojo/node_modules/postcss/package.json
```
I moved both dojo implementations into the broken-frameworks folder.

@agubler Can you take a look at it please?

agubler:
This patches the dojo build to work for the latest node version. Please be aware if you are running on a mac you need to run install with the `--ignore-scripts` flag.

To test this it will need to be moved back to the "frameworks" folder from the "broken-frameworks" folder as that is the path it is expecting to be served from.

Resolves #1212

### DIFF
--- a/broken-frameworks/keyed/dojo/package-lock.json
+++ b/broken-frameworks/keyed/dojo/package-lock.json
@@ -7,12 +7,16 @@
     "": {
       "name": "dojo-keyed",
       "version": "1.0.0",
+      "hasInstallScript": true,
       "dependencies": {
         "@dojo/framework": "8.0.0"
       },
       "devDependencies": {
         "@dojo/cli": "8.0.0",
-        "@dojo/cli-build-app": "8.0.0",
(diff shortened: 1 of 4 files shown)


agubler:
Closing as can just bump postcss which provides the correct exports in package.json

Issue #1231, opened 2023-04-07
"ready memory" for blazor-wasm seems incorrect.


TrieBr:
[...]

PR #1232, opened 2023-04-07
Update Blazor to 7.0

TrieBr:
Updates Blazor tests to use 7.0. Helps resolve issue #1231


krausest:
Thanks. Looks much better (though there's still room for improvement):
Bildschirmfoto 2023-04-08 um 22 38 50

Memory consumption went down, but is still reported as much higher than for all other frameworks:
Bildschirmfoto 2023-04-08 um 22 48 52

PR #1234, opened 2023-04-08
ivi 3.0.0

localvoid:
- Updated to 3.0.0
- Removed `sCU` optimization (shouldn't make any difference, since previous submission had a bug and `sCU` didn't worked)

Haven't done any optimizations in 3.0.0, so there shouldn't be any improvements in results.


leeoniya:
@localvoid imo `@ivi/htm` might be a better choice here instead of `@ivi/tpl`.

krausest:
Thanks, needless to say it's pretty impressive:
Bildschirmfoto 2023-04-08 um 22 38 50
Select rows is one of the weaker benchmarks with high variance. Without it it's on par with vanillajs and doohtml:
 
Bildschirmfoto 2023-04-08 um 22 55 08

PR #1235, opened 2023-04-10
fix: Million.js #772

aidenybai:
Fixes issues laid out in https://github.com/krausest/js-framework-benchmark/issues/772, specifically with the manual patching in select and remove benchmarks.


krausest:
Thanks. I ran select rows twice, both times it reported about 17.5:
Bildschirmfoto 2023-04-15 um 22 24 19

aidenybai:
Thanks @krauset!

PR #1236, opened 2023-04-10
Update file path for Chrome on Windows

puckowski:
* Update file path for Chrome on Windows to use 64-bit version.

On my Windows 11 Version 22H2 machine I needed to update the paths for Chrome to use the 64-bit version, which I believe should be the new standard.


krausest:
Thanks!

PR #1238, opened 2023-04-12
Updated Voby to v0.48

fabiospampinato:
- I switched to a lazy reactivity system, so mandatory update for that.
- I also added an unkeyed implementation, which performs surprisingly poorly 🤔 I should look more into that.


krausest:
Thanks. As usually only keyed results are reported here 😄 
Bildschirmfoto 2023-04-15 um 22 24 19

PR #1239, opened 2023-04-14
Add kobold

Th3Whit3Wolf:


krausest:
Sorry, the `npm run isKeyed keyed/kobold` check reports some errors. Can you please take a look at them?
1. Please check that all aria attributes are present in the generated html: `Expected to find 'aria-hidden'=true on span in third td, but found  null`
2. The framework appears to recyle rows though the key changed and thus is non-keyed::
`Keyed test for swap failed. Expected at least 1 added and 1 removed TR, but there were 0 added and 0 removed
`
Please see https://www.stefankrause.net/wp/?p=342 for more information.

Th3Whit3Wolf:
Sorry, I switched it to non-keyed

maciejhirsz:
Hey @Th3Whit3Wolf, thanks for doing this!

Might be worth to bundle the js files since Kobold produces a lot of inlines. Trunk doesn't have an option for it, so I use `esbuild` (installed via npm) in the TodoMVC example: [see this shell script](https://github.com/maciejhirsz/kobold/blob/fe288137b0ba620ac99e64d3f24457633a9aa376/examples/todomvc/bundle.sh#L1-L4).

Once [#70](https://github.com/maciejhirsz/kobold/pull/70) is done we can do keyed version too.

Th3Whit3Wolf:
@maciejhirsz no problem, thank you for making a framework that was so much fun to play around with. It was incredibly to see kobold produce the smallest (or second smallest depending how you measure) wasm file.

I noticed your build script and I copied it [here](https://github.com/Th3Whit3Wolf/js-framework-benchmark/blob/b15f4e929dbbc29725eec04ddf66170c1774d45d/frameworks/non-keyed/kobold/package.json#L14).

frameworks/non-keyed/kobold/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+                            <Button {state} action={ButtonAction::Update} />
+                            <Button {state} action={ButtonAction::Clear} />
+                            <Button {state} action={ButtonAction::Swap} />
+                        </div>
+                    </div>
+                </div>
+            </div>
+            <table class="table table-hover table-striped test-data">
+                <tbody>
+                { for state.rows.iter().enumerate().map(|(i,l)| view! { <Row {state}  num={i} row={l.to_owned()} />}) }
```
maciejhirsz, line 212:
```suggestion
                { for state.rows.iter().enumerate().map(|(num,row)| view! { <Row {state}  {num} {row} /> }) }
```
No need to clone `RowData`

frameworks/non-keyed/kobold/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+        }
+    }
+    fn append_rows(&mut self, count: usize) {
+        self.rows.reserve_exact(count);
+        let d = self.last + 1;
+        for i in 0..count {
+            let adjective = ADJECTIVES[random(ADJECTIVES_LEN)];
+            let colour = COLOURS[random(COLOURS_LEN)];
+            let noun = NOUNS[random(NOUNS_LEN)];
+            let mut label = String::with_capacity(adjective.len() + colour.len() + noun.len() + 2);
```
maciejhirsz, line 77:
```suggestion
            let mut label = String::new();
            label.reserve_exact(adjective.len() + colour.len() + noun.len() + 2);
```

This may seem counter-intuitive, but it helps to ensure that pushing to the label reallocates...

frameworks/non-keyed/kobold/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+        self.clear();
+        self.append_rows(10_000)
+    }
+    fn add(&mut self) {
+        self.append_rows(1_000)
+    }
+    fn update(&mut self) {
+        self.rows.iter_mut().step_by(10).for_each(|row| {
+            row.label.reserve_exact(4);
+            row.label.push_str(" !!!")
```
maciejhirsz, line 109: ...here...

frameworks/non-keyed/kobold/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    bind! { state:
+        let remove = move |_| state.remove(num);
+        let select = move |_| state.select(row.id);
+    }
+
+    view! {
+        <tr.{is_in_danger}>
+            <td class="col-md-1">{row.id.to_string()}</td>
+            <td class="col-md-4"><a onclick={select}>{row.label}</a></td>
```
maciejhirsz, line 138:
```suggestion
            <td class="col-md-4"><a onclick={select}>{ref row.label}</a></td>
```

...which makes diffing by pointer here safe.

frameworks/non-keyed/kobold/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    view! {
+        <div class="col-sm-6 smallpad">
+        <button
+            id={id}
+            class="btn btn-primary btn-block"
+            type="button"
+            {onclick}
+        >
+            { text }
```
maciejhirsz, line 184:
```suggestion
            { static text }
```

This `text` never changes. Alternative would be to wrap the entire body of `Button` in `fence` on `ButtonAction`, though that probably would require a  `#[derive(Diff)]` to be user friendly (note to self).

frameworks/non-keyed/kobold/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+            ButtonAction::Update => state.update(),
+            ButtonAction::Clear => state.clear(),
+            ButtonAction::Swap => state.swap(),
+        };
+    }
+
+    view! {
+        <div class="col-sm-6 smallpad">
+        <button
+            id={id}
```
maciejhirsz, line 179:
```suggestion
            id={static id}
```

This `id` never changes. Alternative would be to wrap the entire body of `Button` in `fence` on `ButtonAction`, though that probably would require a  `#[derive(Diff)]` to be user friendly (note to self).

maciejhirsz (commented):
Some suggestions that should improve performance and reduce Wasm size further.

Haven't checked if any of this works (some of these suggestions depend on others), so it's a good test of how well I understand my own framework :sweat_smile:.

frameworks/non-keyed/kobold/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+#[component]
+fn Row(num: usize, row: RowData, state: &Hook<State>) -> impl View {
+    let is_in_danger = class!("danger" if state.selected == Some(row.id));
+
+    bind! { state:
+        let remove = move |_| state.remove(num);
+        let select = move |_| state.select(row.id);
+    }
```
Th3Whit3Wolf, line 133: I get an error

> borrowed data escapes outside of associated function
argument requires that `'a` must outlive `'static`

when making [these changes](https://github.com/krausest/js-framework-benchmark/pull/1239/files/163ee40ead8c03b3dc9799883944b1f982baf335#r1171233986)

frameworks/non-keyed/kobold/src/lib.rs
Th3Whit3Wolf: This compiles fine but rust-analyzer complains errors [no such field](https://rust-analyzer.github.io/manual.html#no-such-field). Do you have any idea why?

frameworks/non-keyed/kobold/src/lib.rs
Th3Whit3Wolf: This compiles fine but rust-analyzer complains errors [no such field](https://rust-analyzer.github.io/manual.html#no-such-field). Do you have any idea why?

frameworks/non-keyed/kobold/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+            self.rows.swap(1, 998);
+        }
+    }
+    fn select(&mut self, id: usize) {
+        self.selected = Some(id)
+    }
+}
+
+#[component]
+fn Row(num: usize, row: RowData, state: &Hook<State>) -> impl View {
```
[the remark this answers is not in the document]
Th3Whit3Wolf: I get an error from the `bind!` macro

> error[E0521]: borrowed data escapes outside of associated function
   --> src/lib.rs:132:5
    |
129 |   fn Row<'a>(num: usize, row: &'a RowData, state: &'a Hook<State>) -> impl View + 'a {
    |          -- lifetime `'a` defined here
...
132 | /     bind! { state:
133 | |         let remove = move |_| state.remove(num);
134 | |         let select = move |_| state.select(row.id);
135 | |     }
    | |_____^ argument requires that `'a` must outlive `'static`
    |
    = note: this error originates in the macro `bind` (in Nightly builds, run with -Z macro-backtrace for more info)

frameworks/non-keyed/kobold/src/lib.rs
maciejhirsz: Right, the closure is trying to capture the entire row, copy the id on stack first:

```suggestion
    let id = row.id;
    
    bind! { state:
        let remove = move |_| state.remove(num);
        let select = move |_| state.select(id);
    }
```

frameworks/non-keyed/kobold/src/lib.rs
Th3Whit3Wolf: Adding this ref makes the update button stop working correctly. Sometimes it increments the string sometimes it doesn't.

frameworks/non-keyed/kobold/src/lib.rs
Th3Whit3Wolf: [Committed](https://github.com/krausest/js-framework-benchmark/pull/1239/commits/7bc250c09eb3fe71ca83175c3749975ddabe606d)

frameworks/non-keyed/kobold/src/lib.rs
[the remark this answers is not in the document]
Th3Whit3Wolf: I fixed by adding `let id = row.id.to_owned();` and using that instead of `row.id`;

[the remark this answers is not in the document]
Th3Whit3Wolf: [Committed](https://github.com/krausest/js-framework-benchmark/pull/1239/commits/7bc250c09eb3fe71ca83175c3749975ddabe606d)

frameworks/non-keyed/kobold/src/lib.rs
Th3Whit3Wolf: I fixed by adding let id = row.id.to_owned(); and using that instead of row.id;

Th3Whit3Wolf:
@maciejhirsz kobold now produces the smallest wasm file.

frameworks/non-keyed/kobold/src/lib.rs
maciejhirsz: FYI: I haven't checked it, but I think that's because the allocator is sometimes able to grow the allocation in place so the pointer doesn't change.

Could force a fresh allocation with some such:

```rust
// This creates a new string so that we temporarily have two, ensuring
// the new string has a different pointer.
row.label = String::from_iter([&row.label, " !!!"]);
```

Though that's super clumsy and I wouldn't expect user code to be written like this. Let's keep things as they are for now.

krausest:
Is this PR ready to be merged?

Th3Whit3Wolf:
@krausest yes it is

maciejhirsz (approved):

PR #1241, opened 2023-04-16
Add glob support for rimraf for Windows

puckowski:
* Add glob support for rimraf for Windows hosts.

Tested on:
- Ubuntu 20.04.5 LTS
- Windows 11 Version 22H2

```rimraf``` version 4.2+ puts glob support for systems that don't natively support globs behind the ```--glob``` option.

With this change users on Windows can build the results table. 

Don't believe there are adverse impacts to non-Windows users, but I haven't tested on a macOS machine.


krausest:
Thanks - works on OSX too.

PR #1244, opened 2023-04-17
Master etch 0 14 1 fix

puckowski:
- Fix build issues.

Tested on:

```
npm 9.5.0
node 18.15.0
```

```
 npm run isKeyed keyed/etch

> webdriver-ts@1.0.0 isKeyed
> cross-env LANG="en_US.UTF-8" node dist/isKeyed.js keyed/etch

args.framework undefined true
Frameworks that will be checked etch-v0.14.1-keyed
*** headless undefined
etch-v0.14.1-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as keyed in the results
```


krausest:
Thanks. Results are updated. Here's a screenshot for keyed implementations:
Bildschirmfoto 2023-04-24 um 13 01 46

PR #1245, opened 2023-04-17
Master legend-state 0.14.0 fix

puckowski:
- Fix broken legend-state implementation.

Tested on:
```
node 18.15.0
npm 9.5.0
```

```
 npm run isKeyed non-keyed/legend-state

> webdriver-ts@1.0.0 isKeyed
> cross-env LANG="en_US.UTF-8" node dist/isKeyed.js non-keyed/legend-state

args.framework undefined true
Frameworks that will be checked legend-state-v17.0.1 + 0.14.0-non-keyed
*** headless undefined
legend-state-v17.0.1 + 0.14.0-non-keyed is non-keyed for 'run benchmark' and keyed for 'remove row benchmark' and non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
```


krausest:
Thanks. Results are updated.

PR #1247, opened 2023-04-21
feat(million): memo parity with blockdom

aidenybai:
This PR adds a block memo that caches existing vnodes and recycles them when possible.


krausest:
Thanks. I took a look at it and got the following results. On the left ist the new version, on the right the old version:
Bildschirmfoto 2023-04-24 um 21 08 46
The implementation managed to pass voby, siffr and fullweb-template in this incremental run, but the differences are too close to be significant.
Since the implementation adds some complexity that might be considered over-optimisation by others let me ask: Should I merge the PR or do you prefer to keep the clean original implementation?

krausest:
Yes, I'm pretty sure we'll add a flag for manual node caching if more implementations come up with such solutions.
Maybe "Manual caching of (v)dom nodes" would be a good fit?
@leeoniya @ryansolid What's your take on it?

PR #1250, opened 2023-04-24
Update Svelte to version 3.58.0

puckowski:
* Update Svelte keyed and non-keyed to version 3.58.0.

Tested on:
```
Node v18.15.0
npm 9.5.0
```

Keyed:
```
npm run isKeyed keyed/svelte

> webdriver-ts@1.0.0 isKeyed
> cross-env LANG="en_US.UTF-8" node dist/isKeyed.js keyed/svelte

args.framework undefined true
Frameworks that will be checked svelte-v3.58.0-keyed
*** headless undefined
svelte-v3.58.0-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as keyed in the results
```

Non-keyed:
```
npm run isKeyed non-keyed/svelte 

> webdriver-ts@1.0.0 isKeyed
> cross-env LANG="en_US.UTF-8" node dist/isKeyed.js non-keyed/svelte

args.framework undefined true
Frameworks that will be checked svelte-v3.58.0-non-keyed
*** headless undefined
svelte-v3.58.0-non-keyed is non-keyed for 'run benchmark' and non-keyed for 'remove row benchmark' and non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
```

### DIFF
--- a/frameworks/keyed/svelte/package-lock.json
+++ b/frameworks/keyed/svelte/package-lock.json
@@ -14,7 +14,7 @@
         "rollup": "2.45.2",
         "rollup-plugin-svelte": "7.1.0",
         "rollup-plugin-terser": "7.0.2",
-        "svelte": "3.50.1"
+        "svelte": "3.58.0"
       }
     },
     "node_modules/@babel/code-frame": {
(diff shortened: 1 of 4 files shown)


krausest:
Thanks. (And also here: Those minor version bumps mostly don't impact the results here:)

PR #1251, opened 2023-04-24
Update Sling.js to version 18.1.0

puckowski:
* Update Sling.js to version 18.1.0.

Tested on:
```
Node v18.15.0
npm 9.5.0
```

```
npm run isKeyed non-keyed/slingjs

> webdriver-ts@1.0.0 isKeyed
> cross-env LANG="en_US.UTF-8" node dist/isKeyed.js non-keyed/slingjs

args.framework undefined true
Frameworks that will be checked slingjs-v18.1.0-non-keyed
*** headless undefined
slingjs-v18.1.0-non-keyed is non-keyed for 'run benchmark' and non-keyed for 'remove row benchmark' and non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
```

### DIFF
--- a/frameworks/non-keyed/slingjs/index.html
+++ b/frameworks/non-keyed/slingjs/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>Sling.js v18.0.0</title>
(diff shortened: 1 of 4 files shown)


krausest:

PR #1252, opened 2023-04-25
Update Alpine.js to version 3.12.0

puckowski:
* Update Alpine.js to version 3.12.0.

Tested on:
```
npm 9.5.0
Node v18.15.0
```


krausest:
Thanks. Those minor version bumps mostly don't impact the results here:

PR #1253, opened 2023-04-26
dev: Adding arrowjs implementation

abishekatp:
ArrowJs is new library which tries to achieve whatever frameworks are doing without build setup. More details about this library is [here](https://www.arrow-js.com/docs/).


krausest:
Thanks. On the level of alpine (i.e there's some room for improvement 😄) 

PR #1258, opened 2023-05-03
plaited 3.0.0

EdwardIrby:
Adds plaited library to benchmarks

### DIFF
--- a/frameworks/keyed/plaited/index.html
+++ b/frameworks/keyed/plaited/index.html
@@ -1,0 +1,65 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="utf-8" />
+    <title>Plaited-"keyed"</title>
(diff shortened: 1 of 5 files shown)


krausest:
Thanks - not a bad first run!
I think we have to add note #800 (selected is a property of the row) and #772 (setting danger class directly). Do you agree?

Bildschirmfoto 2023-05-09 um 10 50 13 PM

EdwardIrby:
> Thanks - not a bad first run! I think we have to add note #800 (selected is a property of the row) and #772 (setting danger class directly). Do you agree?
> 
> Bildschirmfoto 2023-05-09 um 10 50 13 PM

Yep I agree this library takes an [State Unmanagement](https://medium.com/better-programming/state-unmanagement-a2b1e2442163) first approach to the DOM. While there's a away to do it with my useStore utility to reflect state that's not really pattern I'm recommending, to users unless there doing something like d3 graphs. 

Thanks for getting this in so quickly!!!

PR #1266, opened 2023-06-04
Rewrite `react-mobx` benchmark

klaseca:


krausest:
Thanks. That mobx implementation was indeed pretty old school, but it was a bit faster:
Screenshot 2023-06-04 at 9 50 20 PM

PR #1269, opened 2023-06-14
Update legend-state to v1.2.9 and add a keyed version

jmeistrich:
I fixed legend-state and updated it to the latest release. And I added a keyed version (the optimized non-keyed behavior is an optional prop) as we discussed in https://github.com/krausest/js-framework-benchmark/pull/1094.


krausest:
Thanks. Nice to see there's now a keyed version!
Screenshot 2023-06-27 at 9 05 27 PM

PR #1271, opened 2023-06-16
Bau

FredericHeem:
This PR adds [bau](https://github.com/grucloud/bau), a reactive library under 300 lines of code.
Being very lightweight, it performs very well on the startup metric, out performing vanilla.js,  and well on the memory side. On the other use cases, some are excellent or good. Only the partial update and replace all rows are average or below average.

Screenshot 2023-06-16 at 03 03 26

Screenshot 2023-06-16 at 03 03 54

Screenshot 2023-06-16 at 03 04 15

### DIFF
--- a/frameworks/non-keyed/bau/css/currentStyle.css
+++ b/frameworks/non-keyed/bau/css/currentStyle.css
@@ -1,0 +1,2 @@
+@import url("/node_modules/bootstrap/dist/css/bootstrap.min.css");
+@import url("/css/main.css");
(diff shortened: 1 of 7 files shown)


krausest:
Thanks. Usually I don't post screenshots of non-keyed implementations, but this time there are so many non-keyed that I made an exception: 
Screenshot 2023-06-27 at 9 04 42 PM

FredericHeem:
Thank you, your time is greatly appreciated.

PR #1275, opened 2023-06-25
Add cample

antonmak1:
Cample.js framework implementation.


krausest:
Thanks. Usually I don't post screenshots of non-keyed implementations, but this time there are so many non-keyed that I made an exception: 
Screenshot 2023-06-27 at 9 04 42 PM

antonmak1:
Thanks a lot for the screenshot! It can be seen that select row is not as fast yet (x3 more time 😅), but in the future I will do my best to fix it! I have a couple of ideas on how to resolve this.

PR #1277, opened 2023-06-28
add metron 0.0.2

robbiespeed:
I put this under the keyed category because it passes the keyed test and functionally does the same thing and only affects the corresponding dom nodes. Instead of keys it uses fine grained messages on data changes to update the dom rather than a key per each item. You can see the dom manipulation paths for each data change type here: https://github.com/robbiespeed/metron/blob/856e14307827b0dc6c6dcf3a7b63fe4ea7328169/packages/jsx/src/web-dom/render.ts#L424

### DIFF
--- a/frameworks/keyed/metron/index.html
+++ b/frameworks/keyed/metron/index.html
@@ -1,0 +1,13 @@
+</html>
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
+    <title>Metron</title>
+    <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+<div id='main'></div>
+<script src="dist/main.js"></script>
+</body>
+</html>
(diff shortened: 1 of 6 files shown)


krausest:
Thanks. I'd add it as a keyed implementation, but there's one catch: `npm` check reports and error for the html structure. It expects aria-hidden to be set to "true" for the span in the third td. Could you please is explicitly to true?
(Otherwise it might be a little advantage for metron and it might indeed be wrong, see https://stackoverflow.com/questions/55061083/does-aria-hidden-need-a-value#:~:text=undefined%20(default)%3A%20The%20element's,if%20it%20is%20not%20rendered )

robbiespeed:
updated to enforce "true" for aria-hidden. Didn't see any perf impact after the change. Was not aware aria-hidden cannot be a boolean attribute, html is weirdly inconsistent.

krausest:
Thanks. Nice results for the first run!
Screenshot 2023-07-13 at 10 20 05 PM

robbiespeed:
Thank you. I appreciate your efforts in maintaining the benchmark, it's been very helpful for narrowing down problem areas in rendering.

PR #1279, opened 2023-06-30
Upgrade ef.js to v0.17.3

ClassicOldSong:
New init API w/ performance boost


krausest:
Thanks. Here's a screenshot of the results:
Screenshot 2023-07-13 at 8 30 20 PM

ClassicOldSong:
That's weird, it shouldn't be slower than v0.16.2, at least tested on my machine. As the logic of partial update and swap rows and hasn't changed, the result being this much slower definitely has some other influence factors inside.

ClassicOldSong:
The much increased swinging in results tells the story. Will wait the new run on Chrome 115.

PR #1280, opened 2023-06-30
`karyon` update

syduki:
An update to the latest version of `karyon` module, no deviations expected from that, nevertheless this merge adds an optimization for the "update rows" test (which I blatantly overlooked 😅), potentially resulting in a slight performance gain.


krausest:
Thanks. Here's a screenshot of the results:
Screenshot 2023-07-13 at 8 30 20 PM

PR #1289, opened 2023-07-13
Update svelte deps

nakrovati:
- Remove @rollup/plugin-commonjs
- Update rollup and @rollup-* deps to latest versions

There is no need for @rollup/plugin-commonjs, as svelte 4 has only 13 dependencies vs ~50 for svelte 3, so the chance of any of them not having an ES version in 2023 is minimal.


krausest:
Thanks for the cleanup!
Screenshot 2023-07-13 at 10 20 05 PM

PR #1292, opened 2023-07-13
Replace webpack with esbuild in keyed/react

nakrovati:
What for?
- Reduced number of dependencies from 3600 to 500 -> node_modules size from 42mb to 15mb
- Build time decreased from 4s to 80ms.

P.S: watch mode added manually. You can enable it by writing --watch after node esbuild.config.js.


krausest:
Thanks. This one is interesting, select row got a lot faster. Is this some side effect of esbuild? If I find the time I'd like to spend some time finding the reason.

Here are the results (we both missed updating the version in the package.json, so it still says 0.4.3:

nakrovati:
I don't know yet, but it could either be test error, or Webpack and Babel are outdated and didn't build newer versions of React well, because Esbuild barely changed the results when building other frameworks.

PR #1295, opened 2023-07-14
Replace webpack with esbuild in keyed/doz & non-keyed/doz

nakrovati:
Just reduced the number of deps and speeded up the build


krausest:
Thanks.  Here are the results. In contrast to react it got slower!

PR #1296, opened 2023-07-14
Update Bdc

nakrovati:
- Update bdc to version 1.1.0
- Add dev mode
- Update rollup, @rollup/plugin-terser and @rollup/plugin-node-resolve
- Enable minification with terser for production mode only


krausest:
Thanks.  Here are the results:

PR #1297, opened 2023-07-14
Update Knockout

nakrovati:
- Update Knockout.js
- Update Rollup
- Convert functions to classes
- Fix build-dev script

Also reduced build time from 4s to 3.2s.


krausest:
Thanks.  Any idea what happened to swap rows?
Here are the results:

nakrovati:
I don't have any ideas.  Later I will see what happened to it, but if this is a framework error and given that it has not been updated for 4 years, then this cannot be fixed.

PR #1298, opened 2023-07-16
Update to Leptos 0.4.4

gbj:


krausest:
Here are the results (we both missed updating the version in the package.json, so it still says 0.4.3:

PR #1304, opened 2023-07-17
Update Silkenweb to 0.5.0

simon-bourne:
This updates the benchmark to the latest version of Silkenweb.


krausest:
Sorry, I failed to run the comparison and right now chrome was updated to version 115. So I'll include 0.5.0 in the official run within the next few days.

simon-bourne:
Great, thanks! I've tested it on my machine, so it should all run smoothly.

PR #1328, opened 2023-07-26
Update cample to v3.0.6

antonmak1:
Also, changed main.js for the new version


nakrovati:
The conflict seems to be because I version package-lock.json v2, and non-keyed is already using v3. Write `rm -rf node_modules && npm install --pacakge-lock-only`.

krausest:

PR #1332, opened 2023-07-28
refactor: configureStyles.js

nakrovati:
- Add default values for `bootstrap` and `minimal`.
- Replace `var` and `let` with `const`.
- Replace synchronous calls to the file system with asynchronous ones.
- Split the code.
- Remove unused imports.
- Remove the commented code because its functionality was duplicated from the `copy.js` files.


krausest:
Thanks for your PRs!

nakrovati:
(≧◡≦)

PR #1337, opened 2023-07-30
Update skruv

SahAssar:
Updates to skruv. Should be quite a bit quicker and smaller, and I chose to minify (with esbuild) it this time. Depending on how the feeling is about https://github.com/krausest/js-framework-benchmark/pull/1217#issuecomment-1656388282 regarding service-workers I think it might be the smallest.

In my testing it lands somewhere between svelte and preact in terms of performance, hopefully that isn't too far off.


krausest:
Thanks - nice improvement!
Screenshot 2023-08-06 at 5 46 39 PM

PR #1339, opened 2023-08-01
Bau update

FredericHeem:
* new api for array
* performance improvement for partial update


krausest:

PR #1342, opened 2023-08-09
Rebuild server on fastify

nakrovati:
Rebuilt the server using Fastify. 

What's the result?
- Speeding up roots (except `/frameworks/*`) by 40 - 60%. 
- Simplification in writing tests (they are built into fastify).
- Reduced RAM usage.
- Less pain in case of switching to typescript.

What else?

- Added prettier and eslint.
- Added JSDoc types.
- Changed CSP routes to be more in line with the REST API style.
- Replaced synchronous `node:fs` calls with asychronous ones.


krausest:
Took a short look at it: `/ls' fails with the following error:
```
file:///Users/stefan/Source/Javascript/js-framework-benchmark/server/src/frameworks/frameworksServices.js:35
        throw new Error(`Package.json not found for ${framework}.`);
              ^

Error: Package.json not found for .DS_Store.
    at PackageJSONProvider.getPackageJSON (file:///Users/stefan/Source/Javascript/js-framework-benchmark/server/src/frameworks/frameworksServices.js:35:15)
    at async loadFrameworkInfo (file:///Users/stefan/Source/Javascript/js-framework-benchmark/server/src/frameworks/frameworksServices.js:97:23)
```

nakrovati:
I'll fix it soon

krausest:
Just noticed there's a ./keyed/.DS_Store on my machine.

nakrovati:
Ah, now it’s clear.  I'll add it to the exceptions for now.  By the way, why do you even need .Ds_Store in the keyed/ directory?

krausest:
As I mac user I suppose you never want that .DS_Store, but OSX sprinkles that directory (I guess when you set some view options in the finder).
Of course I could delete it, but OSX for sure will generate some of these again :)

nakrovati:
Then I'll add it to the exclusion list.

nakrovati:
Fixed it.

server/src/frameworks/frameworksServices.js
```diff
@@ -0,0 +1,178 @@
...
+    }
+  }
+}
+
+/**
+ * @param {string} keyedDir
+ * @param {string} framework
+ * @returns {boolean}
+ */
+async function isFrameworkDir(keyedDir, framework) {
```
krausest, line 76: async causes the error with .DS_Store. It must be a sync function.

krausest:
Found it as the fix didn't help. isFrameworkDir must not be aysnc otherwise the check in line 169 is always true. Then we can remove the exclusion again.

krausest:
I'll make the changes, since it works now nicely on my machine.

nakrovati:
That's right, I forgot about the existence of isFrameworkDir.  Although the option with filteredFrameworks worked for me, since after filtering it excluded Ds_Store, which is why loadFrameworkInfo was not called in it.

PR #1345, opened 2023-08-12
feat: angular 16

HyperLife1119:
Upgrade to Angular 16 and switch to standalone component.


krausest:
Thanks - but we've got a pretty bad regression. No-zone is now slower than zone.js...
And v16 seems a little bit slower (or maybe standalone components) than v15.
Screenshot 2023-09-02 at 9 54 45 AM

eneajaho:
@HyperLife1119 @krausest 
Maybe sth changed with the new typescript compiler and the code it generates. 

I also created a PR to try to use the new esbuild builder, maybe that improves sth. 

https://github.com/krausest/js-framework-benchmark/pull/1357

PR #1353, opened 2023-08-25
Add benchmark for VanJS

Tao-VanJS:
This PR adds the benchmark test for VanJS.

VanJS (https://vanjs.org/) is by far the world smallest reactive UI framework. Here are its key features:
- **Ultra-lightweight:** 0.9kB in gzipped minified bundle. 50~100 times smaller than most popular UI frameworks.
- **Zero-dependency:** No dependencies on any other library, no JSX, no transpiling. A true grab-n-go library that enables people to build UI right away.
- **Unopinionated:** Supports multiple programming paradigms: You can use it in a way similar to React, or occasionally manipulate DOM tree directly for performance or for convenience.

I ran the benchmark many times locally:
image
image

The geometric mean consistently lies around 1.12 ~ 1.15 (close to vanillajs and much lower than react). Some adjustments on implementations (optimization attempts) don't seem to have a significant impact on the result. Thus I am just committing the benchmark for the officially released version.

### DIFF
--- a/frameworks/keyed/vanjs/bundle.sh
+++ b/frameworks/keyed/vanjs/bundle.sh
@@ -1,0 +1,1 @@
+npx esbuild src/Main.js --bundle --outfile=dist/bundle.js
--- a/frameworks/keyed/vanjs/index.html
+++ b/frameworks/keyed/vanjs/index.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
+    <title>VanJS-keyed</title>
+    <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+<div id='main'>
+    <div class="container">
+        <div class="jumbotron">
+            <div class="row">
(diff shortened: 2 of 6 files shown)


Tao-VanJS:
> Out of curiosity I had a couple of questions, hope that's OK:
> 
> 1. Why are you not rendering the header with the button controls in the framework?

I started the code from the `vanillajs` reference, where button controls are in the HTML file, thus it's just less work for me.

> 2. In swaprows is there a reason to manually change the dom? Wouldn't it be easier to make the rows a state and using [State-derived child nodes](https://vanjs.org/tutorial#state-derived-child)? I'm guessing then you wouldn't need to manually clear with `bodyDom.textContent = ""` in clear either.

The initial version was using state-derived child nodes for swapping rows. But later I found DOM manipulation offers a good opportunity of optimization. VanJS [encourages](https://github.com/vanjs-org/van/discussions/64#discussioncomment-6333993) occasional DOM manipulation for performance optimization.

krausest:
Nice start!
Screenshot 2023-09-02 at 9 54 45 AM

PR #1356, opened 2023-09-01
`karyon` update

syduki:
Updating to the latest version of `karyon`, also reverting an optimization which actually is not quite effective in that specific benchmark scenario.


krausest:

PR #1357, opened 2023-09-02
fix: simplify angular example, and use esbuild builder

eneajaho:


krausest:
Thanks. Better and no-zone is faster than the zone.js-version, but still a little bit slower than v15. Here's a screenshot (left is your version, middle #1345 and right v15):
Screenshot 2023-09-06 at 10 28 36 PM

PR #1360, opened 2023-09-08
feat: add alins

theajack:
Add [alins](https://github.com/alinsjs/alins) benchmark


krausest:
Looks like the keyed implementation isn't correct
```
npm run isKeyed -- keyed/alins non-keyed/alins
...

theajack:
@krausest Removed the keyed implementation of Alins

theajack:
@krausest Thanks for your review. Alins is a non-key type framework, and the implementation under the keyed directory has been removed.

krausest:

PR #1361, opened 2023-09-13
Update cample to v3.1.0. Add keyed cample

antonmak1:
The first keyed version is finally finished. The framework no longer supports the non-keyed version, so I just moved the files.


krausest:
Thanks. Great to see there's now a keyed version.

PR #1363, opened 2023-09-16
Add Strve

maomincoding:
Thanks! 😁

Strve is a JavaScript library that converts strings into views (user interfaces). Strve is not only easy to use, but also flexible in breaking down different chunks of code. Developing user interfaces using template strings mainly takes advantage of the power of JavaScript and focuses only on JavaScript files.


krausest:
Thanks. Welcome strve!

Issue #1364, opened 2023-09-17
replace static index.html


krausest:
Instead of the old statically created index.html it would be much better to create the index.html dynamically via the server.

krausest:
@nakrovati If you want I'd welcome a PR on top of #1342 very much...

nakrovati:
Do you mean to do a build by invocation via URL, like `/ls/create-index`? I'm not quite sure what "dynamically" means in the context of the server.

krausest:
Yes, I'd say "/" and "/index.html" would be great URIs (and would be backward compatible). The index route would reuse the /ls logic to find the frameworks and versions and create the index.html via the templates that currently reside in webdriver-ts.

nakrovati:
Got it, I'll do it soon.

nakrovati:
By the way, I fixed the fastify server, check out the fix.

krausest:
Thanks!

PR #1369, opened 2023-09-18
Add index html generation

nakrovati:
#1364 I hope I understood you correctly.
- Add index.html generation via `/` and `/index.html` routes.
- Remove `createIndex` in `webdriver-ts`.
- Remove Axios from `webdriver-ts`. Works on Node.js >= 18


krausest:
Yes, thanks - it was almost exactly what I hoped to get. I removed the index.html creation and added fastify/view to create the index.html on the fly. If you want feel free to add compression or minification. I also renamed helpers/generateIndexHtml.js to match better the new responsibility.

nakrovati:
Oh, I didn't know that fastify has a plugin for ejs. Your solution is much simpler

PR #1371, opened 2023-09-18
ci: remove .circleci dist

nakrovati:
It doesn't seem to be used anymore, so it's best to remove it.


krausest:
Right - we should get rid of it.

Issue #1370, opened 2023-09-18
Switching from Webpack to Vite


nakrovati:
@krausest 

### About the problem

There is a problem with switching to vite. The vite-plugin-singlefile plugin in this case collects `plotly` and `BoxPlotTable` into one 2mb file.
I manually split everything into chunks using vite and added via the `inlinePattern` property in vite-plugin-singlefile, but it seems this plugin doesn't work together with `manualChunks`.

The output is:
- Either leave webpack, but lose acceleration from 10s to 0.5s for `dev` and from 40s to 10s for `build`. And the possibility of simple integration with million.js. 
- Or generate 1 big HTML file for 2mb.
- Or we don't build everything in one file and output files as is, but without `dev` server it will be impossible to view `table.html` (because of cors).

### How can we fix it

You can change the logic of the site a bit:
- For `/[year]/*` give them as they are shared by vite (now it seems exactly the same, only built with webpack)
- For `/current.html`, build one big HTML file (which I will reduce later by replacing plotly with D3), or build by default vite in `current` directories, which can be accessed by `/current/` route.

### SSG site build

You could also do an SSG site. Store the results of new benchmarks in `.json` and use Astro.js in SSG mode to generate a page for each benchmark. For old benchmarks you can make a redirect, if there are no `.json` files with results, and if there are, do the same for them. Then you can build the whole site at once from Vite, without worrying about manual creation of chunks and directory structure on the server. You can also set up GitHub Actions to automatically deploy the site to GitHub Pages when you publish a new release.

P.S: I'm getting a new 16-inch laptop soon, so I'll be able to move the site to Astro (+ react or svelte).

leeoniya:
i would be willing to implement something atop the 50kb uPlot ;)

i have most of the code done already in a demo: https://leeoniya.github.io/uPlot/demos/box-whisker.html

nakrovati:
Wow, 50kb. I would rewrite everything to it, just along with the complete webdriver-ts-results rework.

leeoniya:
heh, not too much [1] :rofl:, but it's pretty straightforward.

[1] https://github.com/leeoniya/uPlot/tree/master/docs

nakrovati:
Really simple. I want to decide about migrating to vite first, and then replace plotly and wait for my laptop to arrive to work properly.

@leeoniya By the way, kudos to you for the fuzzy library, I met it at work.

PR #1374, opened 2023-09-21
feat(web-ts-results): replace webpack with vite

nakrovati:


krausest:
It's 12.5 secs for webpack to 5.6 secs for vite on my machine.
The new table.html is now 2MB in size (2,049,336 vs 747,975 for webpack).
Current I don't think it's worth switching yet, maybe vite-plugin-singlefile will catch up.

nakrovati:
Hmm, I'm down from 35 seconds to 8. I wrote about the build size in the issue of the same name. You don't have to build everything in one file (just run the `build` script), so the result will be similar to what webpack does. Webpack, by the way, simply does not import plotly.cartesian, so the bundle weighs so much.

You don't have to build everything in one file, since you said that now you use the server to view the results page. `vite build` now divides everything into chunks just like webpack, but it doesn't import the code (not plotly) into 1 html file, so it can't be viewed through a browser (because of cors).

nakrovati:
Wrote about this in #1370

krausest:
I'm not sure I understand all the implications yet...
If I run npm run build I get a few files which look fine. If the server from the root directory could serve those files I think I'd be fine (then it should work with github pages too I assume).
Currently the dist/index.html tries to use absolute URLs for /index.css /plotly.css and /index.css and thus fails. Wouldn't relative URLs work fine here?

krausest:
What I'm missing is a hash-value in the generated imports. I prefer to put them in one directory on github pages and they'd break otherwise. Is it possible to make vite add a hash to the names?

nakrovati:
Yes, for github pages you need to specify `base`. Look here https://vitejs.dev/guide/static-deploy.html#github-pages. In general, github pages can easily host files generated by vite (with `base`), but only during development you need to run a server (`npm run dev`), but dev servers are already standard (vite, nuxt (nitro), webpack, etc.).

nakrovati:
Didn't you ask me to remove the hashes? Is there any way to contact you (discord, for example), because it is inconvenient to write here?

nakrovati:
To remove the hashes, remove the part 
        ```javascript
entryFileNames: "[name].js",
        chunkFileNames: "[name].js",
        assetFileNames: "[name][extname]",` from vite.config.js
``` from vite.config.js

PR #1375, opened 2023-09-21
refactor(web-ts): minor code improvements

nakrovati:
- Improve code style according to typescript style guide.
- Move benchmark id's to `enum`.
- Replace nodejs import `fs` with `node:fs`.
- Replace `for` loop with `for..of` where possible.
- Use destructuring to get the number of constants used from the config.
- Reduce the number of `<any>` types.
- Reduce unnecessary `if..else`.
- Remove `@typescript-eslint/no-unused-vars` and `prefer-const`.
- Remove unused ipsorts.
- Remove `await` calls when calling functions that do not return `Promise`.


krausest:
I'm reluctant to merge this PR. I just tweaked the computation of the duration and I'm afraid this PR will reduce traceability in the history.

nakrovati:
Yes, I closed because of this.  It's mostly eslint fixes and nothing major, but I made a few commit mistakes so I want to redo it.

nakrovati:
I opened a new one with approximately the same changes, but the changes are easier to track by commits and I did not remove unnecessary `await`, change `var` and `let` to `const`, and also remove unused var`s (from eslint `no-unused-vars` rules).

New PR is much more secure. I checked each change through eslint and tsc --noEmit.

PR #1382, opened 2023-09-27
feat: add husky and lint-staged

nakrovati:
Add husky and lint-staged. Husky only runs `npx lint-staged`, while lint-staged checks all files (see `.lintstagedrc.json`) in the root directory and all subdirectories except `frameworks` and `broken-frameworks`.


leeoniya:
we've recently switched from Husky to https://www.npmjs.com/package/lefthook in Grafana, due to less intrusive/global changes it makes.

maybe worth considering.

PR #1384, opened 2023-09-28
feat: upgrade DLight.js to 0.10.x

IanDxSSXX:
Upgrading [DLight.js](https://github.com/dlight-js/dlight) to 0.10.x

### DIFF
--- a/frameworks/keyed/dlightjs/package-lock.json
+++ b/frameworks/keyed/dlightjs/package-lock.json
@@ -9,13 +9,15 @@
       "version": "1.0.0",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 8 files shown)


krausest:
Hi, there's an error for the non-keyed implementation. Remove row doesn't work correctly. Luckily the benchmark driver catches that case.
1. Create 1,000 rows
2. Remove row with id 4
3. Remove row with id 5 => nothing happens

Can you take a look at it please?

IanDxSSXX:
Thanks for pointing it out. Fixed in a newer version.

krausest:
Thanks, that worked:
Screenshot 2023-10-01 at 10 59 27 AM

PR #1392, opened 2023-10-02
Bump leptos to 0.5.0

birkskyum:
I noticed https://github.com/krausest/js-framework-benchmark/pull/1391 wasn't in sync with https://github.com/leptos-rs/leptos/blob/main/examples/js-framework-benchmark/src/lib.rs, so I made this pr.

It's not specific to this version, but I noticed `./bundled-dist/js-framework-benchmark-leptos.js` isn't minified when running `npm run build-prod-force` - @gbj , any idea why that is?


krausest:
That's something new: Two updates for the same framework 👍 
@flosse @birkskyum Could you both please agree on which PR should be merged?

flosse:
> That's something new: Two updates for the same framework 👍 @flosse @birkskyum 

:rofl: 

> Could you both please agree on which PR should be merged?

I wouldn't consider myself an expert of Leptos, which is why I give the lead to anyone who knows better :relaxed: 
Of course @gbj is the master here but he's probably also busy with more important topics :wink: 
So @birkskyum I think it's your turn now :smiley:

gbj:
@birkskyum To keep it simpler I made a PR to your fork, not sure if you saw it.

birkskyum:
Saw it only now

birkskyum:
@krausest , this one is ready.

krausest:
I'll post results here soon.

krausest:
So here they are:

PR #1393, opened 2023-10-02
Refactor webdriver ts

nakrovati:
- Improve the code according to the TypeScript style guide.
- Update the prettier configuration.
- Declare benchmark id's in `enum`.
- Declare frequently used LOG* properties used via `config.LOG*` as constants.
- Format the code with prettier.
- Replace `for` loop with `for..of` where possible.
- Replace `indexOf` with `includes` where possible.
- Reduce the number of `<any>` types.
- Remove unused imports.
- Remove .ejs and .dot files, as well as the index.html file generation script (it is now created on the server).

This PR is different. I didn't update `.eslintrc` here, and consequently didn't change `var` and `let` to `const`. I also didn't delete the extra `await` and `if..else`. So this PR is purely stylistic and is unlikely to break anything.


krausest:
Looks indeed easier to integrate than the old PR. But I think I have to fight for my own style 😄.
Most of the cleanup looks good to my, but I really don't like the trailing commas especially for method or function calls. 
To be honest I don't even understand why that should be an improvement.
The method array.sort takes zero or one parameter. So why should anyone prefer a comma there?
```
(results as CPUBenchmarkResult[]).sort(
      (a: CPUBenchmarkResult, b: CPUBenchmarkResult) => a.total - b.total,
    );
```

nakrovati:
Putting in commas is not so much my decision as it is Prettier's. In version 3 they decided to add them everywhere (https://prettier.io/docs/en/options.html#trailing-commas). It just seemed to me that since Prettier releases commas by default, and we have newer versions of TS and Node that support it, why not use it.

If you want to remove it, then remove it. It won't make any difference for Node and TS, just add `trailingComma: "es5"` to .prettierrc.

nakrovati:
@krausest please see the changes. I have some more improvements for webdriver, but they will conflict with this PR until its dead with master.

krausest:
Hi @nakrovati I'm a bit short of time and I'll give the frameworks PR a higher prio, so I'm afraid it'll take some time.
I've decided to review the changes thoroughly and I'll not adapt the prettier style (and I also decline a 80 char line limit).

nakrovati:
Good. Then you don't mind if I close this PR (it's too outdated) and open a new one with 1 commit only with Prettier formatting with `trailingComma:es5` and `printWidth:100 (or 120)`?

nakrovati:
I'll add the rest of the changes in other PRs so I don't mix it up with this PR.

PR #1394, opened 2023-10-04
Cleanup Svelte 4 implementation

trueadm:
This updates the Svelte 4 implementation so it's a bit clearer to follow and also uses `immutable` mode. There shouldn't be any real differences in performance – however this should be slightly easier to follow and read.


krausest:
Thanks - I added a tweak for remove row, since it didn't work. Performance is indeed about the same:

trueadm:
Thanks for fixing that!

PR #1395, opened 2023-10-07
Update Sycamore to 0.9.0-beta.2

lukechu10:
A lot of things have changed since 0.8 so some measurements might be faster and some others slower but I hope its overall faster.


krausest:
I'll post results here soon.

krausest:
All in all almost unchanged, though select got slower and create 10,000 rows faster:

PR #1404, opened 2023-10-10
adds OWL framework

woodholly:
OWL is a 25kb framework with class based components (not functions) with hooks, reactive state and concurrent mode. It was build as a part of Odoo, but can be used independently. 919 github stars.  https://github.com/odoo/owl
This test is almost complete copy of Vue test, but with Owl's syntax.


woodholly:
It's not my fault :)

PR #1412, opened 2023-10-13
Update cample to v3.2.0-beta.1

antonmak1:
I think this update should fix the select row to more adequate speed values, because 32.7 seconds when the maximum speed is somewhere around 10 seconds is...

### DIFF
--- a/frameworks/keyed/cample/package-lock.json
+++ b/frameworks/keyed/cample/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-cample",
-  "version": "3.2.0-beta.0",
+  "version": "3.2.0-beta.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "js-framework-benchmark-cample",
-      "version": "3.2.0-beta.0",
+      "version": "3.2.0-beta.1",
       "license": "MIT",
       "dependencies": {
-        "cample": "3.2.0-beta.0"
+        "cample": "3.2.0-beta.1"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


antonmak1:
I apologize for creating pull requests so often, but this is just nonsense. Since the summer, all values, more or less, except select row.

krausest:
Thanks. Cample apepars to be a bit faster than for beta.0 (even though it looks like the last chrome update made chrome slower in general).

antonmak1:
> Thanks. Cample apepars to be a bit faster than for beta.0 (even though it looks like the last chrome update made chrome slower in general).
> 
> Since the results for this run look somehow a bit odd: This run used chrome 118.0.5993.88 only for the changed implementations. It seems like most (if not all) frameworks got worse results than for the chrome 118 full run (vanillajs and wasm-bindgen we're unchanged and serve as a control group). Maybe some security fix causes the slow down. This is why the preliminary results carry the warning text :)

Heh, the result of select row hasn't changed at all :). The more interesting it is to make a framework! I think in the coming versions I will find something that slows down select row.

Regarding the speed change: When I test, my values fluctuate. It varies. Either the framework is faster on the localhost of some framework, then it is slower. Vanilla js as a control point, I think, has its place, but it seems to me that the ideal implementation (if we take speed) as an example will not be entirely accurate, because the numbers are very small :). But when the framework works a little slower, I noticed that the spread between the values may increase. The framework can have either 28 seconds or 32 seconds. It's just that the computer calculates differently.

antonmak1:
I looked like this, in the preliminary the Cample.js framework was somewhere in 64th place, and in the official result it was already at 48th, although the code did not change in the version.

PR #1415, opened 2023-10-13
Remove forgo

birkskyum:
I tried updating forgo from v2 to v3 / v4, and it doesn't seem worth spending time on. It's one of the absolut slowest in the bench, "newer" (8 months old) versions take forever to run, and where blazor wasm has an exciting future, this doesn't have any users.

Screenshot 2023-10-13 at 19 49 28

Screenshot 2023-10-13 at 19 53 06


krausest:
@spiffytech I see two options: Either updating or removing. If you prefer updating could you please submit a PR?

spiffytech:
Let's remove it.

PR #1416, opened 2023-10-13
Remove choo

birkskyum:
Choo is _the_ slowest keyed framework, and haven't had an update in 4 years.

I tried updating it to the latest (v7), but that didn't change anything. 

Screenshot 2023-10-13 at 20 28 34

Screenshot 2023-10-13 at 20 29 04


krausest:
@douglasduteil According to the [deprecation policy](https://github.com/krausest/js-framework-benchmark/blob/master/README.md#history) choo is going to be removed soon.

PR #1417, opened 2023-10-13
Remove bdc

birkskyum:
Amongst the slowest keyed frameworks, with little to no usage, or work being done.

Screenshot 2023-10-13 at 20 48 51

Screenshot 2023-10-13 at 20 48 35


leeoniya:
it would be useful to keep a reference of removed frameworks in some readme appendix, with these PRs listed so newcomers dont ask for them to be added back when they see them missing.

krausest:
> it would be useful to keep a reference of removed frameworks in some readme appendix, with these PRs listed so newcomers dont ask for them to be added back when they see them missing.

There's a history section on the bottom of the README. We could list them there and link to the last results that included them.

krausest:
@bwhmather Can we remove BDC?

bwhmather:
It does certainly seem to meet the current criterion for deletion.  I'm still using it, but I'm not aware of anyone else who's picked it up.

For my purposes it's finished and complete.  I mostly only use it where I don't have a build system so not much incentive to keep up with the ecosystem.  All the obvious bugs are fixed.  I realistically won't be doing that much more work on it.  

I do think it there's a case to be made for keeping it to answer the question of "why does react have a vdom?".  I think it was also winning on size for a while, although it seems to have dropped to third for Keyed implementations.

bwhmather:
(If you are happy to keep it then I'll fix whatever is causing that #1139 flag.  That really shouldn't be there)

krausest:
I found it very hard to look for eval statements. Thus I added an option to the server (http://localhost:8080/csp/enable or http://localhost:8080/csp/disable) such that a Content-Security-Policy is applied and 1139 is applied if the implementation fails with that CSP. 

BDC fails with
`
Refused to execute inline script because it violates the following Content Security Policy directive: "default-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-n0BFqUgm4Wa0a9k80nsAt5BQILfI3oLIYRYBlSkVuug='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'script-src' was not explicitly set, so 'default-src' is used as a fallback.
`

krausest:
Archived in 3d8a3d102fb651f06f6c6bc2b6dda44e9fb3a05e

PR #1419, opened 2023-10-13
Add react-loops

birkskyum:
Inspired by recent findings of how big a difference a faster for-loop makes in Angular 17
- https://github.com/krausest/js-framework-benchmark/pull/1403#issuecomment-1762062968 

I here present a version of `react` (class components) that replace:

```jsx
{data.map((item) => (
  <Row key={item.id} item={item} selected={selected === item.id} dispatch={this.dispatch} />
))}
```

 with the `<For>` component from [react-loops](https://github.com/leebyron/react-loops) which uses [iterall](https://github.com/leebyron/iterall) underneath (Iterators):

```jsx
<For of={data} as={item =>
  <Row
    key={item.id}
    item={item}
    selected={selected === item.id}
    dispatch={this.dispatch}
  />
} />
```

These results speak for themselves - it's with a significant margin the fastest implementation using React 18, and this For component can likely be used in most of the many existing React implementations - I just only did it to the fastest one which is the one with original Class components called just `react`.

Screenshot 2023-10-13 at 23 46 13


birkskyum:
This For component is in it's essence just this:

```ts
function For(props) {
  // Get the mapping callback
  var mapper = props.children || props.as;

  // Get the list
  var list = props.of;

  // Accept null / falsey as nothing to loop.
  if (!list) {
    return renderEmpty(props);
  }

  // Get the length of the list, rendering as empty if 0.
  var length = list.length;
  if (length === 0) {
    return renderEmpty(props);
  }

  // Map each list item into a React child.
  var mapped = [];
  for (var i = 0; i < length; i++) {
    mapped.push(mapIteration(mapper, list[i], i, length, i));
  }
  
  return mapped;
}

function mapIteration(mapper, item, index, length, key) {
  // Map this item into a React child node.
  // Produce additional info only if the mapper expects it.
  var result =
    mapper.length === 1
      ? mapper(item)
      : mapper(item, {
          index: index,
          length: length,
          isFirst: index === 0,
          isLast: index === length - 1,
        });

  // Clone the element with the provided key.
  return React.cloneElement(result, { key: String(key) });
}

function renderEmpty(props) {
  // If the collection was non-existent or empty, render the empty condition
  // if it exists, otherwise render null.
  if (!props.hasOwnProperty("ifEmpty")) {
    return null;
  }

  var ifEmpty = props.ifEmpty;
  return typeof ifEmpty === "function" ? ifEmpty() : ifEmpty;
}
```

atellmer:
Interesting. How does it solve the swap rows problem? I see significant improvements in your screenshot, but looking at the implementation of the react-loop, I can't understand the impact.

birkskyum:
Ah, dammit, i didn't catch that.
Makes sense

birkskyum:
You're right. It was too good to be true.

react-loops-v18.2.0 + 1.3.0-keyed is non-keyed for 'run benchmark' and non-keyed for 'remove row benchmark' and non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
ERROR: Framework react-loops-v18.2.0 + 1.3.0-keyed is not correctly categorized

birkskyum:
Right, adding the key like this makes it equal to `react`:

```ts
function mapIteration(mapper, item, index, length, key) {
  // Map this item into a React child node.
  // Produce additional info only if the mapper expects it.
  var result =
    mapper.length === 1
      ? mapper(item)
      : mapper(item, {
          index: index,
          length: length,
          isFirst: index === 0,
          isLast: index === length - 1,
        });

  // Clone the element with the provided key.
  return React.cloneElement(result, { key: item.id });
}
```

Well, now I know to check the isKeyed for future reference.

Issue #1421, opened 2023-10-15
Archive inactive frameworks 


krausest:
According to the [deprecation policy](https://github.com/krausest/js-framework-benchmark/blob/master/README.md#history) I'd like to archive some inactive frameworks. Archived frameworks will appear in the history section in the README with a link to the last result table they were included it. The cleanup allows adding new implementations while keeping the benchmark duration bearable.

I'm informing the contributors in this issue. Please let me know if there are good reasons to keep the implementation.

The following candidates appear to have litte activity in the last year. Is it OK to archive them?

- [ ] vuerx-jsx @ryansolid
- [ ] ko-jsx @ryansolid
- [ ] mobx-jsx @ryansolid 
- [x] uhydro @Krutsch: done
- [x] domdiff @WebReflection  (I know feels like groundhog day, but maybe it turns out differently a year later): done
- [x] neverland @WebReflection: done
- [x] lighterhtml @WebReflection: done
- [x] heresy @WebReflection: done
- [x] hullo @arturkulig done
- [x] resonatejs @CodeMonkeyAlpha: done 
- [x] etch (repo is readonly): done 
- ~~s2 @gr0uch~~
- ~~redom @pakastin~~
- [x] endorphin @sergeche: done  
- [x] sifrr @aadityataparia: done
- [x] 1more @Freak613: done   
- ~~doohtml @hman61~~
- [x] sledgehammer (repo is readonly): done  
- [x] domvm: done
- [x] fullweb-helpers (repo is readonly): done 
- [x] fullweb-template (repo is readonly): done 

(legend: checked = will be archived, strike through = will be kept)
Non-keyed will follow...

leeoniya:
you can archive domvm, both keyed and non-keyed. i dont actively maintain it and dont really plan to. ivi does what domvm set out to do, but does it better and much faster & smaller.

pakastin:
RE:DOM is actively used in all my projects and others as well, it's just so minimalistic there's not constant need to push updates. I'd still hope you would keep it, i'm going to update at least the packaging/building part soon.

CodeMonkeyAlpha:
Hi Stefan,

 

Hope you are well. Yes feel free to remove resonatejs if it is creating unnecessary work / noise. I have made a number of updates to it but haven’t published them, and frankly don’t have the time at the moment as it would require work on the documentation. It was designed specifically for one project / team. I wasn’t really under the illusion anyone but me and a few friends would be using it commercially. 

 

[...]

PR #1422, opened 2023-10-16
Update s2 to use published version from npm

gr0uch:
Since I got pinged [here](https://github.com/krausest/js-framework-benchmark/issues/1421), I might as well update it.

### DIFF
--- a/frameworks/keyed/s2/package-lock.json
+++ b/frameworks/keyed/s2/package-lock.json
@@ -7,7 +7,15 @@
     "": {
       "name": "js-framework-benchmark-keyed-s2",
       "version": "1.0.0",
-      "license": "Apache-2.0"
+      "license": "Apache-2.0",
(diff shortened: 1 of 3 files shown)


krausest:
Sorry, it's no longer keyed:
```
npm run isKeyed -- --headless true keyed/s2
s2-v1.0.17-keyed is non-keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
ERROR: Framework s2-v1.0.17-keyed is not correctly categorized
```
We can either move it to non-keyed or you'd have to fix the behavior.

gr0uch:
huh... I'll fix that

gr0uch:
the keyed test has been satisfied, i did have to add a line in the implementation to force the keyed behavior.

PR #1423, opened 2023-10-16
update art to v1.1.0

sullay:
1. Implement runtime and time slicing using the web-scheduler component library.
2. Use setImmediate instead of requestAnimationFrame. Performing extensive DOM operations in the requestAnimationFrame phase is not a good choice.


krausest:
Thanks. Seems like the new art version got slower for select row and clear row.

PR #1424, opened 2023-10-16
Update Silkenweb to latest version (0.6.0)

simon-bourne:
This updates Silkenweb to the latest version. Results should hopefully be the same as the last version (and were within tolerance on my machine).


krausest:
Thanks. The new Silkenweb also was 0.02 on average slower than in the offical results. I guess this is due to the updated chrome version.

PR #1425, opened 2023-10-16
Add Preact Signals and Hooks

### LINKED ISSUE #1418: Preact Signals and Hooks
Preact has 3 ways to handle state, of which only one is implemented

✅ [Classses](https://preactjs.com/guide/v10/components)
❌ [Hooks](https://preactjs.com/guide/v10/hooks)
❌ [Signals](https://preactjs.com/guide/v10/signals)

It would be interesting to have the two others present, since they are the current default.

birkskyum:
Closes #1418 

cc @marvinhagemeister  - Let me know if you can spot anything I do inefficient here, in case it's unexpected that class components come out as the fastest.

 
Screenshot 2023-10-16 at 18 19 00


marvinhagemeister:
@birkskyum This is the expected behaviour. The benchmark cases in this repository only measure how quickly an update is reflected in the DOM. They only measure raw rendering performance, they don't measure the 99% of other reasons why an app might be slow. Signals was written for the other 99% which are not reflected in this repository. The benchmarks here are fantastic for tuning the parts of a framework that exclusively deal with applying updates to the DOM, but not more than that. They don't reflect real world app performance.

Many real world apps have deep component trees and Signals allow you to bypass them to jump straight to the target component or even element in some cases. None of the benchmarks here in the repository look like that, they barely have any components in them at all. That's the main reason you're not seeing any improvements by adding signals here.

An area where you're likely loosing a lot of time compared to the class variant is that you put all the store update functions inside the render methods. Each benchmark run triggers a gazillion render calls and the overhead of creating closures inside the render function quickly adds up, especially in microbenchmarks like here. The class variant easily wins here because the class, including its methods, is only constructed once for each run.

Another thing to check is that are a few cases in this PR where signals are eagerly unwrapped inside the render function. Doing that subscribes the outer component to it, instead of only the underlying DOM node.

birkskyum:
Thanks for such detailed answer! Appreciate it

leeoniya:
> They only measure raw rendering performance, they don't measure the 99% of other reasons why an app might be slow. Signals was written for the other 99%

@marvinhagemeister this feels like too reductionist of an assessment of what's being tested here. this benchmark tests the cpu and memory efficiency of dealing with huge vdoms. your very next statement then describes that signals are optimizing exactly for this case by targeting parts of the vdom or dom directly. i've seen multiple people claim - without evidence - that this is just a microbenchmark that doesnt transfer to "real world" applications; not one person has ever shown that some theoretical 100-level component depth rather than the 10k-breadth + 8 depth i this benchmark is somehow the legitimate test, or is faster at all. are there any case studies (with code) you can point to that drive your point home?

in general ui frameworks are rarely the bottleneck except in cases of binding vdom re-rendering directly to rAF, resize, or mousemove. most of the time it's just poorly thought out algos, ui patterns, or waiting on an i/o fetch() or db call.

marvinhagemeister:
> in general, ui frameworks are rarely the bottleneck except in cases of binding vdom re-rendering directly to rAF, resize, or mousemove. most of the time it's just poorly thought out algos, ui patterns, or waiting on an i/o fetch() or db calls.

@leeoniya My words exactly. The benchmark here doesn't take poor UI patterns, over-engineered business logic or a mess of UI re-rendering loops into account, hence my comment. It's those things that signals help the most with and where the most performance wins can be gained. We cover this topic in our introduction blog post for signals.

leeoniya:
> It's those things that signals help the most with and where the most performance wins can be gained. We cover this topic in our introduction blog post for signals.

i don't understand your reasoning here.

the entire article [1] is literally about how vdom updates are expensive because in complex apps all state ends up getting lifted to root and every state update is a full vdom rerender. there is a profile showing the gains in vdom re-rendering. nothing there says anything about the _state itself_ being expensive to update...because it isn't. what's expensive is the vdom rerender and diff. which is exactly what this benchmark tests and why hand-waving this benchmark away is non-realistic is nonsense; signals optimize vdom rerender in UI frameworks. yes, they are a generic DAG that _can_ optimize things like large dependencies like a spreadsheet grid where the _state itself_ is expensive to update, but that's not what we're talking about here.

[1] https://preactjs.com/blog/introducing-signals/

rschristian (commented):
May I ask why there's the (rather large) difference between the Preact and React hooks implementations? I'd imagine they would match, unless there's no common "best" for them, requiring they differ to show off the framework itself.

birkskyum:
> May I ask why there's the (rather large) difference between the Preact and React hooks implementations? I'd imagine they would match, unless there's no common "best" for them, requiring they differ to show off the framework itself.

@rschristian yes you may. The differences typically falls into these categories:
- Where a framework has a function or concept another framework doesn't have.
- Where two different ways of going about something has a negligible/unmeasurable impact on the test score.

The differences arise when new frameworks are added to the benchmark, if they are based on another implementation (like you pointed out I used Solid here for the signals case, because I know it's been optimized a lot) or done more from scratch.

rschristian:
Hmm, I suppose I just find the implementation direction to be a bit odd given the framework(s) in question.

> Where a framework has a function or concept another framework doesn't have.

React and Preact (mostly) match APIs, so I'm not sure how much this plays here.

> Where two different ways of going about something has a negligible/unmeasurable impact on the test score.

I think this will always raise eyebrows, even if it is non-consequential. If there are two frameworks with matching APIs, but significant (benchmark) implementation differences between them, it'll perhaps raise some questions.

I fully appreciate that Preact & React are pretty unique here, as I don't think any other frameworks have the same sort of relationship, and I appreciate the limit of resources -- only so many hours in the day! But I feel like these two should probably stay mirrored unless there are significant perf deviations between them.

Personally, even as someone who works on Preact stuff, I'd be fine with see worse Preact numbers if was aligned with React's implementation -- that at least provides users with some data if they want to use this benchmark. As it stands, however, it can appear to be an apples-to-oranges comparison.

> Like you pointed out I used the solid implementation as a starting point for some of this, because I know it's been thoroughly optimized

Solid is a _very_ different framework, I'm not sure how many optimizations at play there are necessarily relevant to Preact, even with signals.

birkskyum:
@rschristian , I think you make raise a good point - I only paid attention to the performance when submitting these. If you find it would be an improvement to the Preact implementation to make it more similar to React, a PR will likely be the fastest way to get it sorted.

PR #1435, opened 2023-10-18
fix wasm-bindgen version

birkskyum:
the cargo.toml says 0.2.84, but the package.json only said 0.2.47


krausest:
Thanks. Despite it should be only an optical change I included it in the incremental run.

PR #1439, opened 2023-10-19
Format webdriver-ts dir

nakrovati:
Updating the prettier config (default values everywhere except `printWidth` and `trailingComma`) and formatting the webdriver-ts code.

### DIFF
--- a/webdriver-ts/src/isKeyed.ts
+++ b/webdriver-ts/src/isKeyed.ts
@@ -1,5 +1,10 @@
 import yargs from "yargs";
-import { checkElementContainsText, checkElementExists, clickElement, startBrowser } from "./playwrightAccess.js";
+import {
+  checkElementContainsText,
+  checkElementExists,
+  clickElement,
+  startBrowser,
+} from "./playwrightAccess.js";
 import { config, FrameworkData, initializeFrameworks, BenchmarkOptions } from "./common.js";
 
 import * as R from "ramda";
@@ -10,7 +15,8 @@ let args: any = yargs(process.argv)
     "$0 [--framework Framework1 Framework2 ...] [--benchmark Benchmark1 Benchmark2 ...] [--chromeBinary path] \n or: $0 [directory1] [directory2] .. [directory3]"
   )
   .help("help")
-  .boolean("headless").default("headless", false)
+  .boolean("headless")
+  .default("headless", false)
   .array("framework")
   .array("benchmark")
   .string("chromeBinary").argv;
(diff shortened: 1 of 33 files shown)


webdriver-ts/src/isKeyed.ts
```diff
@@ -196,35 +221,53 @@ async function assertClassesContained(elem: ElementHandle<HTMLElement>, expected
...
+      "tr"
+    ))
+  ) {
     return false;
   }
 
   // first td
   let td1 = await page.$("tbody>tr:nth-of-type(1000)>td:nth-of-type(1)");
-  if (!(await assertClassesContained(td1 as ElementHandle<HTMLElement>, ["col-md-1"], "first td"))) {
+  if (
```
krausest, line 236: That "improvement" really looks silly...

webdriver-ts/src/webdriverAccess.ts
```diff
@@ -193,17 +260,26 @@ export async function clickElementById(driver: WebDriver, id: string, isInButton
...
-export async function clickElementByXPath(driver: WebDriver, xpath: string, isInButtonArea: boolean) {
+export async function clickElementByXPath(
+  driver: WebDriver,
+  xpath: string,
+  isInButtonArea: boolean
+) {
   return await retry(5, driver, async function (driver, count) {
-    if (count > 1 && config.LOG_DETAILS) console.log("clickElementByXPath ", xpath, " attempt #", count);
+    if (count > 1 && config.LOG_DETAILS)
+      console.log("clickElementByXPath ", xpath, " attempt #", count);
```
krausest, line 270: I (hope I) wouldn't use multi line if blocks without braces.

webdriver-ts/src/timeline.ts
```diff
@@ -1,378 +1,523 @@
...
+          type: "click",
+          ts: +e.ts,
+          dur: +e.dur,
+          end: +e.ts + e.dur,
+          pid: e.pid,
+          evt: JSON.stringify(e),
+        });
       }
+    } else if (e.name === "Layout" && e.ph === "X") {
+      if (config.LOG_DETAILS) console.log("Layout", +e.ts, +e.ts + e.dur - click_start);
```
krausest, line 42: Readability was in my opinion MUCH better for the old version.

nakrovati: It's a matter of taste, but for example on my laptop 16 inches old code, although in some moments and looks okay, but it is more difficult to read from left to right lines of 120+ characters. Besides, most of this code can be quickly made to look nice by refactoring it a bit, because in this PR only formatting.

nakrovati: I just thought it would be better to add prettier everywhere and format the code first before refactoring it later

webdriver-ts/src/webdriverAccess.ts
nakrovati: It only happened because of the length of the string. If it is less than 100, the function after `if` is not moved to another line

webdriver-ts/src/isKeyed.ts
nakrovati: Yeah, it looks like shit.

krausest:
I'll merge most of the changes, but there are some that I don't want there (even if that means that automatic formatting won't work). It's essential that e.g. timeline.ts and the benchmark*.ts keep readability in a way that I can change them easily...

nakrovati:
Well, then put it in .ts. And benchmarks*.ts seems to look good after formatting. What's wrong with them?

PR #1441, opened 2023-10-19
Update to Marko 5

birkskyum:
I've managed to make a Marko 5 version using using Marko Run.

It appears to be noticeably slower than Marko 4.

@DylanPiercey - Is that an unexpected result? Can you see something I didn't do in a non-optimal way?

Screenshot 2023-10-19 at 23 42 20


DylanPiercey:
@birkskyum firstly thanks for doing this!

Just glancing at the code here, it appears that Marko is running in debug mode which would impact performance.
As it stands Marko is only bundled in production mode if you set `NODE_ENV=production` while bundling.

This is a bit strange and probably something we'll change once we come around to it.

birkskyum:
Can I set this in Vite config?

DylanPiercey:
@birkskyum I believe that should work. If not you could just update the `build-prod` command to prefix with `NODE_ENV=production` (which we probably need to do on our examples 😓 )

birkskyum:
That did seem to help a bit, but there is still an overhead.

Screenshot 2023-10-20 at 00 19 04

DylanPiercey:
👍 that's possibly accurate. Thanks for the ping!

PR #1450, opened 2023-10-21
Update README.md 

### LINKED ISSUE #1449: README file contains old links to the results.
_Open js-framework-benchmark/webdriver-ts-results/table.html in a browser and take a look at the results. You can open the result table with the link http://localhost:8080/webdriver-ts-results/table.html_

Here are the old links to the table. It’s just that when I started testing and checked the results, out of habit I went to the old path, but it no longer works, as I understand it. The new path is now like this - http://localhost:8080/webdriver-ts-results/dist/index.html

antonmak1:
Fix #1449

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ This will take some time (currently about 12 hours on my machine). Finally creat
 npm run results
 ```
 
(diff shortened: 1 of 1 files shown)


krausest:
Thanks - that's right.

PR #1453, opened 2023-10-21
Update cample to v3.2.0-beta.2

antonmak1:
I think this should definitely fix select row!

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ This will take some time (currently about 12 hours on my machine). Finally creat
 npm run results
 ```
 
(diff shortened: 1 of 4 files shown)


krausest:
Well, I'd say it almost does :)
Screenshot 2023-10-21 at 9 31 04 PM

antonmak1:
Well, the result has improved a little, but for me it’s still complete crap. I need to sit and think. It’s just that on my MacBook it shows that the speed seems to be normal, but it turns out it’s not. I simply wouldn't do pr so often if I wasn't sure the results would improve.

Okay, I’ll continue working on the framework then.

Thanks a lot!

PR #1455, opened 2023-10-21
refactor(webdriver): replace let and var with const

nakrovati:


krausest:
Sorry, I don't want that change. I prefer using const when I really have and constant value and not just if it's accidentally not changed afterwards. 
Dropping vars would be fine.

nakrovati:
Got it

PR #1459, opened 2023-10-22
Update Vue to 3.3.6

JereMiO:
- Update Vue to 3.3.6
- Update webpack deps

### DIFF
--- a/frameworks/keyed/vue/package-lock.json
+++ b/frameworks/keyed/vue/package-lock.json
@@ -9,19 +9,19 @@
       "version": "1.0.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "vue": "3.3.4"
+        "vue": "3.3.6"
       },
       "devDependencies": {
-        "@vue/compiler-sfc": "3.3.4",
+        "@vue/compiler-sfc": "3.3.6",
         "vue-loader": "17.2.2",
-        "webpack": "5.88.1",
+        "webpack": "5.89.0",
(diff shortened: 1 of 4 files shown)


krausest:
Thanks - vue 3.3.6 has some reports of being faster, but it doesn't show in this benchmark:
Screenshot 2023-10-25 at 9 54 31 PM

exodusanto:
Maybe the performance improvements are about new version 3.4.0-alpha.0, could be maybe insert in not keyed?

JereMiO:
@krausest 
I wonder if it is possible to have benchmarks for each version of VueJS to monitor the evolution of performance between versions (major, minor, patch) in a same environment. 

Is it possible to use this project for that?
Have other people already had this need? Do you have any advice?

krausest:
Yes, this would be possible, but I can't provide it. You can't and shouldn't compare historic runs, since the OS, chrome versions differ and the benchmark driver implementation might differ between those runs. The chrome version actually has a significant impact on the results. If we change some slow down factor for a benchmark or adjust the warmup runs the benchmark driver differences might render a comparison invalid.

But it could be solved by running different versions for a framework regularly.  You'd just need different directories like frameworks/keyed/vue-3.3.5 and frameworks/keyed/vue-3.3.6 and you could run the benchmarks and compare the difference between versions. But this would deserve a fork to track performance of one particular framework. I can't do this for all frameworks in here.

PR #1462, opened 2023-10-23
Refactor webdriver

nakrovati:
I added the ESlint plugin unicorn and refactored the code according to most of its rules, and I disabled potentially dangerous rules.


krausest:
I haven't looked at all changes yet, but I already have some points:

- I'm not going to change toFixed() to toFixed(0). Seems perfectly fine (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed?retiredLocale=de#syntax)
- I see no advantage replacing a.concat(b) with [...a, ...b]

nakrovati:
In the first case, to explicitly define the argument in toFixed so that it doesn't feel like it might have been forgotten.

In the second case just for code uniformity. For example, there was no `[...arr1,...arr2]`, `[...arr1, ...arr2, ...arr3 (etc.)]` and `arr1.concat(arr2)` in one code. In addition, `concat`, `toSpliced`, `slice`, `slice(0)`, and `split('')` can be replaced with it. 

P.S: if anything, the `many code improvements` commit contains all the improvements from eslint-plugin-unicorn.

nakrovati:
Even easier, this is a fix from the eslint-plugin-unicorn unicorn/prefer-spread rule

nakrovati:
Hi. Sorry I haven't answered, I've been on vacation in the mountains. I did a --force push and removed the `concat` replacement for `spread`, but left `toFixed(0)` with the unicorn rule not disabled so that it would show an error if the coefficient is forgotten to be added.

krausest:
@nakrovati I'm not going to merge this PR completely.  I've taken some time to think about it, but I came to that conclusion (and it was actually a difficult decision). Improvements are very subjective I guess, but most of the changes do not help me making this benchmark more correct or help me maintaining the code more easily
- some actutally alienate my own code. 
I can live with that for the webdriver-ts-results app, but the benchmark driver itself is a different story. 

Still I want to apply some of those changes (but maybe step by step).

Currently I have two main issues:
`npm lint` in webdriver-ts results in the following error:
```
webdriver-ts % npm run lint

> webdriver-ts@1.0.0 lint
> eslint src/ --ext .ts

Invalid option '--ext' - perhaps you meant '-c'?
You're using eslint.config.js, some command line flags are no longer available. Please see https://eslint.org/docs/latest/use/command-line-interface for details.
```

VSCode doesn't pick up the eslint config correctly. The debug output says:
```
[Info  - 4:14:46 PM] ESLint library loaded from: /Users/stefan/Source/Javascript/js-framework-benchmark/node_modules/eslint/lib/api.js
...
2023-11-01T15:14:46.659Z eslintrc:config-array-factory Error reading package.json file: /Users/stefan/Source/Javascript/js-framework-benchmark/package.json
...
No ESLint configuration (e.g .eslintrc) found for file: /Users/stefan/Source/Javascript/js-framework-benchmark/eslint.config.js
File will not be validated. Consider running 'eslint --init' in the workspace folder js-framework-benchmark
Alternatively you can disable ESLint by executing the 'Disable ESLint' command.
```
Do you know how to fix these two issues?

nakrovati:
It's a shame, but I think the improvements from eslint-plugin-unicorn don't change it much, just make it uniform and prevent the use of old practices, like replacing `f ? f : ""` with `f || ""` and `forEach` with `for...of`. You could say this is just an update to the EcmaScript version without causing any damage to the code.

And about Eslint, linting should be run from root. It should basically lint all code everywhere at once, but you can run it as `npx eslint webdriver-ts/` or change the script to `"lint": "eslint webdriver-ts/"`.

nakrovati:
As for the second one, you need to enable `"eslint.experimental.useFlatConfig": true` in the VSCode configuration or in `.vscode/settings.json`.

nakrovati:
With Eslint 9, the need to enable flat config support in VSCode will disappear, as well as support for old configurations, but in the meantime you have to enable it manually

PR #1464, opened 2023-10-24
DooHTML-DOM-v0.92.1-beta

hman61:
Fixed minor issue in attribute parser


krausest:
Hmmm - got a funny error:
```
doohtml-non-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as keyed in the results
ERROR: Framework doohtml-non-keyed is not correctly categorized
```
It's no shame having no non-keyed implementation (vs. vice versa). Can we just merge the keyed implementations and remove the non-keyed?

hman61:
Yes please. you can skip the non-keyed

hman61:
I will revert the non-keyed shortly

krausest:
Left is the new version, on the right the old:
Screenshot 2023-10-25 at 10 38 33 PM

Issue #1469, opened 2023-10-26
React implementation bug for swap rows


twharmon:
STR:

- create 1k rows
- delete one row only
- swap rows

The screen will go blank and/or you will see an error message in the console:
> TypeError: Cannot read properties of undefined (reading 'id')

This is probably caused in these lines:
```javascript
if (data.length > 998) {
  return this.setState({ data: [data[0], data[998], ...data.slice(2, 998), data[1], data[999]] });
}
```
It is trying to index on `999` when there may only be 999 items in the array.

This error is not encountered when the benchmark is ran.

twharmon:
From what I can tell, this would be resolved with

```javascript
return this.setState({ data: [data[0], data[data.length - 2], ...data.slice(2, -2), data[1], data[data.length - 1]] });
```

krausest:
Yeah, but the spread operator is way cooler 😄 
I fixed all react implementations. Interesting to see how often the faulty code was copied.

krausest:
![Screenshot 2023-10-31 at 7 46 35 PM](https://github.com/krausest/js-framework-benchmark/assets/3374055/3a54c8db-1f47-494f-a290-3ffd790e5f2e)
As expected results did not change for the implementations for most react implementations.
I don't understand the reason why, but react-redux and react-redux-rematch got slower in this run, for example for create rows (so it hardly can be due to this change). I tried multiple runs, but consistently got around 60 msecs (i.e. much higher than the 55 msecs before). 
I'm not sure if there was a minor chrome update in between and whether the OS Update to 14.1 happened in between.

PR #1470, opened 2023-10-26
Rendrjs: Fix swap bug fix

twharmon:
Swap leads to a runtime error. This version was derived from a react implementation, which has the same bug (#1469).


krausest:
Results are updated.
Screenshot 2023-10-31 at 7 46 35 PM
No change for renderjs (as expected).

PR #1471, opened 2023-10-27
Update cample to v3.2.0-beta.4

antonmak1:
Now, I hope, select row should become normal)

### DIFF
--- a/frameworks/keyed/cample/package-lock.json
+++ b/frameworks/keyed/cample/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-cample",
-  "version": "3.2.0-beta.2",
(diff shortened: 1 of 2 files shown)


krausest:
At least a bit better...
Screenshot 2023-10-31 at 8 52 33 PM

antonmak1:
@krausest Thanks a lot! I think now I understand what the problem is with the script in the framework.

antonmak1:
Previously, the results were simply inconsistent. The numbers were near the minimum, but one or two runs were constantly given x2-3 of the current one. As a result, the results on my computer showed a normal number, but in fact, due to instability in the framework, it was fairly higher. Now the numbers are even, but large. And this is exactly what is needed in terms of proper operation.

PR #1474, opened 2023-10-30
fix(web): disappearing border on the left side of tables when scrolling

nakrovati:
Fixed disappearing borders on the left and right of the first columns of tables when scrolling.

### DIFF
--- a/webdriver-ts-results/src/App.css
+++ b/webdriver-ts-results/src/App.css
@@ -75,16 +75,16 @@ thead.dummy th {
   text-align: left;
   overflow: visible;
   height: 42px;
-  padding:0 !important;
-  border:0;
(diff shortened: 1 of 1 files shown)


krausest:
Thanks - looks much better, but not yet absolutely perfect 😄 
The line doesn't disappear anymore, but it gets thinner when scrolling.
Screenshot 2023-10-31 at 8 00 31 PM
Screenshot 2023-10-31 at 8 00 37 PM

nakrovati:
Yes, I noticed, and I don't know why they are not `border` and `:first-element::before` are not equal. Will now create a new PR with a rather silly fix for this.

nakrovati:
Guess I won't. The fix works in safari but doesn't work in chrome.

PR #1476, opened 2023-11-01
chore: replace script `build-dev` with `dev`

nakrovati:
Just renamed the `build-dev` scripts to just `dev` to avoid writing the extra `build-` part, since no file has `build-dev` and `dev` scripts at once.


krausest:
One of the PRs one shouldn't let wait 😄

PR #1478, opened 2023-11-01
Improve cli

nakrovati:
Now you can run both files directly and import functions from them into other files, so that you don't have to create unnecessary child processes.


rebuild-build-single.js
```diff
@@ -82,43 +48,69 @@ function deleteFrameworkFiles(frameworkPath, filesToDelete) {
...
+  const filesToDelete = [
+    "yarn-lock",
+    "dist",
+    "elm-stuff",
+    "bower_components",
+    "node_modules",
+    "output",
+    !useCi && "package-lock.json",
+  ].filter(Boolean);
+
```
krausest, line 73: I'll have to do a manual merge. I didn't like that before and I still don't want this style in here.

PR #1479, opened 2023-11-02
feat: new interface and tiny perf improvements for plaited 5.0.1

EdwardIrby:
Plaited interface has changed significantly since 3.0 so here's a new bench for the new interface.


krausest:
Thanks - maybe a little bit faster:
Screenshot 2023-11-07 at 9 57 35 PM

PR #1481, opened 2023-11-02
Update cample to v3.2.0-beta.5

antonmak1:
I'm apologize for the inconvenience to do pull request again so early. I hope the framework will have time to get into release 119. The next versions will no longer go like this every week, so as not to distract you every time. It's just, well, it really turned out cool according to the code :)

### DIFF
--- a/frameworks/keyed/cample/package-lock.json
+++ b/frameworks/keyed/cample/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-cample",
-  "version": "3.2.0-beta.4",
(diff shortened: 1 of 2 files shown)


krausest:
I'm planning to run chrome 119 this weekend and thus it'll be included. (As the chrome update already happened I can't provide a comparison with chrome 118 anymore)

antonmak1:
@krausest Thanks a lot! Well, yes, it will be a little incorrect to compare 119 with 118. I know for sure that the results of the framework have become better, so it doesn’t matter.

PR #1485, opened 2023-11-06
Doo html dom v0.92.2 beta

hman61:


krausest:
Thanks - there's not much room for improvement (left is the new version):
Screenshot 2023-11-07 at 9 57 35 PM

PR #1467, opened 2023-10-25
Add OpenUI5

albertlast:
I'm not sure if it a keyed or non-keyed framework,
also it generate some error while running but the test are succeding --> so it's fine?

PR #1486, opened 2023-11-06
Improve openui5 benchmark

pskelin:
Hi, openui5 maintainer here.

I am submitting an improved implementation of #1467 with the following changes:
- switch to a direct rendering approach that is comparable to all other framework samples
- the select scenario is handled correctly by storing an index and setting the class via rendering, not via direct DOM manipulation
- use of self-contained bundle like recommended for production builds

This is a much more realistic comparison of rendering performance of the framework, as the built-in controls are optimised for scenarios where less controls are rendered with virtualized scrolling, but offer much more HTML markup and features and should not be compared to plain markup rendering.


krausest:
Sometimes it's hard to say whether results actually improved. Not in in this case 😄

PR #1488, opened 2023-11-08
Update ControlRenderer.js

aborjinik:
- Use the newer apiVersion flag of the Renderer
- ID attribute for the table rows added

### DIFF
--- a/frameworks/non-keyed/openui5/webapp/ControlRenderer.js
+++ b/frameworks/non-keyed/openui5/webapp/ControlRenderer.js
@@ -5,7 +5,7 @@
 sap.ui.define([], function () {
  
     var ControlRenderer = {
-      apiVersion: 2
+      apiVersion: 4
     };
 
     ControlRenderer.render = function (oRM, oControl) {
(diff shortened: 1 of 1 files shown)


aborjinik:
Hi, I am the co-owner of the openui5 rendering.

I am submitting an improved implementation of https://github.com/krausest/js-framework-benchmark/pull/1486 with the following changes:

The apiVersion=4 is the newest syntax for rendering openui5 controls and it should be set correctly.
Table rows should have an ID attribute like <tr id={{this.id}} class={{this._class}}> and this missing attribute is now added.

krausest:
I'm sorry, the implementation doesn't work on my machine:
Screenshot 2023-11-12 at 11 44 48 AM

aborjinik:
I am sorry I just missed out casting IDs to string. Could you please take a look again I just pushed the 2nd commit.

krausest:
That worked! And what worked even better is that the implementation is now keyed, which is a good think. Thus I moved the implementation to the keyed section (actually I messed it up and first pushed the delete, then the add, but at least it's now at the right place):
Here are the results: On the left are the updated keyed results on the right are the old results:

PR #1489, opened 2023-11-08
feat: add frei-hooks

AimWhy:


krausest:
Thanks - here's how it looks like:

PR #1490, opened 2023-11-08
Update Ember app to use template tag component format

IgnaceMaes:
Updates the Ember app to use the [new component authoring format](https://github.com/ember-template-imports/ember-template-imports).

Seems to have no significant impact on performance (before vs. after):

template-tag

Moved to plain functions as a helper, which seems to positively impact performance:

image


NullVoxPopuli (approved):

birkskyum:
Is there anything blocking use of v4 of the ember templates? https://github.com/ember-template-imports/ember-template-imports/releases/tag/v4.0.0

NullVoxPopuli:
> Is there anything blocking use of v4 of the ember templates? https://github.com/ember-template-imports/ember-template-imports/releases/tag/v4.0.0

for apps, no, but tooling packages that folks have in their editors still requires v3 (they declare a peer on v3), so effectively no one can use v4 except in specific cases
(gjs, no linting support, etc (not fun) -- so v3 is where we remain until the tools update)

krausest:
Thanks - here's how it looks on my machine. I see an improvement albeit a small one:

PR #1491, opened 2023-11-09
Bump Angular to stable 17

birkskyum:
Should be a trivial bump, just to get on stable.


pkozlowski-opensource:
@birkskyum thnx for your continuous work on keeping Angular packages up to date! 

Just FYI, we've released [17.0.2](https://github.com/angular/angular/blob/main/CHANGELOG.md#1702-2023-11-09) with a couple of important fixes, one of them impacting performance. In this sense 17.0.1 is "buggy" and we should go directly with 17.0.2

krausest:
Thanks - as expected no big change:

PR #1494, opened 2023-11-11
Upgrade Svelte to v5 alpha

trueadm:
This PR updates Svelte to the recent alpha of Svelte 5. It also updates the implementation to make use of the new features available in Svelte 5.

PR #1495, opened 2023-11-11
Preview - Sveltekit/Svelte 5

birkskyum:
**SvelteKit /w Svelte 4** ("sveltekit-4")
I took the existing Svelte 4 , and ported it to SvelteKit SSG, still with Svelte 4 - because it's the recommended way to build with Svelte (and easier to upgrade/configure).

**SvelteKit /w Svelte 5 (5.0.0-next.1)** ("sveltekit-5")
I then upgraded to Svelte 5 (still without runes), and **partialUpdate** broke, so I had to change the following to again trigger change detection, and I believe that's why this performs worse in partialUpdate test. ( fyi @trueadm - don't know if this is expected ).

```ts
clone[i].label += " !!!";
```

changed to

```ts
clone[i] = { ...clone[i], label: clone[i].label + " !!!" };
````

**Runes** ("sveltekit-5-runes")
I then enabled runes mode, and updated to that syntax (and reverted the prev. change to partialUpdates):
```ts
  let rowId = $state(1);
  let data = $state([]);
  let selected = $state(undefined);
```

**Runes Fine Gained** ("sveltekit-5-runes-fine-grained")
Finally, I also made each label a rune:

```ts
let label = $state(`${adjectives[_random(adjectives.length)]} ${
      colours[_random(colours.length)]
    } ${nouns[_random(nouns.length)]}`);
```

Screenshot 2023-11-11 at 23 18 34


trueadm:
I've already put up a Svelte 5 upgrade PR :) https://github.com/krausest/js-framework-benchmark/pull/1494

frameworks/keyed/sveltekit-5-runes-fine-grained/src/routes/+page.svelte
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  function createItem() {
+    let label = $state(`${adjectives[_random(adjectives.length)]} ${
+      colours[_random(colours.length)]
+    } ${nouns[_random(nouns.length)]}`);
+
+    return {
+      id: rowId++,
+      label,
```
trueadm, line 109: This isn't fine-grain. It needs to be inside a getter/setter. See https://github.com/krausest/js-framework-benchmark/pull/1494 for how you can use `$state` on classes instead.

trueadm:
I'm also not sure we need to bring along SvelteKit here too.

birkskyum:
Excellent, thanks! :)

PR #1497, opened 2023-11-12
Update uhtml to v4

WebReflection:
[uhtml v4 is out](https://github.com/WebReflection/uhtml/pull/86) so I'd like to have it tested and validated with all its improvements in here too, thank you!

### Changes

  * update *uhtml* to its latest
  * update all dependencies to avoid deprecation warnings while bundling
  * changed minimal amount of code to validate results against previous version, without changing any other logic behind the test

### DIFF
--- a/frameworks/keyed/uhtml/index.html
+++ b/frameworks/keyed/uhtml/index.html
@@ -3,4 +3,4 @@
 <title>µhtml keyed</title>
 <link href="/css/currentStyle.css" rel="stylesheet" />
 <div id="container"></div>
-<script src='dist/index.js'></script>
+<script type="module" src="dist/index.js"></script>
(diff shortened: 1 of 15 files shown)


krausest:
Thanks - here's how it looks like:

WebReflection:
It looks like either keyed or non keyed still uses v3 🤦🤦🤦😭

I’ll double check and follow up eventually, thank you

WebReflection:
OK, v4 is present in both keyed and non keyed … was build-prod missing or there’s something else I should look at? One screenshot refers to v3 and now I’m thinking that maybe that was the previous reference instead? Thanks for hints or clarification!

krausest:
Sorry, that was me. I didn't publish the result table until a few moments ago!

krausest:
Ok, maybe I need more explanations 😄. For keyed implementations I use to publish a screenshot of the results for the PR (left side) and the last results for the same frameworks (right side) such that one can compare the new and the old version. So the screenshot above is intended - and I hope it makes sense now.

To express my disgrace with non-keyed frameworks I only update the results table for non-keyed implementations and don't post a comparison. This time some interruption caused me to forget updating the results table.

PR #1498, opened 2023-11-13
Update Svelte 5 alpha

trueadm:
This bumps the Svelte implementation to a newer alpha that removes the `selector` API in favour of idiomatic selection using the `$state` rune. It might come a slight cost to this benchmark, but we feel like having an API that only handles singular selections for the purpose of having improved selection updates wasn't worth the cost in developer ergonomics.


krausest:
Thanks. I'd say no change (on the left are the updated keyed results on the right are the old results):

PR #1499, opened 2023-11-15
Add additional CPU benchmark for updates to large table

trueadm:
This PR adds an additional CPU benchmark – "partial update to large table". Essentially, it's the same benchmark as the partial update to 1k rows, except this is for 10k rows instead (with 2x slowdown instead of 4x slowdown).

This was an interesting one as inspecting and profiling text node updates locally, it occurred to me that the existing 1k row update benchmark wasn't really capturing the effectiveness that some frameworks utilise to ensure updates happen as quickly as possible – something which can be loosely emulated (and I mean loosely, as this benchmark isn't exactly representative) in real-world UIs that have frequently updating UI elements, such as trading applications or live database collection views.

I wasn't fully sure if I implemented this correctly and I had no idea how the weights were calculated, so I just used one of the other numbers in the array. Lastly, I wasn't sure how the warmup runs was calculated, and assumed it wasn't from config so I also updated the input entry for updating 1k rows to reflect that.


trueadm:
@syduki I'd argue that this actually does test the framework performance – just on a much larger subset of DOM nodes. If it didn't then the results wouldn't be so skewed, showing there's obvious gaps in how frameworks go about updating text nodes in the DOM.

It's actually not that uncommon to encounter over 10k DOM nodes with modern web apps today, just going to facebook.com and scrolling for a dozen or so seconds, and doing `document.querySelectorAll('*')` gives me 17k DOM elements. The same can be found on Github and going through a stream, or opening up the code editor, I get around 12k DOM elements viewing the source of this project with blame open. That's just the elements, there's probably another a few thousand text nodes in there and even more if they're using Shadow DOM.

localvoid:
> it occurred to me that the existing 1k row update benchmark wasn't really capturing the effectiveness that some frameworks utilise to ensure updates happen as quickly as possible

This benchmark is already biased towards direct-bindings and block-diffing libraries. Why you want to increase this bias? It would be much more interesting to look at the results on the different side of this spectrum, like decomposing DOM blocks into components with a lot of dynamic properties, scenarios like the [results table in this benchmark](https://krausest.github.io/js-framework-benchmark/current.html) with a lot of interdependencies when small change is going to trigger a lot of mutations, datatables with client-side sorting/filtering, client-side "GROUP BY" scenarios, etc.

leeoniya:
> it's actually not that uncommon to encounter over 10k DOM nodes with modern web apps today

just a friendly reminder that the 10k table is 80k nodes, not 10k ;)

https://github.com/krausest/js-framework-benchmark/blob/59fe5322460fb85e54834aba95450b9d956f77d6/frameworks/keyed/react-hooks/src/main.jsx#L74-L87

trueadm:
> > it occurred to me that the existing 1k row update benchmark wasn't really capturing the effectiveness that some frameworks utilise to ensure updates happen as quickly as possible
> 
> This benchmark is already biased towards direct-bindings and block-diffing libraries. Why you want to increase this bias? It would be much more interesting to look at the results on the different side of this spectrum, like decomposing DOM blocks into components with a lot of dynamic properties, scenarios like the [results table in this benchmark](https://krausest.github.io/js-framework-benchmark/current.html) with a lot of interdependencies when small change is going to trigger a lot of mutations, datatables with client-side sorting/filtering, client-side "GROUP BY" scenarios, etc.

I mean, yes that would be perfect. I just assumed that major changes like that to the benchmark would require updating every implementation – which likely isn't feasible.

trueadm:
> > it's actually not that uncommon to encounter over 10k DOM nodes with modern web apps today
> 
> just a friendly reminder that the 10k table is 80k nodes, not 10k ;)
> 
> https://github.com/krausest/js-framework-benchmark/blob/59fe5322460fb85e54834aba95450b9d956f77d6/frameworks/keyed/react-hooks/src/main.jsx#L74-L87

My point was that it wasn't only about the DOM element count. With Lexical, it wasn't uncommon for someone to paste in a large amount of content and create 100k+ DOM elements. Tables are infamously quite taxing for browser layout engines because of their layout/styling properties.

localvoid:
> I just assumed that major changes like that to the benchmark would require updating every implementation – which likely isn't feasible.

Agree. I just don't understand what is the point of the proposed changes, except for benchmarketing purposes to push some narratives that "X is faster than Y" by omitting a lot of nuances. It is obvious that efficiently implemented direct binding solution will be faster in such scenarios, so it won't provide any useful and actionable information for library authors. So the question is how much overhead should be added to showcase that it is faster in such scenarios, because we can always just add some artificial loops like how it was done in the Sierpinski Triangle benchmark that was used to push narrative that "time slicing improves performance".

trueadm:
@localvoid I actually wasn't using this benchmark to push a narrative. For me ivi was the quickest library for this particular new addition (much quicker than both vanilla implementations) even React performed as well as some other libraries too (such as Vue and Preact). If anything it highlighted areas where libraries might do simple things to improve performance (using `nodeValue` rather than `textContent` was quite significant to updating text nodes).

However, if the sentiment here is that I'm trying to showcase Svelte here or Inferno, which isn't true, then I'm happy to close the PR.

trueadm:
I'll close this for now, as I don't think there's any appetite for another benchmark.

localvoid:
> I actually wasn't using this benchmark to push a narrative.

I just misread your original comment for this pull request and thought that your original intent were to add more diffing overhead to emphasize update effectiveness for some types of libraries. But even if it wasn't intentional, this additional test will increase the gap between libraries that use block-diffing optimization and libraries that don't. In my opinion, block-rendering/diffing optimizations are already overemphasized in this benchmark.

> For me ivi was the quickest library for this particular new addition (much quicker than both vanilla implementations)

Probably some measurement errors. I've submitted ivi implementation without `sCU` optimization, so it performs a full diff in all test cases and it just can't be faster than vanilla.

localvoid:
> as I don't think there's any appetite for another benchmark.

It would be really interesting to see test cases with a lot of reusable components(with many dynamic properties) that receive mostly static values to showcase that with signal-based libraries it is possible to optimize reactive graphs at runtime and it will have way much less memory overhead than diffing solutions.

leeoniya:
we all remember https://github.com/wycats/dbmonster, right?

PR #1501, opened 2023-11-17
Update svelte runes / classic

birkskyum:
This keeps 2 implementations of Svelte (runes / classic), since svelte 5 introduces new Runes syntax, but at the same time stays backwards compatible with the Classic syntax which will stick around in the wild for quite a while. Many people might update existing projects to svelte 5, without an immediate full rewrite to runes, so it's interesting to see how both fare.

Also, selector api was removed from the runes API, which is reflected in this keyed runes implementation.

There is still not a non-keyed runes implementation - and I don't know if it's interesting since I generally pay more attention to the keyed ones.


krausest:
Thanks - @trueadm Can you please review whether the changes are fine for you?

trueadm:
I don't think we want to call the default Svelte 5 "runes". To us, the runes are the new default way of doing Svelte 5, so it should just be Svelte – as it is today. Having the Svelte 5 classic benchmark around is a good idea though. @birkskyum could you also update the npm version for both to the latest Svelte 5 alpha too please?

birkskyum:
Sure, i've moved svelte-runes -> svelte, and updated to next.7

frameworks/keyed/svelte-classic/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "js-framework-benchmark-non-keyed-svelte",
+  "name": "js-framework-benchmark-svelte",
```
trueadm, line 2: Is this change right?

trueadm:
@birkskyum We published a new version that should improve code size, if you don't mind :D

birkskyum:
i updated it now

krausest:
On the left the new results, on the right the results for .1:
Screenshot 2023-11-19 at 9 21 40 PM
The difference between the two new implementations is pretty small. Indeed I haven't seen two implementations with such similar results yet and I checked that they were computed from real values that actual differ.

trueadm:
@krausest I saw the latest results are up and for me I get some strange results locally – inferno and vanillajs-wc both score slower than Svelte 5 alpha, no matter how many times I run it. I guess the whole table will update when the new Chrome version is out anyway.

birkskyum:
Really impressive results

PR #1502, opened 2023-11-17
rename react->react-classes

birkskyum:
"react" appear like the gold-standard, but since it's the initial React Classes API it's really not how people are recommended to use react anymore.

Renaming to react-classes makes it more clear what's inside.


krausest:
Thanks. No improvement at all :) 
Good to see that results are really pretty consistent in this case (on the left are the updated keyed results on the right are the old results):

PR #1503, opened 2023-11-18
Sprae framework

dy:
It adds basic test for [sprae](https://github.com/dy/sprae).
Within analogs: alpine, petite-vue it shows locally best results.
@krausest @trueadm


krausest:
How about a keyed version? Would that be possible?

dy:
No, the method used here is direct mapping of array items to elements, so it's not domdiffing but direct update. I can't think how keying can help here, so it seems fine as is. Just out of curiosity I can put non-keyed version to keyed competitors - it still has good results compared to petite-vue / alpine.

PR #1506, opened 2023-11-19
Update cample to v3.2.0-beta.6

antonmak1:
Corrected the code a little. I think select row will be a little faster :)

### DIFF
--- a/frameworks/keyed/cample/package-lock.json
+++ b/frameworks/keyed/cample/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-cample",
-  "version": "3.2.0-beta.5",
+  "version": "3.2.0-beta.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "js-framework-benchmark-cample",
-      "version": "3.2.0-beta.5",
+      "version": "3.2.0-beta.6",
       "license": "MIT",
       "dependencies": {
-        "cample": "3.2.0-beta.5"
+        "cample": "3.2.0-beta.6"
       },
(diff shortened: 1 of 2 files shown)


krausest:
Here's the comparison (left new, right last run) - yes maybe select got a bit faster:

krausest:
Just as a note: Transpilation is mixing up the code anyway. Left your version, on the right the old version:
Screenshot 2023-11-24 at 8 07 58 AM

krausest:

antonmak1:
> Here's the comparison (left new, right last run) - yes maybe select got a bit faster: Screenshot 2023-11-24 at 8 02 25 AM

Thank you very much! Okay, as you wish. It can be seen that the select row results are still a little better - that’s good! But, I tried to introduce new template processing for the loop, and it can be seen that the results of creating rows have become slower. I'm working on this now, optimizing the new code. I think that before the new year we will be able to achieve a result of 1.08 - 1.09.

antonmak1:
> Just as a note: Transpilation is mixing up the code anyway. Left your version, on the right the old version: Screenshot 2023-11-24 at 8 07 58 AM

Strange, I didn’t change main.js in the new version. I haven’t used Switch case for a long time to process events, maybe 2-3 months. Maybe it's some other main.js?

krausest:
I forgot to git pull before running the benchmark for the updated frameworks. For frameworks that have a real build step (npm run build-prod) there should be no change. But the displayed version for cample was wrong.
 
antonmak1:
Thanks a lot! Yes, this is approximately how the results are on my computer now. I was just thinking today why select row is bigger. Happens.

PR #1507, opened 2023-11-20
frei update v1.0.18 { diff logical small-scale optimization }

AimWhy:
screenshot Can include：？ @krausest 

「Startup metrics (lighthouse with mobile simulation)」
「Memory allocation in MBs ± 95% confidence interval 」


krausest:
Here's the comparison (left new, right last run) - no diff I'd say:
Full results are in the results table.

krausest:

krausest:
I forgot to git pull before running the benchmark for the updated frameworks. For frameworks that have a real build step (npm run build-prod) there should be no change.
 

PR #1508, opened 2023-11-20
feat(plaited): uses new features and improved  function for templating

EdwardIrby:
This PR is for 5.3.0 of Plaited
This version
- drops **useStore** (doesn't make sense with library architecture and coding style)
- splits **$.render** into **$.render** and **$.insert** The later which use the same interface pattern as ` insertAdjacentElement` but takes a spread like `append | prepend | before | after`
- introduces a new **$.clone** method that can be used to clone an element with a **data-target** attribute and apply a callback to the clone before usage. Useful for handling large list and reducing parse time. 
- makes use of a streamlined **createTemplate** aka **h** function to improve templating script times. 

Glad this benchmark repo exist @krausest! It's been super useful in improving my library.

### DIFF
--- a/frameworks/keyed/plaited/package-lock.json
+++ b/frameworks/keyed/plaited/package-lock.json
@@ -5,16 +5,16 @@
   "packages": {
     "": {
       "dependencies": {
-        "plaited": "5.0.3"
+        "plaited": "5.3.0"
       },
       "devDependencies": {
-        "esbuild": "0.19.5"
+        "esbuild": "0.19.6"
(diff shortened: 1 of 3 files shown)


krausest:
Here's the comparison (left new, right last run) - a bit faster:

krausest:

krausest:
I forgot to git pull before running the benchmark for the updated frameworks. For frameworks that have a real build step (npm run build-prod) there should be no change.
 

PR #1514, opened 2023-11-24
`karyon` update

syduki:
This updates the `karyon` module version, also implementation due to removal of the so-called "selector" API, a peculiar feature not really compliant with the module concept. Frankly speaking, it was just a crutch for [#800](https://github.com/krausest/js-framework-benchmark/issues/800) issue. However, a naive approach will hurt performance, thus a similar "selector" was added into the implementation itself, hopefully it won't be a subject for new note, sort of "explicit select delegation".

### DIFF
--- a/frameworks/keyed/karyon/package.json
+++ b/frameworks/keyed/karyon/package.json
@@ -7,7 +7,7 @@
     "issues": [801]
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


krausest:
On the left new keyed results on the right last results:
Screenshot 2023-11-28 at 7 41 30 AM

PR #1515, opened 2023-11-25
Update Strve to v6.2.0

maomincoding:
Strve.js v6.2.0 is released! 

[https://github.com/strveJs](https://github.com/strveJs)


maomincoding:
Thanks for the hard work, author.

maomincoding:
Hello~ Can this branch be merged? thanks！ @krausest

krausest:
... except that your implementation isn't working as advertised:
```
npm run isKeyed -- --headless true keyed/strve
...
strve-v6.2.0-keyed is non-keyed for 'run benchmark' and non-keyed for 'remove row benchmark' and non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
ERROR: Framework strve-v6.2.0-keyed is not correctly categorized
```
Should we move it to the non-keyed section?

maomincoding:
I will pull down the code and retest it

maomincoding:
The test results on my end are the same as yours. I will re release the PR on my end. Thank you！

maomincoding:
My ultimate expected outcome is in keyed.

PR #1517, opened 2023-11-28
New implementation: non-keyed/mutraction

tomtheisen:
I did my best to follow the instructions.  I had to hack up the scripts a little bit (locally) because puppeteer is dead-set on launching chrome from the wrong path.


krausest:
Thanks - [results](https://krausest.github.io/js-framework-benchmark/current.html) are updated.

PR #1518, opened 2023-11-28
update quel benchmark, optimize "select row"

mogera551:
This PR update quel benchmark, optimize "select row".


krausest:
Thanks - [results](https://krausest.github.io/js-framework-benchmark/current.html) are updated.

PR #1521, opened 2023-12-02
update Strve v6.2.6

maomincoding:
🚀 Strve.js v6.2.6 is released!
A approachable, fast, flexible and lightweight JavaScript library for building user interfaces.

[https://github.com/strveJs/strve](https://github.com/strveJs/strve)

This is the document address: [https://strvejs.github.io/strve-doc/](https://strvejs.github.io/strve-doc/)


krausest:
Here's a comparison (left update, right old version). It seems to be faster than the old version:

maomincoding:
Thanks!

PR #1522, opened 2023-12-05
Update rendrjs version

twharmon:
Update rendrjs version and implementation


krausest:
Here's a comparison (left update, right old version). It seems to be close to the old version:

PR #1523, opened 2023-12-06
Add non-keyed/openui5 for completeness

pskelin:
Hi @krausest 

with #1488 openui5 was moved to the keyed section. This has resulted in worse performance and a bug in the rendering code was identified. Until it is fixed, it makes sense to keep the non-keyed version of the benchmark for comparison like other frameworks that can render both keyed and non-keyed.

The change also puts a comment in the code where keyed/non-keyed is decided.


krausest:
Tanks. [Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated.

PR #1524, opened 2023-12-07
Update qwik to 1.3

birkskyum:
[Release notes](https://github.com/BuilderIO/qwik/releases/tag/v1.3.0)


krausest:
Here's a comparison (left update, right old version). It seems to be a bit slower:

PR #1533, opened 2023-12-13
Update cample to v.3.2.0-beta.7

antonmak1:
Optimized the assembly of templates for rows, made select row a little faster. I think this will work for 1.10, but not 1.08-1.09. I don’t know anymore, it’s unlikely that before the new year it will be possible to implement the code that I had in mind, but I think the result will be cool. Due to the specific syntax, “select row” is of course too slow, result certainly stands out from all the green ones, but I think there are still optimization ideas, so that’s fine.


krausest:

antonmak1:
The results got a little worse. I continue to work.

PR #1538, opened 2023-12-21
Add global state implementation for Rendrjs

twharmon:
This implementation uses global state only for table rows and selected item. Some benchmarks are an improvement, but overall about the same as the implementation without global state.


krausest:
Thanks. Here's a screenshot from the results (left new/updated keyed version, right old keyed version):
Screenshot 2023-12-23 at 4 22 57 PM

PR #1539, opened 2023-12-23
Update Svelte dependencies

benmccann:


krausest:
Thanks. On the left new keyed version, on the right the old version:
Screenshot 2023-12-23 at 5 52 56 PM

PR #1540, opened 2023-12-24
Update DLight.js to 1.0.0-alpha

IanDxSSXX:
[DLight.js](https://github.com/dlight-js/dlight) is about to drop its first stable release. It's already in the testing phase. This major update includes a huge boost in performance, less memory usage, a smaller bundle size(currently 3.2KB), plus some cool new features to make the developer experience even better. 
Super excited to see how this DX-first framework performs across the board!
This pull request updates both keyed and non-keyed old dlightjs folder with dlight's 1.0.0-alpha.1 version and adds a subview implementation of the keyed one.
Here're some results I run locally, hope it'll perform as expected:
iShot_2023-12-24_05 45 31
iShot_2023-12-24_05 45 48


krausest:
The isKeyed test fails since the html structure isn't correct.
Seems like ariaHidden("true") is output as ariahidden="true" instead of aria-hidden="true". Can you please take a look at it and add the aria-hidden=true on the span? 
```
Expected to find 'aria-hidden'=true on span in third td, but found null
ERROR: Framework dlightjs-v1.0.0-alpha.35-keyed html is not correct
dlightjs-v1.0.0-alpha.35-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as keyed in the results
Expected to find 'aria-hidden'=true on span in third td, but found null
ERROR: Framework dlightjs-subview-v1.0.0-alpha.35-keyed html is not correct
dlightjs-subview-v1.0.0-alpha.35-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as keyed in the results
Expected to find 'aria-hidden'=true on span in third td, but found null
ERROR: Framework dlightjs-v1.0.0-alpha.18-non-keyed html is not correct
dlightjs-v1.0.0-alpha.18-non-keyed is non-keyed for 'run benchmark' and non-keyed for 'remove row benchmark' and non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results

```

IanDxSSXX:
> The isKeyed test fails since the html structure isn't correct. Seems like ariaHidden("true") is output as ariahidden="true" instead of aria-hidden="true". Can you please take a look at it and add the aria-hidden=true on the span?
> 
> ```
> Expected to find 'aria-hidden'=true on span in third td, but found null
> ERROR: Framework dlightjs-v1.0.0-alpha.35-keyed html is not correct
> dlightjs-v1.0.0-alpha.35-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as keyed in the results
> Expected to find 'aria-hidden'=true on span in third td, but found null
> ERROR: Framework dlightjs-subview-v1.0.0-alpha.35-keyed html is not correct
> dlightjs-subview-v1.0.0-alpha.35-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as keyed in the results
> Expected to find 'aria-hidden'=true on span in third td, but found null
> ERROR: Framework dlightjs-v1.0.0-alpha.18-non-keyed html is not correct
> dlightjs-v1.0.0-alpha.18-non-keyed is non-keyed for 'run benchmark' and non-keyed for 'remove row benchmark' and non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
> ```

Sorry didn't notice that, should be fixed with the latest commit.

PR #1543, opened 2023-12-27
update Mikado package.json repository

ts-thomas:
I have updated the repository to its new location. Nothing more has changed.


ts-thomas:
Added a hotfix version which uses the `eventCache` properly. I don't know if event delegation is covered by the benchmark measurement, so it probably makes no difference in the result.

krausest:
Thanks. Here's a screenshot (left keyed updated versions, right old versions):
Screenshot 2023-12-31 at 10 30 45 AM

PR #1544, opened 2023-12-29
Upgrade mutraction to 0.25.0

tomtheisen:
Upgrades mutraction version.

There are some superficial changes to the app, but there's one thing in particular I want to call attention to.  This implementation uses a sparse array.  I can't find anything specifically allowing or disallowing this.  I searched old issues and notes, and can't find a precedent.  I'm not trying to do anything sneaky, and I actually use sparse arrays in real life.  If it's against the rules for submissions, I'll take it out.

It also adds `--minify` to the final esbuild output for some "free" byte savings.


krausest:
Thanks, [results](url) are updated.

PR #1546, opened 2023-12-29
feat: add vue-jsx

manyuemeiquqi:
I add vue-jsx framwork


krausest:
Thanks. Your implementation had two issues: 
1. Selecting rows didn't work. Please review my fix (https://github.com/krausest/js-framework-benchmark/commit/2a81628b2c55d08002cfb1775e83252b25bfc1f4), I'm not sure if that is the reason for slow performance with select rows.
2. The container class was missing.
I updated to vue 3.4.0 since there's an open PR to push vue to 3.4.0. So I though we could jump to that version.

Here's a screenshot (left new keyed versions + vue as a reference, right old versions):
Screenshot 2023-12-31 at 10 30 45 AM

Even without select rows it doesn't look like using jsx for vue is performance wise a good idea.

manyuemeiquqi:
Thanks, this is pretty much the same as my local test .
I know JSX + Vue performance is not good, I juset approve JSX performance is particularly bad, and there seems has room for optimization

PR #1550, opened 2024-01-02
Update rendrjs version

twharmon:
Update to v0.2.11


krausest:

PR #1553, opened 2024-01-05
upgrade to Svelte 5.0.0-next.28

benmccann:


krausest:

krausest:
Here's the comparison new to old for keyed implementation.
I suppose the difference is mostly noise.
Screenshot 2024-01-09 at 8 05 29 AM

PR #1554, opened 2024-01-06
Glimmer next

lifeart:


krausest:

krausest:
Here's the comparison with the new version on the left and the old glimmer version on the right for the keyed implementation:
Screenshot 2024-01-09 at 8 05 29 AM
Can we archive glimmer2-beta then (the beta is ~3 years old)?

lifeart:
@krausest, thank you for updates! glimmer2-beta should be upstreamed soon (there is a error-recovery work needs to be landed https://github.com/glimmerjs/glimmer-vm/pull/1501)

PR #1555, opened 2024-01-07
Sprae 8.1.2

dy:
This adds few enhancements to objects, doesn't really improve performance, but to be in sync.

Also @krausest I wonder was it really necessary to remove petite-vue - it is unique in class framework, one good viable alternative to alpine - it doesn't seem to be needed to be updated frequently, since it's state of art.
It would be also help to compare alpine, petite-vue, sprae as 3 similar frameworks in class. Wdyt adding it back?


krausest:

krausest:

dy:
I’s sad there’s no same class competitors left for sprae in non-keyed.
petite-vue is gone, alpine has only keyed test. Should I move sprae to
keyed mb?

PR #1559, opened 2024-01-10
upgrade Svelte to 5.0.0-next.33

benmccann:


birkskyum:
I think since the current version (next.8) that the svelte team abandoned this idea of using classes, so it would be great to have that update in the keyed/svelte example.

https://github.com/krausest/js-framework-benchmark/blob/d149b6593d3165e83ccbba9cf29d234e901258c1/frameworks/keyed/svelte/src/Main.svelte#L99-L104

trueadm:
@birkskyum Nope. We still recommend classes for where they make sense and this is a great use-cases.

birkskyum (approved):

trueadm:
Although we should probably update to alpha 35 now it's out before merging this. That contained some beefy internal changes.

krausest:
I've already merged. But I'll push it to .35. Please don't update this PR.

krausest:
No real difference, but it's already damned fast...

PR #1560, opened 2024-01-11
feat(angular): use provide zoneless api

eneajaho:


birkskyum:
When using signals in angular, will the `ɵprovideZonelessChangeDetection` be the recommended path?

eneajaho:
@birkskyum 

This is still a private API and is exposed as private for the community to try and find issues with it. 
That's why I updated only this example.

krausest:
I'm getting the following error:
```
> js-framework-benchmark-angular-nozone@0.0.0 build-prod
> ng build --configuration production

Application bundle generation failed. [12.550 seconds]
✘ [ERROR] NG8: Type 'WritableSignal<Data[]>' must have a '[Symbol.iterator]()' method that returns an iterator. [plugin angular-compiler]

    src/app/app.component.html:31:26:
      31 │             @for (item of data; track item.id) {
         ╵                           ~~~~

  Error occurs in the template of component AppComponent.

    src/app/app.component.ts:15:17:
      15 │     templateUrl: './app.component.html',
         ╵                  ~~~~~~~~~~~~~~~~~~~~~~


ERROR: Rebuilding  keyed/angular-cf-nozone was not successful

node:internal/process/esm_loader:40
      internalBinding('errors').triggerUncaughtException(
                                ^
ERROR: Rebuilding  keyed/angular-cf-nozone was not successful
(Use `node --trace-uncaught ...` to show where the exception was thrown)

Node.js v20.9.0
```
Can you take a look at it?

frameworks/keyed/angular-cf-nozone/src/app/app.component.html
```diff
@@ -28,16 +28,19 @@ <h1>Angular {{version}} (no Zone.js)</h1>
...
-            {
-            <tr [class.danger]="item.id === selected">
-                <td class="col-md-1">{{item.id}}</td>
-                <td class="col-md-4">
-                    <a href="#" (click)="select(item, $event)">{{item.label}}</a>
-                </td>
-                <td class="col-md-1"><a href="#" (click)="delete(item, $event)"><span class="glyphicon glyphicon-remove" aria-hidden="true"></span></a></td>
-                <td class="col-md-6"></td>
-            </tr>
+            @for (item of data; track item.id) {
```
eneajaho, line 31:
```suggestion
            @for (item of data(); track item.id) {
```

eneajaho:
@krausest 
It should be okay now.

krausest:
That worked (almost). I changed selected into selected() one line later, so that it compiled, but the "update every 10th row" button doesn't work.

krausest:
Found it. It must also be data() in the component or reuse the local variable (which is what I'll be using):
```
    update() {
        let data = this.data();
        for (let i = 0; i < data.length; i += 10) {
            data[i].label += ' !!!';
        }
        this.data.set(data);
    }
```
I'll update

birkskyum:
any perf difference?

krausest:
Results are updated and I'm afraid to say they aren't going in the right direction. Here's a comparison (left: keyed new results, right: keyed old results):

birkskyum:
yeah, I expected this. It was why i suggested to update the much slower signals example instead of the fast NgNoZone example. The perf penalty likely comes from the signals.

birkskyum:
Also, this is possibly now the only implementation in the entire benchmark that use private APIs, not representative of how users are expected to use the framework.

eneajaho:
In fact, what's making it worse it's the scheduling, that it is using a setTimeout (which is async) to run change detection, instead of using the cdr.detectChanges() which is sync. 

We can rever back ofc, so at least the zoneless doesn't give the wrong idea of being slow (it definitely it's not)

krausest:
It's reverted. I'll leave it in the result table until the next official run (where it'll appear reverted) if you don't mind. Chrome 121 should be out in a few days.

PR #1561, opened 2024-01-11
update glimmer-next [fix remove rows regression]

lifeart:
Fixed `clear row regression` (44ms -> 25ms)

image


krausest:
Results are updated, clear rows is now much better. Here's a comparison (left: keyed new results, right: keyed old results):

PR #1562, opened 2024-01-11
Update Mikado

ts-thomas:
Do you know why in your latest test the timings "first paint" of Mikado is `179.0 (4.57)`?


krausest:

First paint looks much better now. I'll keep an eye on the first paint metric to find out what caused the outlier.

ts-thomas:
Thanks a lot. I never looked too much into the static assets before, but it would be nice don't including the whole bootstrap library. I never would do that in a project. This has also too much impact on page loading. It might be better to have smaller bootstrap build just including styles which are required.

PR #1564, opened 2024-01-11
Fix rendrjs preloadicon bug; update rendrjs

twharmon:


krausest:
Results are updated and they look good! 
Here's a comparison (left: keyed new results, right: keyed old results):

PR #1566, opened 2024-01-12
Wamup count from config used in assertions

twharmon:
Changing `config.WARMUP_COUNT` will break the benchmarks. This fixes it.


krausest:
Thanks!

PR #1570, opened 2024-01-14
svelte: update on:click on onclick

benmccann:


krausest:

PR #1571, opened 2024-01-15
Warmup counts cfg

twharmon:
I noticed that in `webdriver-ts/src/common.ts` the `config` const contains the number of warm-up runs, but the partial update cpu benchmark does not use that number and instead uses a hard-coded `3`.

Should we move the warm-up count from the shared `config` into the `cpuBenchmarkInfosArray` in `webdriver-ts/src/benchmarksCommon.ts`? This makes it so that the number of warm-ups for each benchmark can be individually configured (similar to how the number of additional timed runs is configured).

A few more changes had to be made to accommodate this, so you see changes in the `BenchmarkInfoBase` type definition. I did my best to make sure that the final description of each cpu benchmark stays exactly as it was before.


krausest:
Thanks. Chrome 121 should be out the next few days. I'll take a look at it after the official run for chrome 121.

PR #1572, opened 2024-01-16
update mikado v0.8.222

ts-thomas:
Includes an improvement of default DOM attributes state and unused attributes


krausest:

PR #1574, opened 2024-01-17
Update Mikado

ts-thomas:
The last feature "skip default DOM state" was way too optimistic :)
I found a situation where this optimization leads into a false state, because it collides with another performance ruleset. Mikado heavily uses runtime optimizations. Almost every optimization has its own cost. Since we can't predict the next render iteration (in a real world scenario) it isn't useful to fully apply every optimization in expectation a component will be recycled on next render loop. That's why Mikado uses a technique called "progressive enhancement", on which each iteration adds optimizations successively:

1. factory phase
2. vdom creation phase
3. full vdom access

To just "enhance" components which are already recycled for at least one time is a better approach. This limitation is because to reach "phase 3." isn't for free. I already tried several times to inject logic earlier (without adding too much cost). But this time I found a solid solution, because the "default DOM state" issue get me the hint. The new version is able to save almost 50% of the time needed for "phase 2." by doing this in "phase 1." by almost zero cost. All creation render tasks don't benefit from it, but "select", "swap", "update" should benefit from it. On specific render task I got an improvement of twice as fast. In this js-framework benchmark it might be a plus of 0.2 :)

### DIFF
--- a/frameworks/keyed/mikado-proxy/package-lock.json
+++ b/frameworks/keyed/mikado-proxy/package-lock.json
@@ -7,7 +7,7 @@
       "name": "js-framework-benchmark-keyed-mikado-proxy",
       "license": "Apache-2.0",
       "dependencies": {
-        "mikado": "^0.8.222"
+        "mikado": "^0.8.227"
       },
       "devDependencies": {
         "google-closure-compiler": "^20230802.0.0"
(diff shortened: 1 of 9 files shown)


krausest:
Benchmark is running. I‘m going to report results tomorrow.

krausest:
Screenshot 2024-01-19 at 7 36 17 AM

ts-thomas:
Thanks a lot. Looks like you get different results on your Mac, in my test it was before vanillajs :)

PR #1576, opened 2024-01-18
Update uhtml to v4.4

WebReflection:
This MR would like to benchmark latest uhtml v4 with the following changes:

  * even lower memory consumption
  * linear render (interpolations are never looped more than once)
  * overall simplified logic but also made it more robust for fragment edge cases
  * hints around both fragments (not used in this demo), via `<!--<>-->` and `<!--</>-->`, and array of items (used in this demo), via `<!--[amount-of-items]-->` pinned comments

Thanks for considering updating this library to its greatest version *yet* 😉

### DIFF
--- a/frameworks/keyed/uhtml/package-lock.json
+++ b/frameworks/keyed/uhtml/package-lock.json
@@ -10,13 +10,13 @@
       "license": "Apache-2.0",
       "dependencies": {
         "js-framework-benchmark-utils": "^0.3.2",
-        "uhtml": "^4.0.0"
+        "uhtml": "^4.4.1"
       },
       "devDependencies": {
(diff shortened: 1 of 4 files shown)


krausest:
Benchmark is running. I‘m going to report results tomorrow.

krausest:
Screenshot 2024-01-19 at 7 36 17 AM

WebReflection:
wow ... I actually have **1.00** with latest VS **1.02** with previous ... how bizarre!

Anyway, the latest design/logic is there to stay and there's very little I can optimize further but at least I hope the memory consumption and/or bootstrap time got better there, as much as it did on my local machine.

Thanks for the merge though, appreciated 👋

PR #1580, opened 2024-01-20
Update Mikado

ts-thomas:
I've done some inspection on a Macbook and I found some strange performance issue which prevents the template functions to get optimized by JIT. Probably this will improve the run on your Mac.

My results (keyed):
image


krausest:
I guess this isn't exactly what you hoped for...
Screenshot 2024-01-20 at 8 49 00 PM

ts-thomas:
It is never good enough, but I am happy with that result 👍

PR #1581, opened 2024-01-22
Update goui version

twharmon:
This should show some all around improvements


krausest:
Indeed! (left new version, right previous version)
Screenshot 2024-01-22 at 9 33 14 PM

PR #1582, opened 2024-01-22
Update rendrjs versions

twharmon:
Showed some improvements on my machine.


krausest:
Pretty close, but maybe a bit faster (left new, right old version):
Screenshot 2024-01-22 at 9 33 14 PM

PR #1584, opened 2024-01-24
Update Mikado v0.8.327

ts-thomas:
Some improvements here and there, all in all a nice package. Creation tasks and the select task could benefit from it.


krausest:
Results will follow (most likely tomorrow).

krausest:
Not bad 😄 .
Here's a comparison (updated keyed version on the left, old keyed version on the right):
Screenshot 2024-02-01 at 7 59 44 PM

PR #1594, opened 2024-01-29
Update cample to v3.2.0-alpha.7

antonmak1:
I think this will be 1.09 (I also committed a version there that improved the code a little)

### DIFF
--- a/frameworks/keyed/cample/package-lock.json
+++ b/frameworks/keyed/cample/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-cample",
-  "version": "3.2.0-alpha.6",
+  "version": "3.2.0-alpha.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


krausest:
Results will follow (most likely tomorrow).

antonmak1:
> Results will follow (most likely tomorrow).

ok

krausest:
Here's a comparison (updated keyed version on the left, old keyed version on the right):
Screenshot 2024-02-01 at 7 48 59 PM

antonmak1:
@krausest  For 7 months now, select row has been crap on the result, so I was sure that it would be 1.09, but in the end it’s 1.10. This is some kind of joke). Okay, this is all technically explained, so I will continue to try to find a solution for select row. Almost everything is perfect now, except create many rows and select row.

antonmak1:
I'll write some code and try something new. I think now select row if I fix it, it won’t be far from 1.05.

antonmak1:
@krausest Thank you very much for creating the results!

PR #1598, opened 2024-02-03
upgrade and add issue 801 to pota

titoBouzout:


krausest:

krausest:
Thanks, a bit better than last run. Here's a comparison (left new keyed results, right old keyed result).

titoBouzout:
Thanks Stefan, I was going for an improvement with memory usage. There seems to be room for improving performance, but I also don't want to go crazy about it. I will see what can be done

PR #1600, opened 2024-02-10
feat: add vue-vapor

sxzz:
Repo: https://github.com/vuejs/core-vapor


antonmak1:
Implementation of a framework without a virtual DOM, respect! Promote the idea in web development!

birkskyum:
Great work! I get these relative numbers for the keyed:

Screenshot 2024-02-11 at 23 55 10

sxzz:
Here are the tests on my local machine. I'm investigating what reason caused it's slower than Vue 3.
image

sxzz:
Here is the latest benchmark on the latest commit. In general, Vue and vue-vapor are currently quite similar, but Vapor still has significant room for optimization.

image

krausest:
Thanks. Here's a screenshot (left new keyed results, right old keyed result).

sxzz:
Thanks for your work!

I would like to remind those who are concerned about Vue Vapor that the current performance of Vapor does not necessarily reflect the stability of the stable version.

PR #1601, opened 2024-02-11
Update cample to v3.2.0-alpha.11

antonmak1:
Now, it should definitely be 1.09!


antonmak1:
I also updated the dependencies (I haven’t updated them for almost a year) and added to the version there so that the code would be esnext standard.

krausest:

krausest:
Thanks. Here's a comparison (left new keyed results, right old keyed result).

PR #1603, opened 2024-02-11
Move vue, vue-pinia to vite

birkskyum:
This streamlines the different vue entries, so they now all run in Vite, which makes sense due to the significant overlap of Vite/Vue community. Basically the vite config is way simpler than webpack:

```ts
//vite
export default defineConfig({
  base:'/frameworks/keyed/vue/dist/',
  plugins: [vue()],
})
```

vs the incumbant webpack config:

```ts
//webpack
const path = require('path')
const { VueLoaderPlugin } = require('vue-loader')
const webpack = require('webpack')

module.exports = (env = {}) => ({
	mode: 'production',
	entry: path.resolve(__dirname, './src/main.js'),
	output: {
		path: path.resolve(__dirname, './dist'),
		publicPath: '/dist/'
	},
	resolve: {
		alias: {
			// this isn't technically needed, since the default `vue` entry for bundlers
			// is a simple `export * from '@vue/runtime-dom`. However having this
			// extra re-export somehow causes webpack to always invalidate the module
			// on the first HMR update and causes the page to reload.
			'vue': '@vue/runtime-dom'
		}
	},
	module: {
		rules: [
			{
				test: /\.vue$/,
				use: 'vue-loader'
			},
		]
	},
	plugins: [
		new VueLoaderPlugin(),
		new webpack.DefinePlugin({
			__VUE_PROD_DEVTOOLS__: JSON.stringify(false)
		})				
	],
})

```

The migration to Vite also make these entries similar to the new entry for vue-vapor, which is configured something like this:

```ts
// vite.config.ts
export default defineConfig({
  base:'/frameworks/keyed/vue-vapor/dist/',
  plugins: [
    vue({
      vapor: true,
    }),
  ],
})
```

related:
- #1600


krausest:

krausest:
Thanks. Here's a screenshot (left new keyed results, right old keyed result).

PR #1609, opened 2024-02-20
Bump to Svelte next.61

birkskyum:
Since below mentioned issue is resolved, "svelte 5 classic" can now (next.61) run the and significantly speed up the code from "svelte 4" without any modification.

- https://github.com/sveltejs/svelte/issues/9521

This is really impressive @trueadm ! Svelte 5 is gonna be a great success with this level of backwards compat

PR #1610, opened 2024-02-20
svelte next.64

birkskyum:
apparently there was bugs in the next.61 release #1609 , which are fixed here.


krausest:
Will be published soon in the chrome 122 results.

krausest:
That would be nice, but in reality I need two nights to complete both keyed and non-keyed frameworks. So hopefully on Friday...

krausest:
Now they're available: https://krausest.github.io/js-framework-benchmark/2024/table_chrome_122.0.6261.69.html

PR #1612, opened 2024-02-28
Upgrade: upgrade dlight to 1.0.0-next

IanDxSSXX:
* Upgrade dlight to version 1.0.0-next
* Removed subview implementation since it's a performance comparison

Performance should remain the same with dlight's alpha version.


krausest:

krausest:

PR #1613, opened 2024-02-28
Bump all vue entries

birkskyum:
vue is making gains, so this bumps all the all the entries affected.

Screenshot 2024-02-28 at 14 42 47


krausest:

krausest:

PR #1614, opened 2024-02-28
Update cample to 3.2.0-alpha.15

antonmak1:
I think that it will no longer be 1.09, but will be somewhere around 1.06 - 1.07! Even if it’s 1.08, I think that it will be quite possible to finish up to 1.03 - 1.06 in the near future. I found a lot of stupid errors in the code, I think there will be more. I think I’ll finish it up to 1.03 - 1.06, although I see that I seem to have accelerated "select row" to almost maximum on the local machine, and the values are higher than others. In principle, I have almost completed the maximum of the framework. Maybe I’ll do a couple of PRs and we can say that the implementation will be ready. It will only be updated as new functionality is introduced, such as working with asynchronous data, etc.

P.S. Well, this time I’m not even 99.99999% sure, but 109% sure)

### DIFF
--- a/frameworks/keyed/cample/package-lock.json
+++ b/frameworks/keyed/cample/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-cample",
-  "version": "3.2.0-alpha.14",
+  "version": "3.2.0-alpha.15",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "js-framework-benchmark-cample",
-      "version": "3.2.0-alpha.14",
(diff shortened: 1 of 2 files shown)


krausest:

antonmak1:
ok

krausest:
With the exception of vue-pinia all results including vanillajs were slower than in the full run. I haven't any proof yet, but it seems to me that security patches after a major update can cause some slowdown for the benchmark.
So cample might have improved more than the 0.1 improvement sugest.

antonmak1:
OK. Well, now the result is 1.09 at least.

PR #1615, opened 2024-03-01
Update Strve v6.7.0

maomincoding:
Update Strve v6.7.0！


krausest:

krausest:
With the exception of vue-pinia all results including vanillajs were slower than in the full run. I haven't any proof yet, but it seems to me that security patches after a major update can cause some slowdown for the benchmark.

PR #1616, opened 2024-03-01
updated: Riot.js version

GianlucaGuarini:
Update to Riot.js 9


krausest:

krausest:
Sorry, you're right. There was an additional empty script element in the index.html. 
The flag is applied when either the page doesn't work with a CSP or a violation is reported by the CSP.
Chrome reported that script as violating the CSP. I'll remove the empty script and the #1139 flag.
I'll open an issue to check the flag for the other implementations and add the check in the rebuild script.

PR #1618, opened 2024-03-02
Add a lefthook and pre-commit hook

nakrovati:


krausest:
I'm not merging this. We've had that discussion before. I don't accept that my code is formatted without my consent.

PR #1619, opened 2024-03-05
refactor(cli): improve code typing and directory structure

nakrovati:
1. Move all CLI files to the `cli` directory.
2. Replace `takeWhile.d.ts` with jsdoc types for `takeWhile` function.
3. Fix types.
4. Remove the ability to call CLI functions directly (e.g. calling `node copy.js`). Now the same-named functions from CLI commands are exported, and all commands can be called via `cli.js`.

And some other improvements.


nakrovati:
# Short answer
It is better to make the "rebuild-ci" script either like this: `node cli.js rebuild-single --ci` and call `nr rebuild-ci -f keyed/svelte...`; or like this: `node cli.js rebuild-single --ci -f` (pass the `-f` flag as the last argument) and call `nr rebuild-ci keyed/svelte...`.

# Unfolded answer

It can be done, but `options` in commander.js is an object with flags, and when passing an array of arguments as in the code below, you can't swap them around and will have to strictly specify the order of flags and arguments in `action` in `cli.js`.

If you do as it was __before__ only with commander.js without extra code, there are 2 problems:

1. 

```typescript
program
  .command("rebuild-single")
  .option("--ci [boolean]", "", false)
  .argument("<...frameworks>", "frameworks")
  .action((options, ci) => {
    console.log(options, ci);
    // rebuildSingleFramework(options;)
  });
```

This code will work when called as it should, but you will need to specify `--ci true` (extra `-true`) in the script, you cannot pass flags in random order and you cannot pass other flags after the frameworks argument.

2. 

```typescript
program
  .command("rebuild-single")
  .argument("<...frameworks>", "frameworks")
  .option("--ci [boolean]", "", false)
  .action((options, ci) => {
    console.log(options, ci);
    // rebuildSingleFramework(options;)
  });
```

You won't be able to pass arguments through flags in random order and will have to specify `--ci` when calling the script.

P.S: you can abbreviate `--ci true` as just `--ci`, and `--frameworks` to `-f`.

PR #1621, opened 2024-03-06
Bump Qwik 1.3 -> 1.5

birkskyum:


krausest:
Thanks. I'll post an update when results are ready.

krausest:
Here's a screenshot (left = new keyed version, right = old keyed version)
Please note that in the incremental runs vanillajs is also a bit slower than in the full chrome 122 run, maybe due to security patches. We'll have to wait for the next full run to get a fair comparison.
Screenshot 2024-03-15 at 12 46 35

PR #1623, opened 2024-03-06
Update cample to 3.2.0-alpha.25

antonmak1:
I'll make one more pull request before 123. I don’t know, we’ll see how it turns out now)


antonmak1:
I'll upload the commit today

krausest:
Thanks. I'll post an update when results are ready.

krausest:
Here's a screenshot (left = new keyed version, right = old keyed version)
Please note that in the incremental runs vanillajs is also a bit slower than in the full chrome 122 run, maybe due to security patches. We'll have to wait for the next full run to get a fair comparison.
Screenshot 2024-03-15 at 12 46 35

antonmak1:
@krausest It’s strange that select row is like this now, when I corrected a lot of things there so that it would be normal. But, okay, I don't know. Even taking into account the delayed results due to security.

antonmak1:
"dist" may have remained from the old version, but, as I understand it, this is impossible, because Once upon a time you fixed it. I don’t remember, maybe I wrote about this about 5 months ago and you definitely corrected it then, strange. I will continue to work on the code. Most likely, I again missed something in the code.

antonmak1:
Okay, never mind, thank you very much anyway. I'll find a way to fix this anyway. This is a 99% error in the framework code, which quickly prevents the code from working.

PR #1625, opened 2024-03-10
Add Vugu implementation

lasagnats:
Add benchmark implementation (keyed & non-keyed) for [Vugu framework](https://www.vugu.org/)


krausest:
Thanks. Something looks odd with vugu. Any click causes a rerender where the css styles are removed and then reapplied. Is that intentional? (I really doubt that measure the performance makes sense when each click causes such glitches.)

Additionally the non-keyed version behaves incorrect. Clicking on create 1,000 rows twice should show row numbers starting with 1,001, but they still start with 1. The keyed version is correct.

krausest:
Closing this PR due to inactivity.

PR #1630, opened 2024-03-15
clean up solid / solid-store devdeps

birkskyum:
make the setup of solid and solid-store identical and updated. This updates the solid-store as well. shouldn't cause significant change


krausest:

Issue #1620, opened 2024-03-05
Check #1139 flag


krausest:
The check if an implementation works with a content security policy hasn't been run regularly.

The following frameworks fail with CSP and currently haven't flag #1139 set with the following error "Refused to apply inline style because it violates the following Content Security Policy directive" (due to 'style-src-elem'). Can you please check if you can remove that inline styling?
In contrast to the other CSP violations in this case the page works fine, so it could be fixable (or maybe ignorable on my side?).
- bobril @Bobris 
- cample @antonmak1 
- openui5 @pskelin @aborjinik 

The following frameworks fail with CSP and currently haven't flag #1139 set with the following error "Refused to execute inline script". 

- dojo
- goui
- owl
- qwik
- quel
- sprae

I'll perform some additional checks and then add the #1139 to package.json for those frameworks.

antonmak1:
@krausest Hello! I will try to fix this issue as soon as possible. Tell me, please, until this is fixed, do I need to manually add such a line to package.json?

`
    "issues": [
      1139
    ]
`
This is important, because it is necessary to clearly determine that there are such errors in the framework (library). I know about CSP, but I didn’t know that it could be tested here. There is no information in the README file about this and about the `npm run checkCSP` function with a [guide](https://github.com/krausest/js-framework-benchmark/wiki/Checking-%231139-flag-(CSP-violation))  that has already been created, as I understand it. Maybe for issue #1139 add a small paragraph to the README? It would be very nice if the authors of new implementations knew about this, so as not to go deep into wikis or issues.

antonmak1:
This week I will then make a PR adding a line about this issue. I would have added earlier, I just didn’t know about it. Sorry. 😕

krausest:
> This week I will then make a PR adding a line about this issue. I would have added earlier, I just didn’t know about it. Sorry. 😕

 No problem, I failed to check the CSP regularly. I added the flag for all frameworks where needed. 
The check is now included in the `npm run rebuild`and `npm run rebuild-ci` or can be invoked with `npm run checkCSP keyed/cample`.

antonmak1:
OK. Then, as I understand it, I will not load a separate pr with the addition of this flag for now. In general, in any case, the problem with CSP is clear. I will then make corrections on this topic to make the framework safer for users.

krausest:
Yes. If you have the corrections ready you can remove the 1139 flag in package.json in the PR.

antonmak1:
Okay

PR #1631, opened 2024-03-16
Update Bobril 20.6.4 without dynamically creating css

Bobris:
Solve #1620 for Bobril.


krausest:

PR #1632, opened 2024-03-16
Update cample to 3.2.0-alpha.29

antonmak1:
If this doesn’t help, I don’t understand yet, but it seems like it should help


krausest:
This time without security updates:

antonmak1:
@krausest Well, this option for select row does not work. I'll try another one.

antonmak1:
All that remains is to fix “select row” and the results will be around 1.04-1.05.

PR #1634, opened 2024-03-21
Update Svelte

trueadm:
Updates Svelte to the latest version


krausest:
Thanks. Here's the comparison: Left new keyed, right old keyed version for comparison:
Looks like svelte got a little bit slower though it's not that easy to decide. Seems like chrome security updates can cause worse results for updated implementations in incremental runs. This time vanillajs and cample weren't impacted so maybe it's really that svelte (and angular) got a little slower.

PR #1636, opened 2024-03-26
Update cample to 3.2.0-alpha.33

antonmak1:
I'll try to do that. I think it will be faster


krausest:
Thanks. Here's the comparison: Left new keyed, right old keyed version for comparison:

antonmak1:
Got it, maybe I'll try with stopPropoganation

PR #1637, opened 2024-03-28
Update Angular

birkskyum:
There is a lot of communication about ongoing work on signals, so it would be interesting to see where things are at.

I updated the dependencies, and applied the migration tool (`ng update @angular/cli --name=use-application-builder`) to move each project to the `application` builder which is the default configuration for [all new projects](https://angular.io/guide/esbuild#for-new-applications).


krausest:
Thanks. I'm afraid it looks like 17.3 is a bit slower than the old version.

PR #1638, opened 2024-03-29
Update dioxus to 0.5

birkskyum:
Dioxus 0.5.0 is 🔥 in terms of improved dx

The previous implementation in this bench of 0.4.0 doesn't build, due to this error:

```sh
error[E0432]: unresolved import `crate::file_engine`
  --> /Users/birkskyum/.cargo/registry/src/index.crates.io-6f17d22bba15001f/dioxus-web-0.4.3/src/lib.rs:59:16
   |
59 | pub use crate::file_engine::WebFileEngineExt;
   |                ^^^^^^^^^^^ could not find `file_engine` in the crate root
```

I did manage to update the implementation to the new 0.5 release, and that builds, so that's a step forward.

I also changed the names to match the other tests data/row/label/id, to cases like `label.label` becomes `row.label`.

@ealmloff, @jkelleyrtp + team - I'm super impressed by this release! It felt great to do a migration virtually only touching the Delete key, and see the code get much cleaner. If you can spot some perf tweaks let me know (e.g. maybe the memos in the RowComponent can be optimized), and I'd be happy to apply them - otherwise I'm excited to see where it goes during the 0.5 cycle. This implementation is on my machine approx on par with the 0.4 implementation as shown below.

Screenshot 2024-03-29 at 01 42 46


krausest:
I can confirm your results. Left new keyed results, right the old results for comparison:

ealmloff:
Thanks for updating the dioxus version! Unfortunately, this implementation doesn't follow the [rules of hooks](https://dioxuslabs.com/learn/0.5/reference/hooks#rules-of-hooks) and will likely leak memory when adding and removing rows. I'll try to open a PR in the next few days to fix the Dioxus implementation

birkskyum:
Thanks @ealmloff ! I had a suspicion it might be the case

birkskyum:
@ealmloff , this might be an easier starting point. https://github.com/krausest/js-framework-benchmark/pull/1646

PR #1639, opened 2024-03-29
Update cample to 3.2.0-alpha.39

antonmak1:
Returned it, everything was as it was. I think select row will return to 3.5 - 3.8. I also updated the removal code, it will work faster there. In main.js, I removed the slice() method from "delete", because for good reason it, of course, should return different arrays, but in this particular case the data does not change quantitatively, so in this particular case the code still works fine without slice(). Somehow, I haven’t noticed this moment in the test for almost a year 😕


antonmak1:
Hooray! I fixed the "select row" and finally found this stupid bug. It should definitely be really fast.

krausest:
Here we go...

antonmak1:
Finally now "select row" works with adequate speed

antonmak1:
"create rows" and "replace rows" seem to have broken down completely, although I didn’t change anything like that

antonmak1:
In short, now I’ll return the code from version 29, that’s all.

PR #1640, opened 2024-03-30
Introduce destam-dom benchmark

Nefsen402:
https://github.com/Nefsen402/destam-dom

This was mostly copy pasted from the solidjs benchmark


krausest:
Welcome destam. On the left is a preview of the results. Please ignore the right part of the table (just relevant for updated implementations). 

Nefsen402:
Thank you!

PR #1641, opened 2024-04-04
Update Svelte

trueadm:
This updates Svelte to the latest version. There were some performance improvements in the latest release.


krausest:
Results are updated. On the left new keyed results, on the right the results from the last run.

PR #1642, opened 2024-04-05
Update pota 

titoBouzout:
Hey, 

I have finally dealt with the "clear rows", also greatly improved other areas. If lucky, it may jump a few dozen places, hopefully forward xD.  I see you have added a package-lock, I haven't touched it, I'm not really sure if it needs an update. 

Thanks!

### DIFF
--- a/frameworks/keyed/pota/package.json
+++ b/frameworks/keyed/pota/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-pota",
-  "version": "1.0.1",
+  "version": "1.0.2",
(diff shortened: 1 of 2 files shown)


titoBouzout:
Thanks Stefan, I expected a bit better results but is not bad at all. 

One thing I can't figure out, from where the number of "run memory 10k" comes from. The memory tab of chrome reports 19MB~ after clicking "create 10k rows" or clicking "append 1k rows" ten times. But on the benchmark it keeps being reported as much higher `30.2`

PR #1643, opened 2024-04-06
Update cample to 3.2.0-alpha.44

antonmak1:
That's it, I understood how the code would work faster and better. I'll combine the code from version 29 and version 39 and it will work fine. In version 29, "select row" was super slower, the rest was normal. Now, "select row" is fixed, so I think it's better this way. I hope the framework has time to get into the results of version 124


krausest:
Here's a comparison (updated keyed results on the left, old results on the right). 1.09 again!

antonmak1:
@krausest OK. I will continue to work

PR #1645, opened 2024-04-08
Doo html dom v0.92.2 beta april 7 2024

hman61:
Minor improvement to updata


krausest:
Here's a comparison (updated keyed results on the left, old results on the right). This time vanillajs was faster by a very small margin.

PR #1647, opened 2024-04-11
Update Svelte

trueadm:
Updates Svelte to the latest version. Svelte 5's latest alpha has a new list reconciliation implementation.


krausest:
Here's a comparison (updated keyed results on the left, old results on the right). Very close, maybe a little bit faster!

PR #1648, opened 2024-04-11
chore: update lwc

nolanlawson:
We have some optimizations in the latest release.

Thank you for continuing to maintain this benchmark! 🙇


krausest:
Here's a comparison (updated keyed results on the left, old results on the right). Looks good!

PR #1654, opened 2024-04-20
Added Eventiveness Web framework

mksunny1:
I have created a powerful new web framework comprising 7 independent libraries and directly benefited from using these benchmarks to test and improve my abstractions. The performance is very good. And thanks for doing these benchmarks.

### DIFF
--- a/frameworks/keyed/deleight/index.html
+++ b/frameworks/keyed/deleight/index.html
@@ -1,0 +1,84 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Benchmarks for Deleight</title>
+        <link href="/css/currentStyle.css" rel="stylesheet" />
+    </head>
+    <body>
+        <div id="main">
+            <div class="container">
+              <div class="jumbotron">
+                <div class="row">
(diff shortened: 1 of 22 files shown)


krausest:
Thanks - I hope to publish results for the PR tomorrow.

krausest:
[Results](https://krausest.github.io/js-framework-benchmark/current.html) are updated. Deleight performs nicely! 

I'll fix the version naming.

mksunny1:
Wow, special thanks! This has been a fantastic learning experience. Your VanillaJS implementation changed some of my assumptions about JavaScript performance. JS is faster than I thought. 

Deleight is not a framework. I created some cost-less (or nearly cost-less) high-level primitives to simplify and accelerate front-end programming. The features must not already exist directly in JavaScript and must provide clear benefits. The performance and how the libraries are used should be up to the programmer. JavaScript is my favourite language and I would like a few more people to see and use it like I do.

mksunny1:
For VanillaJS-3, I am adding async and defer attributes explicitly in the script tag to force asynchronous loading. That improves the time to first paint.

Issue #1661, opened 2024-05-02
Make vanillajs score fastest on chrome 124 (again)


krausest:
With chrome 124 the vanillajs and vanillajs-1 implementation don't score fastest.
Currently it's not clear whether actually some change favors other frameworks or it's a bug with chrome 124 traces.
I'd believe chrome 124 results if we could create a new vanillajs implementation that becomes the fastest implementation again.
The first goal would be an implementation that runs fastest in chrome 124 for create 1,000 rows, replace all rows and create 10,000 rows.
It would be really appreciated to get vanillajs PRs. I promise to run them in a timely fashion. Thanks.

antonmak1:
#1633 Here is written about an assumption to increase the speed of work. I think that after the vanilla-2 js version I will need to close this issue.

krausest:
Thanks. I think I applied the suggestions for vanillajs-2:
*  Spaces in the tr template such that createRow can change nodeValue instead of textContent
* Clear table by cloning the tbody
* (not relevant for the benchmarks below): Use nodeValue instead of data when updating a row
![Screenshot 2024-05-03 at 17 47 47](https://github.com/krausest/js-framework-benchmark/assets/3374055/f60627b0-dfd1-43b3-b6ce-1ac268e04e6e)

Ivi is still faster for create 1k rows, it's a tie for replace rows and vanillajs-2 is faster for create 10k rows.

krausest:
Here's vanillajs with appendInline and the other optimizations applied:
![Screenshot 2024-05-03 at 22 25 03](https://github.com/krausest/js-framework-benchmark/assets/3374055/69e32992-6d4a-4898-8fef-ec93c991c1c9)
Good improvement for create 10k rows. Sadly create 1k rows is stuck and ivi is still ahed.

mksunny1:
Sorry, I have had my hands full the whole time. My implementation in Eventiveness uses the StringBuilder pattern from Java to greatly improve the performance for adding rows, especially for the 10K rows. It's like this:

1. Compose an array containing the outerHTML of all the rows.
2. Join the array at once. This is faster than building up the string incrementally.
3. Create a fragment from the joined string
4. Append the fragment to the tbody.

It is hard to imagine another implementation should be faster than this. You need to realise that faster for 10K rows is always better than for 1k rows. It suggests better scalability.

mksunny1:
[...]

PR #1663, opened 2024-05-02
#1661 Test vanillajs-2

antonmak1:
#1661 I made the code here just as an example, maybe it will help someone (You don’t need to upload it. I’ll delete it later after vanillajs new)

### DIFF
--- a/frameworks/keyed/vanillajs-2/src/Main.js
+++ b/frameworks/keyed/vanillajs-2/src/Main.js
@@ -1,0 +1,205 @@
+'use strict';
+
+function _random(max) {
+    return Math.round(Math.random()*1000)%max;
+}
+
+const rowTemplate = document.createElement("tr");
+rowTemplate.innerHTML = "<td class='col-md-1'></td><td class='col-md-4'><a class='lbl'></a></td><td class='col-md-1'><a class='remove'><span class='remove glyphicon glyphicon-remove' aria-hidden='true'></span></a></td><td class='col-md-6'></td>";
+
+var rowId = 1;
+function buildData(count = 1000) {
+    var adjectives = ["pretty", "large", "big", "small", "tall", "short", "long", "handsome", "plain", "quaint", "clean", "elegant", "easy", "angry", "crazy", "helpful", "mushy", "odd", "unsightly", "adorable", "important", "inexpensive", "cheap", "expensive", "fancy"];
+    var colours = ["red", "yellow", "blue", "green", "pink", "brown", "purple", "brown", "white", "black", "orange"];
+    var nouns = ["table", "chair", "house", "bbq", "desk", "car", "pony", "cookie", "sandwich", "burger", "pizza", "mouse", "keyboard"];
+    var data = [];
+    for (var i = 0; i < count; i++)
+        data.push({id: rowId++, label: adjectives[_random(adjectives.length)] + " " + colours[_random(colours.length)] + " " + nouns[_random(nouns.length)] });
+    return data;
+}
+
+var getParentId = function(elem) {
+    while (elem) {
+        if (elem.tagName==="TR") {
+            return elem.data_id;
+        }
+        elem = elem.parentNode;
+    }
+    return undefined;
+}
+class Main {
+    constructor() {
+        this.data = [];
+        this.selectedRow = undefined;
+
+        document.getElementById("main").addEventListener('click', e => {
+            //console.log("listener",e);
+            if (e.target.matches('#add')) {
+                e.stopPropagation();
+                //console.log("add");
+                this.add();
+            }
+            else if (e.target.matches('#run')) {
+                e.stopPropagation();
+                //console.log("run");
+                this.run();
+            }
+            else if (e.target.matches('#update')) {
+                e.stopPropagation();
+                //console.log("update");
+                this.update();
+            }
+            else if (e.target.matches('#runlots')) {
+                e.stopPropagation();
+                //console.log("runLots");
+                this.runLots();
+            }
+            else if (e.target.matches('#clear')) {
+                e.stopPropagation();
+                //console.log("clear");
+                this.clear();
+            }
+            else if (e.target.matches('#swaprows')) {
+                e.stopPropagation();
+                //console.log("swapRows");
+                this.swapRows();
+            }
+            else if (e.target.matches('.remove')) {
+                let id = getParentId(e.target);
+                let idx = this.data.findIndex(row => row.id === id);
+                //console.log("delete",idx);
+                this.delete(idx);
+            }
+            else if (e.target.matches('.lbl')) {
+                let id = getParentId(e.target);
+                let idx = this.data.findIndex(row => row.id === id);
+                //console.log("select",idx);
+                this.select(idx);
+            }
+        });
+        this.tbody = document.getElementById("tbody");
+    }
+    run() {
+        this.removeAllRows();
+        this.data = [];
+        this.appendRows(buildData(1000));
+        this.unselect();
+    }
+    add() {
+        this.appendRows(buildData(1000));
+    }
+    update() {
+        for (let i=0;i<this.data.length;i+=10) {
+            this.data[i].label += ' !!!'
+            this.tbody.childNodes[i].firstChild.nextSibling.firstChild.firstChild.data = this.data[i].label;
+        }
+    }
+    unselect() {
+        if (this.selectedRow !== undefined) {
+            this.selectedRow.className = "";
+            this.selectedRow = undefined;
+        }
+    }
+    select(idx) {
+        this.unselect();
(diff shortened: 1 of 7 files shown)


frameworks/keyed/vanillajs-2/src/Main.js
```diff
@@ -0,0 +1,205 @@
...
+        // });
+    }
+    swapRows() {
+        if (this.data.length > 998) {
+
+            let tmp = this.data[998];
+            this.data[998] = this.data[1];
+            this.data[1] = tmp;
+
+            let a = this.tbody.lastChild.previousSibling,
```
antonmak1, line 157: Thank you @trueadm

leeoniya: i was about to point this out :)

antonmak1:
I think that `matches` can also be replaced with something faster, but I'm not sure about that.

leeoniya:
> I think that `matches` can also be replaced with something faster, but I'm not sure about that.

sure, `e.target.matches('#clear')` can be `e.target.id === 'clear'`. same for `.className` prop. you might need to do `.className.indexOf()` of `.className.includes()` is there are multiple classes.

antonmak1:
> > I think that `matches` can also be replaced with something faster, but I'm not sure about that.
> 
> sure, `e.target.matches('#clear')` can be `e.target.id === 'clear'`. same for `.className` prop. you might need to do `.className.indexOf()` of `.className.includes()` is there are multiple classes.

Yes, I think it will be faster, but I haven’t tested it, but after `filter` and `forEach` I understand that it is sometimes better to write specific functionality yourself than to use such functionality (in terms of speed)

krausest:
Cool - thanks! Screenshot for the currently relevant benchmarks (was unintentionally chrome 126):
Screenshot 2024-05-02 at 22 08 44
For those benchmarks it's just preventDefault vs stopPropagation, isn't it?

frameworks/keyed/vanillajs-2/src/Main.js
```diff
@@ -0,0 +1,205 @@
...
+        // }
+        // ~251 msecs
+        // for(let i=0;i<this.rows.length;i++) {
+        //     tbody.removeChild(this.rows[i]);
+        // }
+        // ~216 msecs
+        // var cNode = tbody.cloneNode(false);
+        // tbody.parentNode.replaceChild(cNode ,tbody);
+        // ~212 msecs
+        this.tbody.textContent = "";
```
trueadm, line 126: You could also try creating a new `tbody` element and replacing the existing `tbody` with it

frameworks/keyed/vanillajs-2/src/Main.js
```diff
@@ -0,0 +1,205 @@
+'use strict';
+
+function _random(max) {
+    return Math.round(Math.random()*1000)%max;
+}
+
+const rowTemplate = document.createElement("tr");
```
trueadm, line 7: You can try adding a space in the actual template for where the dynamic text nodes will appear. This will create a text node with just a space, which you can then update using `nodeValue` rather than using `textContent`

frameworks/keyed/vanillajs-2/src/Main.js
```diff
@@ -0,0 +1,205 @@
...
+            tbody.appendChild(this.createRow(newData[i]));
+            this.data.push(newData[i]);
+        }
+    }
+    createRow(data) {
+        const tr = rowTemplate.cloneNode(true),
+            td1 = tr.firstChild,
+            a2 = td1.nextSibling.firstChild;
+        tr.data_id = data.id;
+        td1.textContent = data.id;
```
trueadm, line 199: It's much faster to use `firstChild.nodeValue` than using `textContent`. If you apply the change to the template then a text node will exist.

frameworks/keyed/vanillajs-2/src/Main.js
```diff
@@ -0,0 +1,205 @@
...
+        this.appendRows(buildData(1000));
+        this.unselect();
+    }
+    add() {
+        this.appendRows(buildData(1000));
+    }
+    update() {
+        for (let i=0;i<this.data.length;i+=10) {
+            this.data[i].label += ' !!!'
+            this.tbody.childNodes[i].firstChild.nextSibling.firstChild.firstChild.data = this.data[i].label;
```
trueadm, line 94: I've found `nodeValue` faster to use than `data`

trueadm:
@krausest I added comments to this without realising it was merged. Apply those changes and you should see a boost in perf :)

antonmak1:
> Cool - thanks! Screenshot for the currently relevant benchmarks (was unintentionally chrome 126): Screenshot 2024-05-02 at 22 08 44 For those benchmarks it's just preventDefault vs stopPropagation, isn't it?

I just didn’t understand why `preventDefault` should be done for the `button` tag when this is usually done for links. `stopPropagation` just assigns the `cancelBubble` property to stop bubbling.

antonmak1:
@krausest There are still many improvements that can be made, described above by @leeoniya and @trueadm. The community will need to test them and upload them as a pull request.

antonmak1:
@krausest  I don’t know, maybe @hman61 or @ts-thomas will tell about more what is better to do for vanilla-js. @localvoid says that "ivi and vanillajs is that when table is cleared, it removes rows by replacing <tbody> DOM node with a new one instead of removing rows with textContent="""

trueadm:
I'd also argue that https://github.com/krausest/js-framework-benchmark/blob/master/frameworks/keyed/ivi/src/main.ts#L109-L111 falls in a grey area here where no other major framework is doing this. Replacing the `tbody` with an empty one is O(1) in Chrome, whereas `textContent = ''` is O(n) looking at the Chrome code.

antonmak1:
Vanillajs should be an example for other frameworks, which is correct both in terms of speed and logic of operation. Everyone needs to think about this!

antonmak1:
@krausest @trueadm "There are lots of these whacky optimizations though, so it's up to you if the vanilla JS implementation should also do this." - says correctly. You can even put `stopPropagation` on `selectRow` so that it goes faster without bubbling to `tr` and it will also speed up, but this is not correct, as I understand it.

antonmak1:
I still don’t understand whether it’s normal to put `addEventListener` on `#main`? Shouldn't it be on `window` or at most on `document`?

frameworks/keyed/vanillajs-2/src/Main.js
antonmak1: "Replacing the tbody with an empty one is O(1) in Chrome, whereas textContent = '' is O(n) looking at the Chrome code" - so, it turns out it's better not to do it?

trueadm: Yeah you want to replace the node instead of clearing text content

antonmak1: Well, yes. If you also take into account, for example, that logically there can be a bunch of event listeners on tbody, then it’s better not to do this as an example to other people. So, in principle, in this particular speed task, this is not critical. But, it’s really better to leave textContent then.

trueadm: The event listeners can be on the table.

antonmak1: @syduki cloneNode() does not copy event listeners. At the very least, this solution cannot be used for vanilla-js. But, in general, in a specific task of cleaning a table, it is not some kind of direct cheating if you change the tbody.

trueadm: It's arguable a cheat too, as if this was a stateful DOM element like a custom element, then replacing it with an empty clone is also not acceptable and in the real world would cause major problems. I'm not sure if we should push ivi to change its implementation here (or add a note) or instead allow this. That's up to @krausest to decide.

trueadm:
@antonmak1 It doesn't make any real difference in reality if the event listener is on the document or `#main`.

antonmak1:
@trueadm If we take the bubbling process, then this will be several more `parentNode`'s (if without `stopPropagation`). What I mean is that in most popular frameworks an event listener is attached to `document` or `window`. Maybe this practice could be used as an example for future frameworks? But, yes, you are right. Specifically, in this task it doesn’t really matter where the event listener will be added.

krausest:
Please see #1661 for updates.

antonmak1:
> > I think that `matches` can also be replaced with something faster, but I'm not sure about that.
> 
> sure, `e.target.matches('#clear')` can be `e.target.id === 'clear'`. same for `.className` prop. you might need to do `.className.indexOf()` of `.className.includes()` is there are multiple classes.

@krausest instead of matches you can try this too

antonmak1:
@krausest no need to change tbody to a new one - this is garbage!

PR #1666, opened 2024-05-07
Add `vanillajs-lite` implementation (#1661 fix)

syduki:
This is intended as a fix for issue #1661 (you may need to test it first in your environment before making a decision).
Due to convoluted [vanillajs](https://github.com/krausest/js-framework-benchmark/blob/master/frameworks/keyed/vanillajs/src/Main.js) implementation which makes the optimization difficult, I did a complete rewrite with minimal abstractions and cleaner execution flow, for a better visibility of current and future optimization paths.

It shows promising results in my environment (Ubuntu 23.10, Google Chrome for Testing
 Version 124.0.6367.91 (Official Build) (64-bit)):

[results.json](https://github.com/krausest/js-framework-benchmark/files/15237974/results.json)

image
image

### DIFF
--- a/frameworks/keyed/vanillajs-lite/index.html
+++ b/frameworks/keyed/vanillajs-lite/index.html
@@ -44,7 +44,7 @@
                 <span class="preloadicon glyphicon glyphicon-remove" aria-hidden="true"></span>
             </div>
         </div>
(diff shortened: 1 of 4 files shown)


krausest:
Thanks. The isKeyed test can't handle this implementation yet. But it looks to be keyed.
Looks like your implementation removes the tbody. If we solve #1661 then I'll fix isKeyed to also count the child nodes. I think isKeyed might work correctly then.
I'll run it anyway and report the results in the other branch.

PR #1674, opened 2024-05-11
Use last-resort gc, fixes #1667

### LINKED ISSUE #1667: Modify GC for chrome 125
Use a single call to `window.gc({type:'major',execution:'sync',flavor:'last-resort'})` as described in #1661 .

robbiespeed:
Use last-resort gc, fixes #1667

Relevant discussion here as to why this change is being made https://github.com/krausest/js-framework-benchmark/issues/1661#issuecomment-2099578793. TLDR: regular GC passes no longer seem enough to collect all orphaned dom nodes from warm-up runs.


robbiespeed:
Oops this fails compile since a new param was added to forceGC, will fix.

PR #1675, opened 2024-05-11
Non-keyed implementation

mksunny1:
- added non-keyed implementation for Vanillajs-3
- small code enhancements
- semantic bug fix (in remove) for the keyed implementation.


krausest:
I plan to publish results tomorrow.

krausest:
Left new keyed results, right old keyed results. Nice!
Screenshot 2024-05-16 at 19 51 44

mksunny1:
So sorry I introduced a bigger semantic bug when fixing the first one. I have now corrected it. Additionally, I updated the `append` function to clone the item template retrieved from DOM before using it. Without cloning, it will grow to a count of 100 when adding 10K rows hogging memory and affecting overall performance (I think)...

PR #1676, opened 2024-05-13
`vanillajs-lite` update

syduki:
This update fixes memory issue for "creating/clearing 1k rows" test which seems to be caused by hybrid rows clearing feature (tbody.remove()/tbody.textContent = '') issued in the most recent update. Also, some refactoring to reduce the file size.
The changes seems to not have much impact on CPU tests results:

image
image


krausest:
I plan to publish results tomorrow.

krausest:
Left new keyed results, right old keyed results. Can't say yet if it's just a bad run or impact of OSX 14.5 (please note: doohtml was updated)
Screenshot 2024-05-16 at 19 51 44

I'll run this new version and the old version such that we can compare on my machine. After that I'll look at chrome 125.

PR #1678, opened 2024-05-15
refactor solid & solid-store implementations

qeleb:
### Changes
1. Clean up `keyed/solid` & `keyed/solid-store` implementations by running Prettier & adopting patterns from other framework setups for clearer code & better uniformity
1. Update `keyed/solid` & `keyed/solid-store` rollup configs with a minimal terser setup to more reasonably represent what the average user will experience when using a boilerplate or popular framework like `SolidStart`.

   > Is this over-optimization? _No._ I believe this is firmly in the spirit of representing what an average user will realistically experience with Solid. Plus, this is considerably less bundler customization than what is used in many of the other framework benchmarks, like [million](https://github.com/krausest/js-framework-benchmark/blob/9fcbe016d312baaf4c2a14dff3d5f2fc2ffbe290/frameworks/keyed/million/rollup.config.mjs#L5), [gxt](https://github.com/krausest/js-framework-benchmark/blob/30c247a897cfe147a94ff82edafd8d066f4ffb36/frameworks/keyed/gxt/vite.config.mts#L29), [cydon](https://github.com/krausest/js-framework-benchmark/blob/30c247a897cfe147a94ff82edafd8d066f4ffb36/frameworks/non-keyed/cydon/rollup.config.js#L18), etc.

### DIFF
--- a/frameworks/keyed/solid-store/rollup.config.js
+++ b/frameworks/keyed/solid-store/rollup.config.js
@@ -1,25 +1,23 @@
 import resolve from "@rollup/plugin-node-resolve";
-import babel from "@rollup/plugin-babel";
-import  terser  from "@rollup/plugin-terser";
+import { babel } from "@rollup/plugin-babel";
+import terser from "@rollup/plugin-terser";
 
-const plugins = [
-  babel({
-    babelHelpers: "bundled",
-    exclude: "node_modules/**",
-    presets: ["solid"],
-  }),
-  resolve({ extensions: [".js", ".jsx"] }),
-];
-
-if (process.env.production) {
-  plugins.push(terser());
-}
+const TERSER_OPTIONS = {
+  module: true,
+  compress: { passes: 3 },
+  mangle: true,
+};
 
 export default {
   input: "src/main.jsx",
-  output: {
-    file: "dist/main.js",
-    format: "iife",
-  },
-  plugins,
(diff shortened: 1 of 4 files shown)


krausest:
@trueadm Are you fine with those changes?

trueadm:
Did you mean @ryansolid?

krausest:
Sorry, you're right!

krausest:
Here's a comparison (left new keyed, right old keyed version):
Screenshot 2024-05-24 at 18 08 32

qeleb:
@krausest thank you! i’m surprised it went up for one and down for the other.. the changes to each are the same & I expected minification to only improve things.

The bundle size is about 4.5% less so there’s probably a win in that section at least

ryansolid:
@qeleb One thing I noticed is you removed event delegation. Which yes reduced bundle size but probably made performance worse.

qeleb:
@ryansolid Did i? I thought `onClick` was delegated compared to on:click?

The only place where I knowingly reduced performance was in removing the event binding in some places, adding a few extra () => invocations. 

In my testing this didn’t really hurt scores measurably so, but I agree this part was better before. It’s mostly just because it’s extremely rare I see people bind events this way so I thought this particular feature being unused might be more representative, but i’m positive you would know better than I would.

PR #1679, opened 2024-05-17
Fixed bug in remove (again!)

mksunny1:
Er sorry I introduced a bigger semantic bug in remove when fixing the first o

Screenshot 2024-05-17 at 05 47 03


mksunny1:
It should be `splice(index, count)` but it was just `splice(index)`

krausest:
Sorry, the non-keyed version is broken (due to _random). Would you take a look at that please?
Though I'm not sure having faster non-keyed vanillajs implementations is a rewarding goal. Maybe we just should delete all vanillajs implementations for non-keyed.

mksunny1:
I have fixed non-keyed now and also worked further on keyed. I don't think it is necessary to delete non-keyed JS implementations. Especially for `replace` and `swap` I am more comfortable with that implementation. It is wasteful to rebuild just for 1 or 2 property changes.

For the keyed implementation I noticed that the previous `update` implementation may have been wasteful. Also switching to regular scripts now brings all the globals into the large `window` object. So I decided to use a class now.

mksunny1:
How it's looking from here:

Screenshot 2024-05-19 at 02 32 32

Screenshot 2024-05-19 at 02 32 43

Screenshot 2024-05-19 at 02 32 54



Vanillajs-4 shown is the Vanillajs-3 implementation in the Chrome 125 results table.

krausest:
Thanks - here's a comparison (left new keyed, right old keyed version):
Screenshot 2024-05-24 at 18 08 32

P.S: Please don't commit the result files when updating a lib. There's no reasonable way to merge them.

mksunny1:
Thanks. It seems my system struggles to produce consistent results complicating my efforts to optimise. And obviously I don't understand the browser internals enough. I will look again, reevaluate my assumptions, use more tools, etc. Can you suggest how to get more reliable testing results? 

Also I have noted about the extraneous commit.

mksunny1:
I think I should use a closure instead of a class or global. This is showing the best performance from looking at your results. For `select` and `remove`, the new implementation uses closures to hold some of the 'outside' variables instead of globals or class members. Partial update was  understandably better as I optimised that. `Swap` was almost the same as I hardly changed anything there. 

I really thought `runlots` would do better in the new implementation and they were almost the same. I am not surprised about the poorer score for `run`. But the new implementation feels much better when testing manually. With the old one, after creating 10K rows and clearing, creating 1k rows subsequently feel just as slow as creating 10k rows again. It performs expectedly better in the new one. So I like how the new version creates rows better, but it ends up scoring less. I feel this may be down to how the measurement is done.

krausest:
> It seems my system struggles to produce consistent results complicating my efforts to optimise.

I've put a discussion here. It's currently a draft since I didn't have the time to check that all commands are actually correct:
https://github.com/krausest/js-framework-benchmark/discussions/1687
Feel free to submit feedback. Especially if increasing the count helps to produce statistically significant differences.

PR #1680, opened 2024-05-17
Update README.md

antonmak1:
This is no longer necessary, as I understand it.


krausest:
Yes, I'm pretty happy you're right 😄

Issue #1531, opened 2023-12-13
sprae should be flagged #1139


0-v-0:
It fails with CSP.
See https://github.com/dy/sprae/blob/5a7cfc436d0140b6d15194a40675b46e08342f75/src/directives.js#L415

dy:
As well as alpinejs https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/evaluator.js#L81.
I wonder though which flags you mean

dy:
Ok, sprae@10 supports safe-eval via justin.

krausest:
sprae now has #1139

PR #1684, opened 2024-05-20
Sprae v10

dy:
Update [sprae](https://github.com/dy/sprae) version to v10.
It is more compact & performant than ever before.
Also it enables sprae-safe version with preact-signals and safe eval (CSP-friendly #1620, #1531)

@krausest


krausest:
Building fails, since ./node_modules/@preact/signals-core/dist/signals-core.min.mjs doesn't exist after npm install.
There's a signals-core.mjs or a signals-core.min.js, but no signals-core.min.mjs. Can you take a look at it please?

dy:
Ok' I've removed sprae-safe since results are not much different, except for package size.

PR #1686, opened 2024-05-25
Updated keyed implementation to use a closure

mksunny1:
As a follow up to your feedback yesterday on the relative performances of previous implementations I decided to use a closure in preference to globals or a class. The `create` functions should do better also as I moved the global arrays which were only aliased earlier into the function instead.


krausest:
Thanks - I hard a hard time getting good results for both vanillajs and vanillajs-3, seems like a reboot was needed to get those results:
Screenshot 2024-05-26 at 19 22 25

mksunny1:
After posting my response the other day, I was disappointed when I increased the count to 50 and later 100. Therefore I resolved to combine the good things from both implementations. That increased the size quite a bit. 

It seems these results are showing poorer performance than the former Implementation. I believe they have the same performance when you increase the count (I checked) at least for 1K rows. I had more consistency between benchmark count of 50 and 100.

PR #1690, opened 2024-05-27
destam-dom: Update library

Nefsen402:
+ some optimizations and simplifications

This removes the last low hanging fruit in terms of perf improvements. I don't think I can get much better than this.


krausest:
Thanks. Here's a comparison (left keyed new, right keyed old results):
Screenshot 2024-06-04 at 21 57 50

PR #1691, opened 2024-05-27
`vanillajs-lite` update

syduki:
A little bit of refactoring, including micro-optimization of update/select/remove.


krausest:
Thanks. Here's a comparison (left keyed new, right keyed old results):
Screenshot 2024-06-04 at 21 57 50

PR #1695, opened 2024-06-05
update angular

krausest:
This is the PR for angular 18 that didn't perform as well as expected.

### DIFF
--- a/frameworks/keyed/angular-cf-nozone/src/main.ts
+++ b/frameworks/keyed/angular-cf-nozone/src/main.ts
@@ -1,10 +1,9 @@
 import { bootstrapApplication } from '@angular/platform-browser';
 import { AppComponent } from './app/app.component';
-import { NgZone, ɵNoopNgZone } from '@angular/core';
+import { provideExperimentalZonelessChangeDetection } from '@angular/core';
 
 bootstrapApplication(AppComponent, {
   providers: [
-    // https://github.com/angular/angular/issues/47538
-    { provide: NgZone, useClass: ɵNoopNgZone }
+    provideExperimentalZonelessChangeDetection()
   ]
 });
--- a/frameworks/keyed/angular-cf-signals/src/main.ts
+++ b/frameworks/keyed/angular-cf-signals/src/main.ts
@@ -1,4 +1,9 @@
 import { bootstrapApplication } from '@angular/platform-browser';
 import { AppComponent } from './app/app.component';
+import { provideExperimentalZonelessChangeDetection } from '@angular/core';
 
-bootstrapApplication(AppComponent);
+bootstrapApplication(AppComponent, {
+    providers: [
+      provideExperimentalZonelessChangeDetection(),
+    ],
+ });
--- a/frameworks/keyed/angular-cf-signals/src/main.ts
+++ b/frameworks/keyed/angular-cf-signals/src/main.ts
@@ -1,9 +1,4 @@
 import { bootstrapApplication } from '@angular/platform-browser';
 import { AppComponent } from './app/app.component';
-import { provideExperimentalZonelessChangeDetection } from '@angular/core';
 
-bootstrapApplication(AppComponent, {
-    providers: [
-      provideExperimentalZonelessChangeDetection(),
-    ],
- });
+bootstrapApplication(AppComponent);
--- a/frameworks/keyed/angular-cf-new-nozone/.browserslistrc
+++ b/frameworks/keyed/angular-cf-new-nozone/.browserslistrc
@@ -1,0 +1,8 @@
+# This file is used by the build system to adjust CSS and JS output to support the specified browsers below.
+# For additional information regarding the format and rule options, please see:
+# https://github.com/browserslist/browserslist#queries
+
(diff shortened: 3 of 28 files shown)


frameworks/keyed/angular-cf-nozone/src/main.ts
```diff
@@ -1,10 +1,9 @@
 import { bootstrapApplication } from '@angular/platform-browser';
 import { AppComponent } from './app/app.component';
-import { NgZone, ɵNoopNgZone } from '@angular/core';
+import { provideExperimentalZonelessChangeDetection } from '@angular/core';
 
 bootstrapApplication(AppComponent, {
   providers: [
-    // https://github.com/angular/angular/issues/47538
-    { provide: NgZone, useClass: ɵNoopNgZone }
+    provideExperimentalZonelessChangeDetection()
```
pkozlowski-opensource, line 7: I would suggest reverting for now based on the discussion in https://github.com/pkozlowski-opensource/js-framework-benchmark/commit/307580b3c4bcd1cd83aef10bef3f075f3cd7f20f

frameworks/keyed/angular-cf-signals/src/main.ts
```diff
@@ -1,4 +1,9 @@
 import { bootstrapApplication } from '@angular/platform-browser';
 import { AppComponent } from './app/app.component';
+import { provideExperimentalZonelessChangeDetection } from '@angular/core';
 
-bootstrapApplication(AppComponent);
+bootstrapApplication(AppComponent, {
+    providers: [
+      provideExperimentalZonelessChangeDetection(),
```
pkozlowski-opensource, line 7: I would suggest reverting for now based on the discussion in https://github.com/pkozlowski-opensource/js-framework-benchmark/commit/307580b3c4bcd1cd83aef10bef3f075f3cd7f20f

krausest: Ok. But then it seems like it have to add zone.js to angular.json again.

pkozlowski-opensource: Yes, in this benchmark, yes.

krausest:
I tried to merge your suggestions.
It seems to me that the double CD happens only when provideExperimentalZonelessChangeDetection() is used. The implementation from your gist didn't work with { provide: NgZone, useClass: ɵNoopNgZone }.
Thus I created two implementations: angular-cf-nozone is the old { provide: NgZone, useClass: ɵNoopNgZone } implementation with detectChanges calls and angular-cf-new-nozone with your gist and provideExperimentalZonelessChangeDetection().
All implementations are updated to angular 18 and use @angular/build

I'll run and report results soon...

pkozlowski-opensource:
@krausest thnx for doing all this!

> It seems to me that the double CD happens only when provideExperimentalZonelessChangeDetection() is used.

Correct.

> The implementation from your gist didn't work with { provide: NgZone, useClass: ɵNoopNgZone }.

Right, it should not. I'm sorry if I've confused you with my patch commit that was only meant to be used with provideExperimentalZonelessChangeDetection. But glad it was resolved.

>  Thus I created two implementations: angular-cf-nozone is the old { provide: NgZone, useClass: ɵNoopNgZone } implementation with detectChanges calls and angular-cf-new-nozone with your gist and provideExperimentalZonelessChangeDetection().

This make sense.

> I'll run the benchmark and report results soon...

Awesome. I will poke around angular-cf-new-nozone as I expect that we are going to see "bad" results again (at least this is what I saw locally today). I do suspect that this is linked to the setup as numbers reported by the benchmark are _widely_ different from what I see under a profiler. But yeh, let's wait and see.

Appreciate your work on this one!

krausest:
Thanks! Here's a comparison (left new keyed results, right old keyed)
Screenshot 2024-06-06 at 22 03 40
I'd say no (or almost no) change for cf-nozone and ngfor. cf and cf-signals are 0.5 slower (0.5 is usually beyond noise). Avoid double change detection helps at least a bit for cf-new-nozone.

If you don't mind I'd merge the PR to get it into chrome 126 results. Should we add  cf-new-nozone for chrome 126 or postpone that?

pkozlowski-opensource:
> I'd say no (or almost no) change for cf-nozone and ngfor. cf and cf-signals are 0.5 slower (0.5 is usually beyond noise). Avoid double change detection helps at least a bit for cf-new-nozone.

Agree with the conclusions, need to look into / profile those scenarios.

> If you don't mind I'd merge the PR to get it into chrome 126 results.

Go for it, we shouldn't block on my investigations.

>  Should we add cf-new-nozone for chrome 126 or postpone that?

Postpone. There is clearly more work that we need to do there and the new zoneless is highly experimental. Happy to work with you / anyone else on adding those in the (near) future, though.

krausest:
I moved cf-new-nozone to #1696 and merged the rest.

pkozlowski-opensource:
> I moved cf-new-nozone to https://github.com/krausest/js-framework-benchmark/pull/1696 and merged the rest.

This is perfect. I will dig into this one.

PR #1699, opened 2024-06-16
Update quel to 0.13.1, add keyed/quel

mogera551:
This PR update quel to 0.13.1, add keyed/quel


krausest:

krausest:
Here's a screenshot of the updated [results](https://krausest.github.io/js-framework-benchmark/current.html) (left new version, right old version if existing). Vanillajs was about 0.3 slower than in the full run, so you might expect a bit better performance for chrome 127 full run.
Screenshot 2024-06-25 at 20 31 59

PR #1702, opened 2024-06-20
Update cample to 3.2.0-alpha.49

antonmak1:
I decided to look at the benchmark and saw that it was possible to redo the selection of rows. For some reason, I haven’t thought of such an implementation in a year or so :confused:

### DIFF
--- a/frameworks/keyed/cample/package-lock.json
+++ b/frameworks/keyed/cample/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-cample",
-  "version": "3.2.0-alpha.45",
+  "version": "3.2.0-alpha.48",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


antonmak1:
Wow, how stupid I am. I could not use the data that comes from the component, because in this case it is not necessary to store it. the line does not directly depend on them (but the label does). My code ended up going through the components 1000 times, and I kept wondering what was going on.

pawelblaszczyk5:
I feel like the optimization added with this update fails under [this](https://github.com/krausest/js-framework-benchmark/issues/772) 😃

antonmak1:
@pawelblaszczyk5 Yes, you may be right. I understood where the error in the slow operation of cample’s select row is in the code, but it’s better to fix it at the framework level.

antonmak1:
@pawelblaszczyk5 Thank you! Yes, it looks like I was in a hurry and did it wrong.

PR #1712, opened 2024-07-02
Update cample to v3.2.1-beta.0

antonmak1:
I solved the problem with the click without resorting to the methods described [here](https://github.com/krausest/js-framework-benchmark/issues/772). The main idea is a stack of real UI, where I store state that is not in contact with the main data. That is, this can be compared to how if we entered some data into the "input", and so this data is, as it were, reflected in this idea.

I don't know, maybe this is a trivial idea, but it sounds interesting to create a separate stack for each DOM element.

### DIFF
--- a/frameworks/keyed/cample/package-lock.json
+++ b/frameworks/keyed/cample/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-cample",
-  "version": "3.2.0-alpha.45",
+  "version": "3.2.1-beta.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "js-framework-benchmark-cample",
-      "version": "3.2.0-alpha.45",
+      "version": "3.2.1-beta.0",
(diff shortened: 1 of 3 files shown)


krausest:
The [results](https://krausest.github.io/js-framework-benchmark/current.html) are updated (left updated keyed results, right old results). I guess we must take a look at the big picture in the next full run.
Screenshot 2024-07-05 at 07 37 07

antonmak1:
@krausest Since the last release, I haven’t changed almost anything in other benchmarks, although the speed has dropped significantly. I don't know what the reason is. Well, this is a one-off launch, so that's ok. The main thing is that now select row shows quick results!

antonmak1:
Everything is green - finally. Then only apply graphomaniac changes like changing classList to className, textContent to nodeValue, etc.

antonmak1:
Anyway, thank you very much! I'm just glad that select row was fixed properly!

antonmak1:
Everything is green. How beautiful!

Issue #885, opened 2021-05-07
Running tests on an M1 Mac/ARM builds of Chrome


brainkim:
Has anyone tried running the benchmarks on an M1 Mac yet? What I’m seeing is that 16x slowdowns are disproportionately slow compared to intel, such that it almost seems like the tests are just hanging, even on VanillaJS the first run of benchmarks 03_ and 04_ seem to take forever.

MacBook Pro (13-inch, M1, 2020)
macOS Big Sur Version 11.2.2
Google Chrome Version 90.0.4430.93 (Official Build) (arm64)

ryansolid:
Yeah I have one as well. MacBook Air M1. Exact same problem. All the slowdowns are bad. They lead to a performance drop off progression. You can see even with 2x or 4x slowdown that each subsequent run gets slower. Just at 16x it is insane, like 10seconds + per test run. I was hoping this was just a chrome driver thing that would go away, but had this problem in 89 and in 90.

So yes very much real.

krausest:
I also couldn't resist and bought a MacBook Air M1. Just as a rant: It's really a lovely machine, though I started to realize how much I should have loved my linux based Razer Blade 15 after I found that the MacBook isn't perfect neither (even its great touchpad). All the possibilities in Linux make me aim for perfection but only leave me disappointed. But I really enjoy working with the Razer much more since I started to use both 😄 

I saw the same effect. The CPU slow down isn't comparable and much slower than on linux. But I even saw that when switching between linux and windows on the Razer Blade 15. I'm a bit afraid of CPU throttling due to the fact that the MacBook has no fan (for anything but benchmarks this is of course a great feature). 
@ryansolid Did you see CPU throttling effects for the benchmark?

I'm a bit hesitating to publish results for the MacBook since results won't be comparable. A possibility would be to temporarily remove all CPU slowdowns to allow a better comparison and add other slowdowns when we decide that we migrate back to a Mac.

krausest:
I put a preview on https://krausest.github.io/js-framework-benchmark/2021/table_chrome_90_osx.html and https://krausest.github.io/js-framework-benchmark/2021/table_chrome_90_linux.html

ryansolid:
I tend to not run the whole test suite as Im always focused on specific implementations. I have not seen throttling but I only run maybe 3-4 frameworks at a time.

[...]

PR #1714, opened 2024-07-04
Update svelte to 5.0.0-next.174

trueadm:
There were some bugs in Svelte 165 that affected performance, they should be fixed in this update.

### DIFF
--- a/frameworks/keyed/svelte-classic/package-lock.json
+++ b/frameworks/keyed/svelte-classic/package-lock.json
@@ -13,7 +13,7 @@
         "@rollup/plugin-terser": "^0.4.4",
         "rollup": "^4.14.3",
         "rollup-plugin-svelte": "^7.2.0",
-        "svelte": "^5.0.0-next.165"
+        "svelte": "^5.0.0-next.174"
       }
     },
     "node_modules/@ampproject/remapping": {
@@ -759,9 +759,9 @@
       }
     },
     "node_modules/svelte": {
-      "version": "5.0.0-next.165",
-      "resolved": "https://registry.npmjs.org/svelte/-/svelte-5.0.0-next.165.tgz",
-      "integrity": "sha512-r/dr89AAzIsT/tpIJsenhkWjJ7j2Q1Q3tdPu2qgi/8Vc42dtvHPr1XHn9qJhjXIGVE/cS1BxEuwG+Qwp7YUqkw==",
+      "version": "5.0.0-next.174",
+      "resolved": "https://registry.npmjs.org/svelte/-/svelte-5.0.0-next.174.tgz",
+      "integrity": "sha512-TUjE3ba4wyiLa67/07aIDwEy7rAY12SR+v2FBMKBXrqN92pUI8bDYIJA0CER74lSE4WwVcqapFxe7JIGq/swPA==",
       "dev": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.1",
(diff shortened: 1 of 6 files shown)


krausest:
Thanks - I'll post the results tomorrow.

krausest:
The [results](https://krausest.github.io/js-framework-benchmark/current.html) are updated (left updated keyed results, right old results). svelte-classic is a bit slower, but it's within the range of noise.
Screenshot 2024-07-05 at 07 37 07

trueadm:
@krausest That's strange. I get very different results locally – with 174 being significantly faster than before.

krausest:
I took another look at it just to check if I made some mistake or my update scripts wouldn't work right. So I created two new versions with just the package.json adjusted to .165 and .174, built the two and ran them together with the unchanged keyed/svelte version (i.e. also .174). 
I confirmed that the shasum of main.js in keyed/svelte and keyed/svelte-174 was identical and differed from keyed/svelte-165, so I'm quite confident that I didn't mix up versions. Here's the result on my machine (performance looks completely identical for the three versions):
Screenshot 2024-07-10 at 18 33 43 

The package-lock of .165 looks like that:
```
"node_modules/svelte": {
      "version": "5.0.0-next.165",
      "resolved": "https://registry.npmjs.org/svelte/-/svelte-5.0.0-next.165.tgz",
      "integrity": "sha512-r/dr89AAzIsT/tpIJsenhkWjJ7j2Q1Q3tdPu2qgi/8Vc42dtvHPr1XHn9qJhjXIGVE/cS1BxEuwG+Qwp7YUqkw==",
```

Can you post a screenshot of the results on your machine? What are the specs of your machine?

trueadm:
I notice that in my runs I get far more variance on all the test cases (select row is +- 1.1). I'm running an M2 Max with 32GB RAM + Chrome 126. So that is probably why the numbers might be off for me. The variance never used to be that bad in earlier versions of Chrome either :/

krausest:
Just to make sure: Is the testdriver up to date and was it built recently (i.e. after chrome 124)? 
(if not => `npm run compile` in webdriver-ts)

trueadm:
@krausest I'll do that and re-run the benchmarks for both and also the other frameworks.

trueadm:
Screenshot 2024-07-10 at 18 42 26

You were right, after updating webdriver-ts, there's almost nothing in it. Interestingly though the overalls core is considerably different from yours and also in comparison to the other frameworks. I assume the scores will normalise out when you do the next browser update?

krausest:
That's good!
This is something I have no good explanation. Incremental runs (just the updated frameworks + vanillajs) tend to perform worse than full runs. I eliminated minor browser updates by switching to a fixed chrome for testing version, I'm stopping onedrive and I'm quitting all apps whenever I'm running the benchmark.
For a full run I'm also rebooting before running and I guess there are less background processes (like karabiner elements, betterdisplay, etc. won't be started), but I'm not sure what is guilty here. Seems to have become worse since a few months.
The last [full run ](https://krausest.github.io/js-framework-benchmark/2024/table_chrome_126.0.6478.55.html ) was closer to your results. So I indeed hope that results look better in a full run.

P.S. I'm surprised the absolute numbers from the full run for my M1 are so close to your M2.

krausest:
If you want to read more about the chrome 124 issue, this is a good start: https://github.com/krausest/js-framework-benchmark/issues/1661#issuecomment-2099578793

trueadm:
@krausest M1 and M2s are sensitive to temperature. I have a window near my Mac which helps keep it cool. If I close my window and its warm outside, the results jump all over the place pretty quickly (you can use a core speed tracker).

krausest:
I'll have to keep an eye on that. When I had a MacBook Air M1 I looked at powermetrics: https://github.com/krausest/js-framework-benchmark/issues/885#issuecomment-841810772 but didn't find signs of throttling (at least that way).
Maybe I'll take a look at Mx Power Gadget. What do you use to find out about throttling?

PR #1716, opened 2024-07-12
Add Xilem Web 0.1.0

flosse:


flosse:
@Philipp-M even though xilem-web is still in its infancy, I think it's important to share the potential publicly at an early stage.
In addition, the double implementation might not have happened if I had already found your code here :wink:
And feel free to take a critical look at it here :relaxed:

Philipp-M:
I'm a little bit short on time currently, to review/investigate this more closely.
Though I want to note, that my implementation results in 20kb less wasm bin size (or roughly 19% less).

I suspect it has to do with the random number generation, I'd just use `js_sys::Math::random()`. I haven't tested yet how it compares otherwise, I guess it behaves similar.

One reason why I haven't added this bench yet here (apart from the keyed impl), is because I wanted to add hydration + templates before in xilem_web, which seems to have a significant effect on the "create" part of this bench. I try to get to that in the near future.

krausest:
Should we merge it or consider it as a draft?

Philipp-M:
Hmm if @flosse doesn't have anything against it, I'd like to wait, (i.e. draft) (and possibly at least optimize the bundle size as much as possible), until it (xilem_web) is slightly more polished to be able to more publicly announce it. It's now here as PR, so I hope that avoids further duplicate work.

flosse:
> Hmm if @flosse doesn't have anything against it, I'd like to wait, (i.e. draft)

sure!

> (and possibly at least optimize the bundle size as much as possible), until it (xilem_web) is slightly more polished to be able to more publicly announce it. It's now here as PR, so I hope that avoids further duplicate work.

:+1:

krausest:
I'm closing old draft PRs. Please re-submit when ready.

PR #1717, opened 2024-07-24
Update quel to 0.16.1

mogera551:
This PR updates quel to 0.16.1 and adds support for single file component like a .vue and .svelte


krausest:
Thanks - here's a screenshot of the results (left updated version, right old version):

PR #1718, opened 2024-07-25
Fix dioxus benchmark implementation

ealmloff:
This PR applies a few optimizations to the dioxus implementation continuing from #1646:
- Each label in rows are a signal
- Re-use memory for the rows instead of reallocating every time
- Rows are a child component that reruns separately from the parent
- Switch the selected row to use the `use_set_compare` hook which is similar to a selector in solidJS
- Use js-sys for random number generation for better performance and smaller bundle sizes

Running the benchmarks locally, performance seems a lot closer to dioxus 0.4 with less memory usage:

Screenshot 2024-04-09 at 12 29 14 PM
Screenshot 2024-04-09 at 12 30 05 PM


krausest:
Here's a screenshot of the results (left updated version, right old version):

PR #1719, opened 2024-07-26
use preventDefault() to prevent scroll to top due to href="#"

sudo-barun:


krausest:
Thanks - here's a screenshot of the results (left updated version, right old version):

PR #1723, opened 2024-08-06
Add Hono client-side benchmarks

aminya:
This adds Hono client-side benchmarks. The code was adapted from the React-hooks implementation

### DIFF
--- a/frameworks/keyed/hono/.gitignore
+++ b/frameworks/keyed/hono/.gitignore
@@ -1,0 +1,28 @@
+# dev
+.yarn/
+!.yarn/releases
+.vscode/*
+!.vscode/launch.json
+!.vscode/*.code-snippets
+.idea/workspace.xml
+.idea/usage.statistics.xml
+.idea/shelf
+
+# deps
+node_modules/
+
+# env
+.env
+.env.production
+
+# logs
+logs/
+*.log
+npm-debug.log*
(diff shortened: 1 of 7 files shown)


aminya:
The initial results comparing Hono to Solid. Hono is much slower than Solidjs.

Note that I didn't sit through the 10K row creation for Hono as it was too slow. I waited 20 minutes, but the benchmarks didn't finish.

image

ryansolid:
As far as I can tell `findInsertBefore` is called an enormous amount of time recursively and this makes everything slower and causes a stack overflow in the 10k row case. Scheduling everything through individual requestAnimation frames is also costly on performance.

aminya:
The profiler confirms the inefficient implementation. 

image

ryansolid:
You're probably best served to open an issue on their repo and link this. I don't think it is worth adding to the benchmark until they fix this.

aminya:
I will open an issue for Hono, but, I think it is crucial to have a reference benchmark for the users when they are making a choice about a certain technology. Hono's performance issues were probably unknown until I added this benchmark.

aminya:
Updated to the latest version of Hono, which fixes some of the performance issues.

krausest:
Thank. Hope to post results soon!

krausest:
Way better than expected!

Issue #1721, opened 2024-07-28
Incremental runs are worse than full runs


krausest:
There are two kind of runs: 
* Full runs. For a new chrome version I'm running all frameworks on my machine. The first night only the keyed implementations and when that looks good the second night the non-keyed implementations. Before running the keyed implementations I'm rebooting and stopping all the helpers (BetterDisplay, OneDrive, Karabiner Elements), stopping the search index (sudo mdutil -i off /), and making sure that there are no active apps running on the machine (other browser instances, VSCode etc.)
* Incremental runs. For updated implementations I'm running vanillajs and the updated implementations. I'm stopping all active applications, but neither rebooting nor stopping the helper apps. vanillajs serves as a reference whether the achievable performance looks right. Sometimes it doesn't. It's often the case that results are worse than in a full run.

This issue tracks the impact of the measures for the full run. Maybe we can work out what's causing the effect.

krausest:
For the first update since chrome 127 I decided to prepare the machine like for a full run. Results are pretty close and show how much noise is in the measurement.
Left incremental run, right last official run:
![Screenshot 2024-07-28 at 09 05 53](https://github.com/user-attachments/assets/4162a636-9109-4923-94ce-0cb9f65bc045)

krausest:
With indexing enabled and on battery, every else like like in a full run. A bit slower!
![Screenshot 2024-07-28 at 12 29 23](https://github.com/user-attachments/assets/636149b0-1914-4050-a9bd-65b7d62948f3)

krausest:
Once again, with indexing disabled and on AC:
![Screenshot 2024-07-28 at 13 10 10](https://github.com/user-attachments/assets/6b858ad4-d443-4ea6-b6c8-1c78b63b3ac4)
Really hard to tell if that's better than the last one. Nevertheless the full run was slightly better...

krausest:
Just running create 1k rows for vanillajs, but with 40 iterations for the following cases:

1.  Firefox, Chrome, Mail, Notes, Onedrive, Betterdisplay, VSCode, Preview, Settings
2. Firefox, Chrome, Mail, Notes, Onedrive, Betterdisplay, Preview, Settings
3. Only: Onedrive, Betterdisplay
4. No Apps running, Onedrive stopped, Betterdisplay stopped
5. After Rebooting: No Apps running, everything stopped except indexing
6. After Reboot: No Apps running, everything stopped incl. indexing 

[...]

PR #1725, opened 2024-08-13
Update Svelte version

trueadm:
As per the title, comes with some API changes too.

### DIFF
--- a/frameworks/keyed/svelte-classic/package-lock.json
+++ b/frameworks/keyed/svelte-classic/package-lock.json
@@ -13,7 +13,7 @@
         "@rollup/plugin-terser": "^0.4.4",
         "rollup": "^4.14.3",
         "rollup-plugin-svelte": "^7.2.0",
-        "svelte": "^5.0.0-next.174"
+        "svelte": "^5.0.0-next.220"
       }
     },
     "node_modules/@ampproject/remapping": {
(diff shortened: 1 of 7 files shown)


krausest:
Here's the incremental run:
Screenshot 2024-08-15 at 20 52 59

I ran on AC, but vanillajs was still slower than in the last full run, so with that as a reference I'd say no change for svelte. Maybe I'll try later and quit some background processes and results get closer to a full run. see #1721

krausest:
Here's a screenshot after a reboot on AC, most background processes stopped:
Screenshot 2024-08-15 at 22 15 35

vanillajs got faster, but it didn't help svelte.

trueadm:
Thanks for that. Separate note:

Is there something we can do with the really high variance on "select row"? I've been noticing it locally too, each run will result in a value that is almost +/- 1, which skews the geometric mean each time. Maybe that particular benchmark could run more times and exclude more outliers to reduce the variance?

krausest:
select row is indeed painful.
I'm already using two measures:
1. select rows takes 25 samples instead of 15 as for the other benchmarks
2. the weighted geometric mean reduces the influence of select row (see https://github.com/krausest/js-framework-benchmark/wiki/Computation-of-the-weighted-geometric-mean)

Here's an example for svelte and select rows what the average score would be if the duration was - and it's pretty close thanks to the weight:
| Example duration for select rows | factor value / fastest | total weighted score |
| --- | --- | --- |
| 4.9 |  1.88 | 1.162694559 |
| 3.9 |  1.5 | 1.150468554 |
| 5.9 |  2.27 | 1.17273775 |

Another check: A duration between 4.3 msecs and 5.1 msecs yields a total result of 1.16.

With a weight of 1 for each benchmark the total score would be 1.18 and 1.23 for 3.9 and 5.9. So I really think the weighted average makes sense.
 
Attached you can find an excel to play with the results.

Select rows is just too fast to find a winner, but it can identify slow frameworks. 

[results.xlsx](https://github.com/user-attachments/files/16644439/results.xlsx)

trueadm:
> select row is indeed painful. I'm already using two measures:
> 
> 1. select rows takes 25 samples instead of 15 as for the other benchmarks
> 2. the weighted geometric mean reduces the influence of select row (see https://github.com/krausest/js-framework-benchmark/wiki/Computation-of-the-weighted-geometric-mean)
> 
> Here's an example for svelte and select rows what the average score would be - and it's pretty close thanks to the weight:
> 
> Example duration for select rows	factor value / fastest	total weighted score
> 4.9	1.88	1.162694559
> 3.9	1.5	1.150468554
> 5.9	2.27	1.17273775
> Another check: A duration between 4.3 msecs and 5.1 msecs yields a total result of 1.16.
> 
> With a weight of 1 for each benchmark the total score would be 1.18 and 1.23 for 3.9 and 5.9. So I really think the weighted average makes sense.
> 
> Attached you find an excel to play with the results.
> 
> Select rows is just too fast to find a winner, but it can identify slow frameworks.
> 
> [results.xlsx](https://github.com/user-attachments/files/16644439/results.xlsx)

That definitely makes a lot of sense. Does making it 6x slowdown have any improvement on the metric if it's already so quick?

krausest:
I'll take a look at it when I'm back from vacation.

PR #1728, opened 2024-08-18
Upgrade Svelte

trueadm:
Updates to the latest version of Svelte.


krausest:

krausest:

PR #1729, opened 2024-08-19
perf: improve vue-vapor test code

LittleSound:
related https://github.com/vuejs/core-vapor/pull/268

Improve the test code based on the architectural features of Vue Vapor

This can more accurately reflect the best situation. Then we will try to fix some of the slower parts.

This is the result of my computer:

image


krausest:

krausest:

PR #1730, opened 2024-08-21
Updated Vanillajs-3 and Deleight implementations

mksunny1:
I found some ideas to improve both implementations. They show good improvements locally.

### DIFF
--- a/frameworks/keyed/deleight/index.html
+++ b/frameworks/keyed/deleight/index.html
@@ -79,6 +79,7 @@
             <span class="preloadicon glyphicon glyphicon-remove" aria-hidden="true"></span>
             </div>
           </div>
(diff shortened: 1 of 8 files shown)


krausest:

mksunny1:
Sorry just saw this.

I am a bit surprised by the partial update performance of vanillajs-3 here. I really thought I 'over-optimised' that taking advantage of how the benchmarks run. 

I did notice something from running the benchmarks locally. It appears that the overall results tend to get better or worse depending on the performance of individual benchmarks. My point is that the results for individual benchmarks might not be isolated enough from each other.

I feel the results could be very different / more indicative if the combined benchmarks are run 'laterally'. Something like running `create rows` for all the frameworks first, then running `update rows` and so on instead of running everything for individual benchmarks.  Perhaps this is how you run the official results, I am not sure. But that is not how we run it locally.

PR #1731, opened 2024-08-28
Add Endr

caseywebdev:
Endr is my attempt at a lightweight React-inspired library that tries to focus on (what I think are) the best parts of React and ditching the rest. I think it's interesting to see how it performs relative to the rest of the pack. Let me know if any changes are necessary!


krausest:

caseywebdev:
Awesome, thanks!

krausest:

PR #1733, opened 2024-09-03
update pota

titoBouzout:
Hi Stefan, thanks for your time

I took the opportunity to change the formatting to what I use, and simplify the roll-up config, but I didn't touch the package lock.

On this version, I was able to get rid of a placeholder attribute that worked well but was slowing it a bit. Also noticed by luck that it was trying to remove the class danger on initialization, as it defaults to false. It made me implement a prev/next logic that I already saw in solid long ago but mostly forgotten. 

The _weird_ attribute which was a `hasAttribute` and `removeAttribute` call was consuming ~25ms, and the class removal some ~10ms. So on my tests it went from ~135ms to ~100ms, solid is at ~70ms so 30 more to glory but I consider that unreachable. 

Ive been hesitant of the benchmark and dont like _to code for it_, but it did highlight real issues and Im grateful about it. :)

Currently ranking 43 of 153, not bad, looking forward for the update

Thanks

### DIFF
--- a/frameworks/keyed/pota/package.json
+++ b/frameworks/keyed/pota/package.json
@@ -1,14 +1,12 @@
 {
   "name": "js-framework-benchmark-pota",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "type": "module",
   "main": "dist/main.js",
   "js-framework-benchmark": {
     "frameworkVersionFromPackage": "pota",
     "frameworkHomeURL": "https://pota.quack.uy/",
-    "issues": [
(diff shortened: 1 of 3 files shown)


krausest:

krausest:

titoBouzout:
Thanks, I expected a bit better, but probably these improvements get lost in all the noise as I am measuring only the lib runtime, not the dom/browser, maybe something else is happening, I should probably start looking at garbage collection some day.

PR #1740, opened 2024-09-06
Add react-compiler-hooks (React 19 RC)

vasylenkoval:
Adds a React benchmark that uses [React Compiler](https://react.dev/learn/react-compiler) babel plugin.

**Notes:**
- This benchmark aims to be as simple as possible and demonstrate performance without (almost) any manual memoization. 

- It's not using the `useReducer` [trick](https://github.com/krausest/js-framework-benchmark/blob/master/frameworks/keyed/react-hooks/src/main.jsx) of passing single `dispatch` to all children. While it's a valid pattern I don't think it reflects the majority of the "regular" react code as it forces children to know action types and how to call dispatch which couples them to this interface. This benchmark is keeping it simple with just `useState` and regular functions created during render, which, in my opinion, reflects the vast majority of code written in the wild.

- The only component that uses the `memo` HOC is the `Row` component. If that component is not wrapped, code produced by React Compiler will invalidate cached row elements in the parent when the reference to the list of items changes and start rendering the Row component even if its props have not changed. While the Row is still optimized for re-renders after being processed by the compiler, not short-circuiting the render earlier ends up costing more.

### DIFF
--- a/frameworks/keyed/react-compiler-hooks/index.html
+++ b/frameworks/keyed/react-compiler-hooks/index.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>React Compiler Hooks</title>
+    <link href="/css/currentStyle.css" rel="stylesheet" />
+  </head>
+  <body>
+    <div id="main"></div>
+    <script src="dist/main.js"></script>
+  </body>
+</html>
--- a/frameworks/keyed/react-compiler-hooks/package-lock.json
+++ b/frameworks/keyed/react-compiler-hooks/package-lock.json
@@ -1,0 +1,4337 @@
+{
+  "name": "js-framework-benchmark-react-compiler-hooks",
+  "version": "1.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "js-framework-benchmark-react-compiler-hooks",
+      "version": "1.0.0",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "react": "19.0.0-rc-4c58fce7-20240904",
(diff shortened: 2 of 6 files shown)


ryansolid:
The fact the memo is still needed for the row component is by far the most interesting observation. This whole benchmark centers around the row. I suspect it is the place where post compiler people will miss adding it.

I agree in general the reducer pattern is a bit specific in the hooks implementation. When I originally put it together it was what Dan Abramov suggested. The compiler likely removing the desire for that optimization feels good.

krausest:
Thank. Hope to post results soon!

krausest:

As one can see from vanillajs the run wasn't very good (though on AC, indexing off, but without rebooting) so I'd expect faster results for the next full run. So it's hard to say if it's actually slower than the react-hooks implementation. We'll have to wait for the next full run.

vasylenkoval:
@ryansolid I was very surprised that it doesn't try to cache each element in the list and instead invalidates the whole list. 

Here is the unminified logic to showcase it better:

Screenshot 2024-09-06 at 11 15 05 AM

I imagine that there is a cost to fanning out these comparisons in terms of both complexity and bundle size that they are trying to balance, however, it does seem like they could have wrapped the array items in `memo` equivalent inside of the compiler produced code to delegate the caching to another unit of work downstream. Maybe decided not to do it for consistency with how non-array caching is done? Would be interesting to hear someone talk about the hard parts of this implementation.

@krausest Wow, thank you for merging and posting the results so quickly! 

When running locally the performance was similar-ish to `react-hooks` for updates:
Screenshot 2024-09-06 at 11 59 05 AM

I would not be surprised if it was slower on average. `react-hooks` implementation is well optimized! It would be interesting to see a comparison to just vanilla react without any memoization APIs used and no compiler. Happy to submit that if you think it would be valuable.

PR #1741, opened 2024-09-06
chore: update lwc to v7.2.6

nolanlawson:
This is just a regularly-scheduled dependency update.

Thank you as always for continuing to maintain this repo! 🙇


krausest:
Please note that vanillajs was slower on the left. I'm a bit disappointed with that because I'm trying to solve https://github.com/krausest/js-framework-benchmark/issues/1721, but as you can see vanillajs perfomed was 0.2 slower (even after reboot and on AC). So I think it's plausible to assume that performance for all updated versions would be better in a full run.

PR #1593, opened 2024-01-28
Update GXT to 0.0.37

lifeart:
image

Prev: https://github.com/krausest/js-framework-benchmark/pull/1561

PR #1743, opened 2024-09-14
chore: update gtx to 0.0.57

lifeart:
image

image

Bumping `gxt` from `v37` to `v57`, memory usage & cleanup logic should be improved.
Also, see some regression in create-many case.

Prev: https://github.com/krausest/js-framework-benchmark/pull/1593

### DIFF
--- a/frameworks/keyed/gxt/package-lock.json
+++ b/frameworks/keyed/gxt/package-lock.json
@@ -9,15 +9,15 @@
       "version": "0.0.1",
       "license": "MIT",
       "devDependencies": {
-        "@glint/core": "1.3.0",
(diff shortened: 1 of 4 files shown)


krausest:
Please note that vanillajs was slower on the left. I'm a bit disappointed with that because I'm trying to solve https://github.com/krausest/js-framework-benchmark/issues/1721, but as you can see vanillajs perfomed was 0.2 slower (even after reboot and on AC). So I think it's plausible to assume that performance for all updated versions would be better in a full run, especially for gxt which was even a bit faster in the incremental run.

lifeart:
@krausest thank you for merge, run and additional details!

PR #1744, opened 2024-09-19
update vue-vapor

sxzz:
Vue Vapor's performance has improved.


krausest:
Please note that vanillajs was slower on the left. I'm a bit disappointed with that because I'm trying to solve https://github.com/krausest/js-framework-benchmark/issues/1721, but as you can see vanillajs perfomed was 0.2 slower (even after reboot and on AC). So I think it's plausible to assume that performance for all updated versions would be better in a full run. Vue-vapor was faster even under those circumstances.

antonmak1:
@sxzz Hi! It's really cool that you're promoting a topic without VDOM! I was so blown away by this comment, like VDOM isn't crap, that it's fast, it's just that js-framework-benchmark is a disgrace that it's being compared to it in terms of speed. It's just mind-blowing.

antonmak1:
@sxzz Like, I just came across this comment and was just shocked. https://x.com/dan_abramov2/status/1796156983575232942

antonmak1:
@krausest It turns out that the benchmark is crap, that's how it is, it doesn't show the speed. It's a shame what we're doing. I'm just shocked. This stinking VDOM is being promoted to the masses tightly, then billions of people load the site a second or two faster, and that's millions of years of time saved for people.

antonmak1:
For convenience, people simply waste millions of seconds of time that could be saved.

PR #1747, opened 2024-09-21
Doo html dom v0.92.2 beta september 21 2024

hman61:


krausest:
Please note that vanillajs was slower on the left. I'm a bit disappointed with that because I'm trying to solve https://github.com/krausest/js-framework-benchmark/issues/1721, but as you can see vanillajs perfomed was 0.2 slower (even after reboot and on AC). So I think it's plausible to assume that performance for all updated versions would be better in a full run. doohtml is also in the 0.2 - 0.3 range which could be due to the difference between the last full run and the vanillajs results from the incremental run.

PR #1752, opened 2024-10-21
Upgrade to Svelte 5 final

benmccann:


krausest:
Thanks! I merged it during the preparation for chrome 130 and thus results are included in the c[hrome 130 run](https://krausest.github.io/js-framework-benchmark/2024/table_chrome_130.0.6723.58.html).

PR #1754, opened 2024-10-23
update pota

titoBouzout:
Hi again! if anything it will get slower... 

dx: 
- it batches event listeners by default, so no need to batch as before

stuff not measured/used in the benchmark:
- better handling of custom elements attributes/properties
- better handling of `value` so the button `reset` works in forms
- so more updates unrelated to the benchmark that I do think it should be included here, but you are welcome to read the change log https://github.com/potahtml/pota/commits/master/

Maybe related to the benchmark is that I am making the arrow functions not return a value, we will see if that makes any difference

Thanks for everything!

### DIFF
--- a/frameworks/keyed/pota/index.html
+++ b/frameworks/keyed/pota/index.html
@@ -3,7 +3,10 @@
   <head>
     <meta charset="utf-8" />
     <title>Pota</title>
-    <link href="/css/currentStyle.css" rel="stylesheet" />
+    <link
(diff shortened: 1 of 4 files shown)


krausest:
Thanks! I merged it during the preparation for chrome 130 and thus results are included in the c[hrome 130 run](https://krausest.github.io/js-framework-benchmark/2024/table_chrome_130.0.6723.58.html).

PR #1755, opened 2024-10-24
Bump Michijs to v2

lsegurado:
- Bump Michijs to v2
- Deprecated michijs-map variant


krausest:
Thanks. Here's a comparison:
Nevertheless michijs v2 is so much slower that I really believe there's a performance regression.

PR #1758, opened 2024-10-26
Bump sonnet to `0.0.33`

13770129:
Bumped sonnet to `0.0.33` 🐈‍⬛


krausest:
My test tells me that the non-keyed version is actually keyed. Can we just remove the non-keyed table? I mean it's definitely not a shame to be proud of being only keyed 😄

krausest:
Thanks. Here's a comparison:
Sonnet is 0.3 slower than in the full run and now on the level of vanillajs. We'll see in the next full run whether it sill outperforms vanillajs and vanillajs-3.

PR #1759, opened 2024-10-26
Bump mikado to `0.8.400`

13770129:
Bumped mikado to `0.8.400` 🐈‍⬛


krausest:
Thanks. Here's a comparison:
I'd say mikado is about as fast as before.

PR #1761, opened 2024-10-29
update solid and solid-store

birkskyum:
This is only minor releases, no code changes needed.


krausest:
Thanks. Here's a comparison:
Solid is also rather close to the full run.

PR #1763, opened 2024-11-02
Bump Michijs to 2.0.8 - Performance improvements

lsegurado:
Found some performance improvements - should be better now. Also it should fix the CSP issue


krausest:

krausest:
Nice improvement!

PR #1764, opened 2024-11-04
Add @rvjs/core

ILikeMintChocolate:
Hi!
I've developed a library called @rvjs/core based on the concept of only updating areas where the state is used.
Hope it integrates smoothly. Thanks!


krausest:

krausest:
Not bad, maybe there's a way to optimize clear rows?

ILikeMintChocolate:
I'm continually working on optimizing performance, but it's not easy right now. It seems like I'll need to make major changes to optimize clear rows.

PR #1765, opened 2024-11-05
Add keyed/strve-rv

maomincoding:
Strve-rv is a reactivity library for building user interfaces on the web.

Powered by [@vue/reactivity](https://github.com/vuejs/core/tree/main/packages/reactivity) and [strve-js](https://github.com/strveJs/strve).

### DIFF
--- a/frameworks/keyed/strve-rv/.gitignore
+++ b/frameworks/keyed/strve-rv/.gitignore
@@ -1,0 +1,25 @@
+.DS_Store
+node_modules
+/dist
+/buildDir
+.parcel-cache
+
+# local env files
+.env.local
+.env.*.local
+
+# Log files
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+
+# Editor directories and files
+.idea
+.vscode
+*.suo
+*.ntvs*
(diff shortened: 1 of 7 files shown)


maomincoding:
Thank you, author

krausest:

maomincoding:
Okay, I'm really looking forward to the results.

krausest:
I added a #801 note due to the explicit event delegation.

maomincoding:
Thank you. When can I access the website for the results?

maomincoding:
> Sooner than expected. Here's the comparison (vanilla and vanillajs-3 look OK for an incremental run) with the new / updated results on the left and the old version on the right: Screenshot 2024-11-11 at 08 11 41 I added a #801 note due to the explicit event delegation.

Thank you for pointing out the point of explicit event delegation. I did not display the mechanism of automatic event delegation in the framework. I would like to enable users to manually optimize it. I specifically wrote this code for benchmark testing. If we consider the accuracy and rigor of benchmark testing, then my approach is indeed not very good for other frameworks that have done automatic event delegation. We should not excessively manually optimize for the sake of optimization, as it would lose the meaning of testing.

PR #1766, opened 2024-11-07
Update cample to v3.2.1-beta.1

antonmak1:
Okay. I'll do what I can. Namely, just refine the code to an ideal state. For myself, I've already solved the problem. In general, in fact, a benchmark is a mathematical problem that many people solve differently. @krausest No matter what anyone says, you've created a really useful repository that solves one of the main problems today - speed. In fact, this has long been solved in mathematics, it just needed to be depicted in code, so people began to practice in variations of the solution. Today, it is already obvious that the problem has been solved. Even 2-4 years ago, or maybe earlier @ts-thomas @xania @Freak613 @hman61 @alexcrichton, or someone else had already solved this problem. I'm not the first, I'm not the last, or you're not the only one who has had to do it, call it what you want. I hope the main point is that the ideal algorithm should be the standard in modern frameworks, so that they all work perfectly fast by default. I just realized this, but it's just super obvious. It makes no difference at all how to implement it when the backbone is already created. Obviously, many frameworks can already use the ideal algorithm, it just takes time.

### DIFF
--- a/frameworks/keyed/cample/package-lock.json
+++ b/frameworks/keyed/cample/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "js-framework-benchmark-cample",
-  "version": "3.2.1-beta.0",
+  "version": "3.2.1-beta.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "js-framework-benchmark-cample",
-      "version": "3.2.1-beta.0",
+      "version": "3.2.1-beta.1",
       "license": "MIT",
       "dependencies": {
-        "cample": "3.2.1-beta.0"
+        "cample": "3.2.1-beta.1"
       },
       "devDependencies": {
         "@babel/core": "7.23.9",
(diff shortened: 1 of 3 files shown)


antonmak1:
I wrote an [article](https://dev.to/antonmak1/js-framework-benchmark-variations-of-the-ideal-solution-to-the-mathematical-problem-of-speed-or-why-it-is-standard-2j3d) about speed because it's important.

krausest:

krausest:
I'm afraid it's a bit slower than the old version. 0.4 is a bit larger than the usual noise (something like 0.1 up to 0.2). Any idea whether it's actually slower or do you think it's due to noise?

antonmak1:
@krausest Yes, it's a normal noise. It's happened several times already. It's obvious just by "clear rows", it can't be 14.7, because I didn't change anything there. There's nothing to worry about, just a normal startup alone

antonmak1:
@krausest Happy New Year!

PR #1768, opened 2024-11-09
add kr-observable

nihil-pro:


krausest:

krausest:
Sooner than expected. Here's the comparison (vanilla and vanillajs-3 look OK for an incremental run) with the new / updated results on the left and the old version on the right (so it's not relevant to you):
I guess that partial update, swap and select row do a full update. Maybe you can optimize that.

nihil-pro:
I am thinking about it. It seems that the problem is in the Proxy([]), I tested locally just wrapping the array in a proxy, and the result is not much better. Thanks

nihil-pro:
Screenshot 2024-11-16 at 21 29 26
Screenshot 2024-11-16 at 21 29 54
Done ))
Now is fast enough

PR #1769, opened 2024-11-10
Update pota

titoBouzout:
It avoids trying to sort when only appending, I have missed that case, thanks!


krausest:

krausest:
I'd say there's no visible change for pota.

titoBouzout:
Thanks, I see. seems this case was already handled by groupBy, all this does then is avoiding creating the grouped objects. I will leave that code there because its small enough. 

Thanks!

PR #1770, opened 2024-11-11
Adjust the main logic code

maomincoding:
#801 
Make corresponding adjustments for this point.


maomincoding:
#801

maomincoding:
The positioning of each framework is different. For example, my framework pursues flexibility and ease of use, but if this degree is not well mastered, the design will be difficult to understand. So, the trade-off between frameworks is very important.

krausest:
As expected a bit slower.

PR #1772, opened 2024-11-15
Update vue

jeremio:
Updates Vue to the latest version


krausest:
Usually patch releases show are uninteresting regarding performance, but in this case I'm not sure if the 0.4 difference is noise or whether it's really a bit slower. Maybe I'll run it again.

krausest:
Did that, but vue didn't perform better. We'll see it clearer in the next full run.

PR #1775, opened 2024-11-16
update pota

titoBouzout:
Should in theory improve deletion/clearing. There may be an improvement in memory, but not reallly sure.  There are also a bunch of stylish changes in the template and goes back to use a namespace for the `danger` class. Thanks!


krausest:
No significant change I'd say.

titoBouzout:
Thanks for the update, yeah I was thinking when doing this if its worth the code size increase. As you said it may not be significant. From 17 to 15, is closer to 13.5 vanilla. oh well

PR #1778, opened 2024-11-23
update ivi and switch template to @ivi/htm

leeoniya:
this switches ivi to use its [@ivi/htm](https://github.com/localvoid/ivi/blob/master/packages/@ivi/htm/README.md) template syntax, which i think makes this awesome framework feel a lot less esoteric. the compiled bundle is identical.

i've kept the `.textContent` prop optimizations in place, but will do a follow-up PR to see how much omitting these affects performance. (@localvoid says memory usage would go up slightly)

### DIFF
--- a/frameworks/keyed/ivi/package-lock.json
+++ b/frameworks/keyed/ivi/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.0",
       "license": "Apache-2",
       "dependencies": {
-        "@ivi/tpl": "3.0.1",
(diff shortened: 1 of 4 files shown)


krausest:
From 1.11 to 1.11 is acceptable I'd say. Results are [up](https://krausest.github.io/js-framework-benchmark/2024/table_chrome_131.0.6778.85.html).

leeoniya:
heh, the compiled bundle is identical, so that's great news...for the benchmark itself 😅

PR #1779, opened 2024-11-26
Update pota, implements throwing `moveBefore`

titoBouzout:
This change implements the new "node state preserving api" `moveBefore`.   see: https://github.com/whatwg/dom/issues/1255

This is the set of changes I had to do https://github.com/potahtml/pota/commit/51da4b85e57d3eb56d1b0ec5554adb88d1119a9a

Unfortunately, they are making it throw., you check a ridiculous amount of conditions or just wrap it in a try and catch, I suspect the performance of the swap rows will degrade on frameworks implementing  this. The benchmark unfortunately only tests 2 moves[when the framework does the optimal thing]. So I'm not sure if it would be representative of the degradation. I am still curious what the results would be.

Thanks!

### DIFF
--- a/frameworks/keyed/pota/package.json
+++ b/frameworks/keyed/pota/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-pota",
-  "version": "1.0.5",
+  "version": "1.0.6",
   "type": "module",
   "main": "dist/main.js",
   "js-framework-benchmark": {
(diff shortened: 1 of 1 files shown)


WebReflection:
imho, we should not implement stuff that has not officially landed yet or we'll make "*their game*" in doing so, when the "*it's already shipped, too late to change it*" card will be thrown next round of complains.

WebReflection:
I want to believe until it’s out it’s not as bad as it currently is, seeing lit, or others, already dogfooding that broken method name hurts back, doesn’t feel or sound as a win for anyone involved

titoBouzout:
You are right, just reverted, thanks for your comments

PR #1780, opened 2024-11-28
Bump michijs to 2.1.4

lsegurado:
More performance improvements


krausest:
Vanillajs* performed a bit slower in the incremental run (how I hate that sentence...), so there's definitely a large improvement.

PR #1781, opened 2024-11-28
Added vanilla implementation with tc39 signals

lsegurado:
I wanted to share a VanillaJS implementation using the TC39 Signals polyfill.


krausest:
Thanks. It's included on the left
Any chance of moving from michjs-build to some standard build or even go with modules only?

PR #1783, opened 2024-12-01
Update quel to version 0.23.1 and refactor code

mogera551:
Upgrade the quel framework to version 0.23.1, refactor component registration, and rename the ViewModel class to State for improved clarity.


krausest:
Vanillajs* performed a bit slower in the incremental run (how I hate that sentence...), which makes the improvement for quel even better.

PR #1511, opened 2023-11-21
Update Blazor to .NET 8

birkskyum:
closes #1509 

Trivial update, only changed .net version number according to the [migration guide](https://learn.microsoft.com/en-us/aspnet/core/migration/70-80?view=aspnetcore-7.0&tabs=visual-studio#update-the-net-sdk-version-in-globaljson).

Didn't run all tests, but it seems like there might be a > 10% performance improvement here

Screenshot 2023-11-21 at 02 06 16

PR #1784, opened 2024-12-02
update to .NET 9 for blazor-wasm and blazor-wasm-aot

birkskyum:
Similar to

- https://github.com/krausest/js-framework-benchmark/pull/1511

This version appear to use [Emscripten 3.1.56](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#3156---031424)


krausest:
Definitely Faster! But still very far off. The checks even timed out on my VPS. Seems to need a very powerful machine just to run the benchmark :(

PR #1785, opened 2024-12-06
update leptos to 0.7

gbj:


krausest:
Vanillajs* also performed a bit slower (how I hate that sentence...), but it really seems like leptops got a litte slower.

gbj:
Ha, I think it was a 1.20 on my machine.

I think we'd seen the Leptos benchmark drift from ~1.15 to 1.30-something over the last year without any meaningful updates to the code (or without any updates at all) — and this was the case for other WASM frameworks as well — so I'm not really surprised. It seems like something in the benchmark has just changed enough over time that it's not capturing particularly useful data for the WASM-based frameworks any more, IMO. But I think it's still worth keeping them up to date.

PR #1805, opened 2024-12-16
Updated package.json to correct the repo location

mksunny1:
I have made the following changes to this fork:

1. Updated package.json of keyed deleight implementation to use the correct repo location.

2. Re-implemented the non-keyed implementation to use the latest version of deleight (v5.5.10).

### DIFF
--- a/frameworks/keyed/deleight/package.json
+++ b/frameworks/keyed/deleight/package.json
@@ -4,7 +4,7 @@
   "description": "Deleight demo",
   "main": "src/Main.js",
   "js-framework-benchmark": {
-    "frameworkHomeURL": "https://github.com/markersunny/deleight",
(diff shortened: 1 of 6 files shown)


krausest:
Just for completeness here's the result (left updated keyed, right old keyed):
There were no relevant changes so it's good to see that results are very close to the old ones.

mksunny1:
Happy to see it running slightly faster &#128512;. I mostly just tried to organise the library so it would be easier to use. Also the bundle size I would hope/expect it should be smaller now, since the imports were more precise.

I still have some things to change before I am fully satisfied with my work, but they are not going to affect the ranking and they won't break the current implementation., because they belong in other modules. 

In the end, I would love to see something like the `apply` function in Vanilla JavaScript.

PR #1809, opened 2024-12-20
Add TargetJ to js-framework-benchmark

wasfi2:
This requests adds the TargetJ framework to the js-framework-benchmark repository.

**Key changes:**
index.html: contains the main HTML structure
src/main.js: implements benchmark functionality using TargetJ.
src/data.js: creates the data sample
package.json and package-lock.json: define the dependencies

### DIFF
--- a/frameworks/keyed/targetj/index.html
+++ b/frameworks/keyed/targetj/index.html
@@ -1,0 +1,84 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8">
+        <title>TargetJ</title>
+        <link href="/css/currentStyle.css" rel="stylesheet"/>
+        <script src='dist/main.js'></script>  
+    </head>
+    <body>
+        <div id="main">
+            <div class="container">
+                <div class="jumbotron">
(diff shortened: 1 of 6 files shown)


wasfi2:
TargetJ is available at https://github.com/livetrails/targetj. Feel free to reach out if you need more information.

krausest:
Thanks. To conform with the keyed implementations the swap operation must actually swap the corresponding dom nodes. It can be invoked in the webdriver-ts folder with the following command:
```
```
targetj-v1.0.118-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results

wasfi2:
Thank you.. I will fix it over the next few days. I thought it would be included in the keyed tests.

wasfi2:
I think I resolved the swap test issue. I couldn't run npm run isKeyed keyed/targetj as it required Chrome 100, and I wasn't sure how to set that up. However, I checked all the JSON files under webdriver-ts/results, and they all show 'keyed,' so I hope I addressed he issue.

PR #1813, opened 2024-12-24
destam-dom: Update dependencies

Nefsen402:


krausest:
Here are the results (left updated keyed, right old keyed):
Very close to the usual noise, but maybe a little bit slower than the old version.

Nefsen402:
That's mostly expected. The efforts were mostly in memory usage optimizations. Since the tests are run on a fresh tab, the garbage collector doesn't play as much as a role versus a longer term test.

PR #1814, opened 2024-12-27
update pota

titoBouzout:
I've decided to namespace all event handlers., that lead to some changes. The reason is to easily disambiguate between event handlers, properties, and attributes.

Question, is there something I can do so version '0.17.x' always gets updated before making a full run? That's to be, to always update `0.17.x` to whatever patch version pota is currently in.

Oh, btw, I do not need a partial run, I know these runs are slower, I can wait till the next full run.  Partial runs are not useful to me as these behave differently to full runs. I know there's an issue open about this and I do not mind, I can just wait.

Thanks!

### DIFF
--- a/frameworks/keyed/pota/package.json
+++ b/frameworks/keyed/pota/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-pota",
-  "version": "1.0.5",
+  "version": "1.0.6",
   "type": "module",
   "main": "dist/main.js",
   "js-framework-benchmark": {
@@ -20,7 +20,7 @@
     "url": "https://github.com/krausest/js-framework-benchmark.git"
   },
   "dependencies": {
-    "pota": "^0.16.163"
+    "pota": "^0.17.170"
   },
   "devDependencies": {
     "@babel/core": "7.25.2",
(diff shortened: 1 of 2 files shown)


krausest:
I used to do regular full builds, but they always broke. I tried to use only fixed versions, but they also broke with fixed versions. Thus we used npm ci with package-lock.json checked in and - although less often - they break even with package-lock.json. So I'm afraid currently the only way to update is to make a PR.
Just as if the JS god wanted to add weight to my statement, I ran into the following build error with your PR:
```
npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: js-framework-benchmark-pota@1.0.6
npm error Found: @babel/core@7.25.2
npm error node_modules/@babel/core
npm error   dev @babel/core@"7.25.2" from the root project
npm error
npm error Could not resolve dependency:
npm error peer @babel/core@"^7.26.0" from pota@0.17.170
npm error node_modules/pota
npm error   pota@"^0.17.170" from the root project
```
I was able to fix it by bumping @babel/core 😄

krausest:
Just for completeness here's the result (left updated keyed, right old keyed):

titoBouzout:
On the other hand if a partial run is better than a the latest full run then we are up to something! Thanks Stephan and best wishes for this holidays!

PR #1815, opened 2024-12-28
Adding skruv/liten benchmarks

SvanteRichter:
This adds skruv/liten benchmarks, a abstraction I built for smaller stuff than my other framework skruv.

I did a few things for size that other frameworks might not do (inline into the html for better compression and latency, etc). Those are things that I started to do in my own apps for better perf, but let me know if that is an issue here.

In my testing it should set a new bar for smallest in non-keyed.


krausest:
No problem. Actually I don't know what went wrong. In the last run all benchmarks for skruv-liten ran into an error. After that I just tried to build again from a npm ci and found that this worked. I don't know what it was, but the effect was that I had to run it after the original benchmark run. I hope to mitigate #1721 by making sure that for one benchmark all frameworks are run in one batch (such that the system is in a constant state), which didn't happen now to skruv-liten.
Long story short: I don't think there's anything you can do and I don't think there was anything wrong about skruv-liten. Thanks for asking.

PR #1820, opened 2025-01-16
`karyon` update

syduki:
This updates the `karyon` module to the latest version, also a memory leak fix in the framework implementation + some refactoring.


krausest:
It could be that it's a little bit slower than 3.0.

PR #1821, opened 2025-01-19
update pota

titoBouzout:
It doesn't attempt to sort the rows when a row is deleted, so "remove row" should in theory improve, the other changes I somewhat unsure if are going to make any difference.


krausest:
Since vanillajs* was a bit slower in this incremental run, maybe pota indeed improved a litte bit.

PR #1822, opened 2025-01-20
Update @rvjs/core version

ILikeMintChocolate:
I changed some parts of how @rvjs/core works and made the performance better.

Thanks!


krausest:
Nice improvement. Could even perform something like 0.2 better in the next full run (vanillajs* were a bit slower than in the last full run).

PR #1823, opened 2025-01-21
Add Aurelia 2 to non-keyed frameworks.

Vheissu:
Adds Aurelia 2 into the non-keyed frameworks list.

### DIFF
--- a/frameworks/non-keyed/aurelia2/vite.config.js
+++ b/frameworks/non-keyed/aurelia2/vite.config.js
@@ -14,7 +14,7 @@ export default defineConfig({
   },
   plugins: [
     aurelia({
-      useDev: true,
+      useDev: false,
     }),
     babel(),
     nodePolyfills(),
--- a/frameworks/non-keyed/aurelia2/vite.config.js
+++ b/frameworks/non-keyed/aurelia2/vite.config.js
@@ -1,0 +1,22 @@
+import { defineConfig } from 'vite';
+import { nodePolyfills } from 'vite-plugin-node-polyfills'
+import aurelia from '@aurelia/vite-plugin';
+import babel from 'vite-plugin-babel';
+
+export default defineConfig({
+  server: {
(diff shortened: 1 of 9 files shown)


frameworks/non-keyed/aurelia2/vite.config.js
```diff
@@ -0,0 +1,22 @@
...
+import { nodePolyfills } from 'vite-plugin-node-polyfills'
+import aurelia from '@aurelia/vite-plugin';
+import babel from 'vite-plugin-babel';
+
+export default defineConfig({
+  server: {
+    open: !process.env.CI,
+    port: 9000,
+  },
+  base: '/frameworks/non-keyed/aurelia2/dist/',
```
Vheissu, line 11: I had to add this or when the built app was run, it was trying to reference the bundled assets from the root, instead of using the `aurelia2/dist` directory as the root.

krausest:
Thanks. Very welcome. Turns out that aurelia2 is keyed which is really good news. I moved it to better side of the benchmark table :)
One minor issue. `npm ci` fails on my VPS. Any idea why?
```
running 'npm ci' in 'frameworks/keyed/aurelia2'
npm error code EUSAGE
npm error
npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm error
npm error Invalid: lock file's rollup@4.32.0 does not satisfy rollup@3.29.5
npm error Missing: rollup@4.32.0 from lock file
npm error Missing: rollup@4.32.0 from lock file
npm error
npm error Clean install a project
```
It's not a show stopper but I don't understand why this happens. The package-lock.json stays even unchanged after running `npm install`. Any idea? (Node v20.17.0 on Ubuntu Linux)

krausest:
Good to see aurelia is coming to the keyed camp. Performance is a bit behind react and angular though.

PR #1826, opened 2025-01-25
Removing Sling.js as no longer developed

puckowski:
* Removing Sling.js as it is no longer actively being developed.

Simple removal of non-keyed slingjs files.


krausest:
Thanks!

PR #1827, opened 2025-01-27
migrate ui5-webcomponents to 2.x version

pskelin:
the 2.x version of ui5 web components makes some changes to the project structure that are reflected with this change.

@krausest I am making the fixes in the `broken-frameworks` folders to see the diff easier. Moving the folders to `frameworks` should work automatically. If you want to, I can do it in the same change, or you can do it separately like you moved them from `frameworks` to `broken-frameworks` but in reverse. Let me know please.


krausest:
Thanks a lot. Here's a screenshot from the result:
Screenshot 2025-02-11 at 20 57 51

PR #1829, opened 2025-01-29
Update targetjs

wasfi2:
This is a new implementation of TargetJS. It is a lighter implementation than the previous one. I also now use the click event instead of pointer up events.


krausest:
Thanks a lot, well worth the wait. Here's a screenshot from the result:
Screenshot 2025-02-11 at 20 57 51

wasfi2:
I agree with #772.
#800 I am not sure if I understand it. Can you please elaborate more?
#801. Events are not directly tied to the DOM. They trigger a task cycle process. The process then collects all the events happen before it runs, so the events can be handled synchronously by the visible models.

krausest:
If an implemenation moves the selection state to the row (instead of the table) then it's a candidate for #800 
But I think you're right. #772 might be enough - the selected class is directly set on the node.
So I'll just add #772.

wasfi2:
Sounds good. Thanks for clarifying.

wasfi2:
I noticed the results have been published. Thanks a lot for getting it done so quickly!

PR #1836, opened 2025-02-20
Remove react-starbeam

NullVoxPopuli:
This is using dev assets, and I just didn't notice until later.

It's not representative of the production-build experience, and since starbeam is (right now) a research project anyway, I think it's best to remove the react-version of this for now.

This was introduced here: https://github.com/krausest/js-framework-benchmark/pull/1073


krausest:
Thanks!

PR #1843, opened 2025-03-04
feat: update viewfly

viewfly:


krausest:
Left updated results (control group + updated frameworks), on the right old results:
Screenshot 2025-03-09 at 09 14 22
Nice improvement!

PR #1844, opened 2025-03-04
remove ivi micro-optimizations

leeoniya:
this removes:

- `/* preventClone */` annotations
- `List`-less additional tbody template
- `.textContent=${text}` on the Button elements

hopefully @localvoid agrees that these are not worth their cognitive weight here.

### DIFF
--- a/frameworks/keyed/ivi/src/main.ts
+++ b/frameworks/keyed/ivi/src/main.ts
@@ -65,6 +65,7 @@ interface RowProps {
 const Row = component<RowProps>((c) => {
   const onSelect = () => { dispatch(c, { type: ActionType.Select, entry: getProps(c).entry }); };
   const onRemove = () => { dispatch(c, { type: ActionType.Remove, entry: getProps(c).entry }); };
+
   return ({ entry, selected }) => html`
     <tr class=${selected === true ? "danger" : ""}>
       <td class="col-md-1" .textContent=${entry.id}/>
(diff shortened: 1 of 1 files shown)


localvoid:
> hopefully @localvoid agrees that these are not worth their cognitive weight here.

:+1:

krausest:
Left updated results (control group + updated frameworks), on the right old results:
Screenshot 2025-03-09 at 09 14 22
Control group was slightly slower, so maybe +0.2 in score without those optimizations.

leeoniya:
thanks, interesting. kind of odd to see that these changes have a measurable effect since two of the three don't affect rows. the only one that might be legit is clear rows, but i would expect ivi's `List` to be smart enough to handle the non-empty-array -> empty array case optimally, @localvoid?

curious to see what the next full run shows.

localvoid:
> but i would expect ivi's `List` to be smart enough to handle the non-empty-array -> empty array case optimally

In real world scenarios when table/list is empty, it is usually implemented as:

```js
  isEmpty
    ? <ul>{...}</ul>
    : <p>Empty list</p>
```

I've removed `textContent=""` optimization couple of years ago. In my opinion, `textContent=""` optimization is useful only in replace scenarios, but I am pretty sure that nobody is using it in such scenarios.

Also, replacing container node instead of using `textContent=""` optimization significantly reduces mutation events and outside of a benchmarking sandbox it will be way much faster.

PR #1852, opened 2025-03-18
doohtml-lite-v0.96.1 Initial preparation Doo from class to fn based

hman61:
Simplifying , removing option to run in shadow,  and preparing to convert it to make it function based instead of class based


krausest:
Thanks (on the left updated keyed results, on the right last results for updated frameworks):
Screenshot 2025-03-22 at 21 07 55

Not too bad 😄 . 
doohtml has always been a challenge for the shadow dom access in my test driver...

krausest:
Would be great if at some point (when ready) you'd switch to a npm build. This helps me versioning and deciding which frameworks are outdated.

PR #1854, opened 2025-03-27
feat: deku -> add

srghma:
@mikesol Could You help improve performance? Thanks!

Performance is bad bc of         `[ rowsPoll <#~> \rows -> D.tbody_ $ map renderRow rows`

I basically recreate table

and this is actually my first try of deku, and I wonder, should I do custom actions on every `create, ...etc`, or deku has some magic function that can do this for me
_____
very interesting problem: one source of truth, but changes can be inside and outside, looks like problem for vdom, but we dont have it

interesting how You will solve it, will You add vdom to deku?)

____

store signal with data? https://github.com/krausest/js-framework-benchmark/blob/9d6a759dd959a2a377b51f1a95adfdcfc59fa89f/frameworks/keyed/solid/src/main.jsx#L18

but how to do? https://github.com/krausest/js-framework-benchmark/blob/9d6a759dd959a2a377b51f1a95adfdcfc59fa89f/frameworks/keyed/solid/src/main.jsx#L42

### DIFF
--- a/frameworks/non-keyed/deku/.gitignore
+++ b/frameworks/non-keyed/deku/.gitignore
@@ -1,0 +1,12 @@
+# Dependencies
+.psci_modules
+.spago
+bower_components
+node_modules
+
+# Generated files
+.psci
+.psc*
+output
+output-es
+
--- a/frameworks/non-keyed/deku/index.html
+++ b/frameworks/non-keyed/deku/index.html
@@ -1,0 +1,11 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Deku v0.12.1</title>
+    <link href="/css/currentStyle.css" rel="stylesheet" />
+  </head>
+  <body>
+    <script src="output-es/bundle.js"></script>
+  </body>
+</html>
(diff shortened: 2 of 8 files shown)


srghma:
reflex is using PatchIntMap https://github.com/reflex-frp/reflex-dom/blob/4084bed6a05a996380c398cd427b11b780ea07f0/reflex-dom/benchmarks/krausest.hs#L32

https://github.com/reflex-frp/patch/blob/a8c3c3ae0c35215dc04271c38af85b6b65008398/src/Data/Patch/IntMap.hs#L37

krausest:
I'm getting the following error when I'm building:
```
> js-framework-benchmark-non-keyed-deku@1.0.0 postinstall
> spago install

[error] Failed to read the config. Error was:
[error] There's no "spago.dhall" in your current location.

If you already have a spago project you might be in the wrong subdirectory,
otherwise you might want to run `spago init` to initialize a new project.
npm error code 1
```
Would it be easier to submit the compiled files?

srghma:
updated, sorry

krausest:
One more thing to fix. The `npm run check` reports the following:
```
Expected to find 'aria-hidden'=true on span in third td, but found null
```
Can you check that please?

srghma:
added, tnx

krausest:
That worked. It's now available in the results:
I'm afraid to say but it performs pretty poorly.

PR #1856, opened 2025-03-31
Update sprae to v11.1.1

dy:
It's expected to be slightly smaller & faster.


krausest:

PR #1859, opened 2025-04-10
chore: remove useless import of ngFor

nigrosimone:
ngFor import is useless becouse template use @for


krausest:
Here's the comparison (updated keyed version left, old version right).
Screenshot 2025-04-16 at 10 29 28
No change (which is good in this case).

PR #1861, opened 2025-04-12
feat: nanoviews library

dangreen:


krausest:
Here's the comparison (updated keyed version left, old version right for other implementations).
Screenshot 2025-04-16 at 10 29 28
Not bad!

dangreen:
Thanks)

I ran these tests locally, and sometimes "select row" is slow (yellow), I don't know why, it shouldn't be so slow. Sometimes Sometimes I got good results:

results for my local runs:
local run

PR #1865, opened 2025-04-16
refactor: Change angular-cf-nozone to do the same DOM reconciliation …

atscott:
…as other Angular benchmarks

This benchmark should, at the very least, be calling `ApplicationRef.tick` instead of `ChangeDetectorRef.detectChanges`. Every other Angular benchmark app with automatically scheduled state synchronization goes through `ApplicationRef.tick` (which eventually enters the same code path as `ChangeDetectorRef.detectChanges`). This ensures we're at least measuring the same set of operations.

In reality, many "nozone" applications that are not using the new supported zoneless change detection are likely to be running `ApplicationRef.tick` with some interval at the top of the application rather than calling `detectChanges` or `tick` directly inside an event handler.

### DIFF
--- a/frameworks/keyed/angular-cf-nozone/src/app/app.component.ts
+++ b/frameworks/keyed/angular-cf-nozone/src/app/app.component.ts
@@ -1,5 +1,4 @@
-import { NgFor } from '@angular/common';
-import { ChangeDetectorRef, Component, VERSION, inject } from '@angular/core';
+import { ApplicationRef, Component, VERSION, inject } from '@angular/core';
 
 interface Data {
     id: number;
@@ -12,11 +11,10 @@ const nouns = ["table", "chair", "house", "bbq", "desk", "car", "pony", "cookie"
 
 @Component({
     selector: 'app-root',
-    imports: [NgFor],
     templateUrl: './app.component.html'
 })
 export class AppComponent {
-    private cdr = inject(ChangeDetectorRef);
+    private applicationRef = inject(ApplicationRef);
 
     data: Array<Data> = [];
     selected?: number = undefined;
@@ -44,7 +42,7 @@ export class AppComponent {
     select(item: Data, event: Event) {
         event.preventDefault();
         this.selected = item.id;
-        this.cdr.detectChanges();
+        this.applicationRef.tick();
     }
 
     delete(item: Data, event: Event) {
@@ -55,34 +53,34 @@ export class AppComponent {
                 break;
             }
         }
-        this.cdr.detectChanges();
+        this.applicationRef.tick();
     }
 
     run() {
         this.data = this.buildData();
-        this.cdr.detectChanges();
+        this.applicationRef.tick();
     }
 
     add() {
         this.data = this.data.concat(this.buildData(1000));
-        this.cdr.detectChanges();
+        this.applicationRef.tick();
     }
 
     update() {
         for (let i = 0; i < this.data.length; i += 10) {
             this.data[i].label += ' !!!';
         }
-        this.cdr.detectChanges();
+        this.applicationRef.tick();
     }
     runLots() {
         this.data = this.buildData(10000);
         this.selected = undefined;
-        this.cdr.detectChanges();
+        this.applicationRef.tick();
     }
     clear() {
         this.data = [];
         this.selected = undefined;
-        this.cdr.detectChanges();
+        this.applicationRef.tick();
     }
     swapRows() {
         if (this.data.length > 998) {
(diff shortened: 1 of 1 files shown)


krausest:
Here's a comparison (left new keyed, right old version):
Screenshot 2025-04-20 at 09 29 31

But to be honest I don't understand the change. AFAIK ChangeDetectorRef is/was the best performing way to update just one component in performance critical code and this angular implementation tries to squeeze everything out (Reference: https://medium.com/@kamil.galek/mythical-angular-angular-myth-there-is-no-difference-between-applicationref-tick-cde08a3f289a). According to that ApplicationRef should be slower. 
In the comparison it's a bit slower, but close to the noise level.

Would be nice if you could explain a bit more your change.

atscott:
The TL;DR is that the blog post is not wrong. Technically `ChangeDetectorRef.detectChanges` is a smaller operation than `ApplicationRef.tick`.

Though there's only 1 component in this app so the first half about checking parent components isn't really relevant. The second half is outdated and `ApplicationRef` calls the same function as `detectChanges` immediately - there's no `RootViewRef`. 

In general, I think this would fall into the following note from the readme: 

> Please don't over-optimize. This benchmark is most useful if you apply an idiomatic style for the framework you're using.

There are other things that `ApplicationRef.tick` does as well, such as ensuring render hooks and effects in services are run. Most frameworks don't expose APIs like this where you can force a partial update of the application just because it's faster and you maybe don't technically need the rest (add big red flashing caution sign here). It's definitely not always correct to do a partial update and `CDR.detectChanges` is a big trap, not only because of the incomplete updates, but also for refreshing the same view multiple times (more on this later).

Separately, I think it's probably valuable to have the state reconciliation operation be the same in each of the Angular applications in this repo. When Angular schedules and runs change detection for you, this always goes through `ApplicationRef.tick`. This came out of an investigation into why the benchmark numbers between "angular-cf-new-nozone"  and "angular-cf-nozone" show that the latter is "faster". Prior to this change, there were really only 2 differences between the applications. 

The first difference is the use of `ChangeDetectorRef.detectChanges` instead of `ApplicationRef.tick`. This should really only be a _very_ small difference and potentially only noise, as you noted. The "additional" work in `tick` is effectively a bunch of no-ops in this app. I think it's worth eliminating this difference, especially since we would strongly caution against using `detectChanges` in a real application.

The second difference is the scheduling mechanism. The "old" nozone application synchronously triggers an update while the "new" one schedules the update. Most major frameworks batch updates ([Vue](https://vueframework.com/guide/optimizations.html#virtual-dom), [React](https://github.com/reactwg/react-18/discussions/21), [Qwik](https://qwik.dev/docs/components/rendering/#render-batching), [Svelte](https://svelte.dev/docs/svelte/lifecycle-hooks#tick), and [Angular](https://github.com/angular/angular/blob/037dede0a0e645f817c285f958128f18a3ae3424/packages/core/src/change_detection/scheduling/zoneless_scheduling_impl.ts#L218)). [Solid](https://www.solidjs.com/tutorial/reactivity_batch) is a bit of an outlier due to its very granular rendering. From above, this is another reason a direct or synchronous call to `detectChanges` or `tick` isn't recommended. In reality, an old-style zoneless app should likely have some scheduling mechanism to batch updates. 

Most frameworks use `Promise.resolve()` for the batching while Angular races `requestAnimationFrame` and `setTimeout(0)` (Qwik also uses `setTimeout`). This is the bigger reason for the difference and does show up in the benchmarks. In my testing, `rAF` and `setTimeout` average about 4-6ms delay before executing in the benchmark runs compared to a fraction of a millisecond for `Promise.resolve()`. This number accounts for all the differences in timing between "angular-cf-nozone" and "angular-cf-new-nozone". 

---

In the end, I think it may come down to what exactly the benchmarks _should_ be including in the timings. Should the change go further and implement a batching mechanism for this app, potentially using `Promise.resolve()`? Should the other benchmarks instead all synchronously call `detectChanges` or `tick` to avoid potential penalization for the scheduling choice? Should they all use `detectChanges` since you don't technically _need_ whole-application `tick` to update the required DOM for these tests?

atscott:
Note on the 4-6ms in the benchmarks for the delay of `setTimeout`: I think this could be due to the warmups in the tests and Chrome then throttling the `setTimeout`. If I add the [MessageChannel trick](https://github.com/jasmine/jasmine/blob/e5d46e86240239615d4c6a7060a33973fad0928f/src/core/Clock.js#L222-L232), the 4-6ms hit in the benchmark goes away.

krausest:
Thanks a lot for your explanation!
I‘ll have think about the timeout delay. This benchmark always measured wall clock duration, so on the one hand it‘s okay to include it. If on the other hand warming up causes the timeout delay then we‘re doing something wrong. 
Until your post I thought that RAF based frameworks might have a slight disadvantage (waiting on avg half a frame).
Not sure how to solve it though.

PR #1872, opened 2025-04-25
perf: improve buildData by preallocating array

nigrosimone:


krausest:
Hi Nigro, I'm not going to merge it as it doesn't have any measurable impact on the results (optimized is your version with id++ inlined and the preallocated array);
Screenshot 2025-05-10 at 20 29 30
Neither create rows (1k and 10k) nor update rows were not faster. Those are the benchmarks which should be impacted most by the change.

PR #1874, opened 2025-04-26
Performance improvement

nihil-pro:
Screenshot 2025-04-26 at 22 52 42


krausest:
Here's the comparison (left new keyed, right previous old):
Screenshot 2025-05-10 at 19 59 19
I'm afraid if anything I can only see a small improvement.

PR #1876, opened 2025-05-01
DooHTML - v0.98.2-april-31-2025 - several minor improvements

hman61:


krausest:
Here's the comparison (left new keyed, right previous old):
Screenshot 2025-05-10 at 19 59 19
Not a big change, but also not much a chance for an improvement 😄

PR #1880, opened 2025-05-15
Benchmark Butterfloat

WorldMaker:
[Butterfloat](https://worldmaker.net/butterfloat/) is a small "framework" that tries to be a "Modern Knockout" (more of a "view engine" than a "framework"). Uses TSX but does not take a Virtual DOM approach. Lone dependency is RxJS.

Butterfloat changes very slowly (and deliberately) at this point, so I'm not sure if it will survive the "one-year" rule, but I'm curious to see how well it does in the benchmark in at least a year.

### DIFF
--- a/frameworks/keyed/butterfloat/.gitignore
+++ b/frameworks/keyed/butterfloat/.gitignore
@@ -1,0 +1,2 @@
+**/*.js
+index.html
--- a/frameworks/keyed/butterfloat/.prettierrc.json
+++ b/frameworks/keyed/butterfloat/.prettierrc.json
@@ -1,0 +1,5 @@
+{
+  "semi": false,
+  "singleQuote": true,
+  "endOfLine": "auto"
+}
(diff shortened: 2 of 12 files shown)


krausest:
I made some changes to make it fully compliant. The build process seems not very optimal. As a workaround I added the js and html files and removed it from gitignore. It would be better to have a build that stores all generated files into a dist folder.
But what's even more - it's pretty slow. I'm afraid to say it's even that slow that I consider removing it again since it considerably slows down the runs. Can you please check if I'm observing the expected performance?

Here's a screenshot, but please be aware that the updated and reference implementations were aldredy running on chrome 137:
Screenshot 2025-05-25 at 20 40 57
The next full run might happen later this week.

krausest:
I moved butterfloat into the broken-frameworks folder (which isn't correctly named in this case).
I don't think butterfloat currently benefits from being published in this benchmark.
The reported performance is way worse than it actually is, due to the principles how this benchmarks measures duration and the odd behavior of butterfloat preforming lots of RAFs with redraws.

PR #1881, opened 2025-05-16
doohtml-lite-v0.98.2-may-16-2025

hman61:
improved update method


krausest:
Here's a screenshot, but please be aware that the updated and reference implementations were aldredy running on chrome 137:
Screenshot 2025-05-25 at 20 40 57
The next full run might happen later this week.

PR #1884, opened 2025-06-04
Add React Native Onyx benchmark

blazejkustra:
Added a benchmark for keyed/react-native-onyx.

https://github.com/Expensify/react-native-onyx

Thank you!


krausest:
Thanks, here's the comparison (updated keyed version left, old version right which is not relevant in this case):
I'm afraid to say that performance doesn't look that good yet.

blazejkustra:
Thank you! That's why I added it as we are trying to improve the performance of this library 😄

PR #1887, opened 2025-06-06
doohtml-lite-v0.98.2-june-6-2025

hman61:
Changed the partial update method


krausest:
I'm afraid the change is to small to judge whether it's an improvement or noise.

PR #1889, opened 2025-06-11
endr@0.1.28

caseywebdev:
This version removes some needless object creation which slightly improves perf/memory usage.


krausest:
I'm afraid the change is to small to judge whether it's an improvement or noise.

caseywebdev:
Yea that makes sense, I think the bigger win is a little less memory usage in one of the busiest functions. Thanks for the update!

PR #1890, opened 2025-06-13
update pota

titoBouzout:
Ive restructured the project and added typings. It defaults to attributes instead of properties to align better with html. 

Of relevance, it uses `usePrevious` instead of `useSelector` which performs better for the case at hand. And have added some `data-` attributes, I'm not sure if this will cause any problems. 

Im unsure why the whitespace changed, its supposed to follow the prettier config from the package.json. This link is better https://github.com/krausest/js-framework-benchmark/pull/1890/files?diff=unified&w=1

Thanks!

### DIFF
--- a/frameworks/keyed/pota/package.json
+++ b/frameworks/keyed/pota/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-framework-benchmark-pota",
-  "version": "1.0.7",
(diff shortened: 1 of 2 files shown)


krausest:
Looks like a small improvement (but there's not much to shave off).

PR #1891, opened 2025-06-21
Update imports

nihil-pro:


krausest:
Thanks. Here's a comparison:
Screenshot 2025-07-03 at 20 12 58
I actually ran it twice to check and both runs gave the same results.
Preact became (quite certain) faster, while react seems a little bit slower or unchanged (vanillajs was also 0.02 slower in this run).

It would be good if future PRs didn't contain updates to the result files.

nihil-pro:
Yes, sorry about that. My mistake. 
Actually I only separate bindings for react/preact into separate sub packages, there shouldn’t be difference between 2.0.0 versions for react, and maybe that’s why preact is a little bit faster.

PR #1896, opened 2025-06-27
Bump michijs and vanillajs signals

lsegurado:
- Bump michijs
- Bump vanilla signals
- Now vanilla signals uses esbuild


krausest:
Thanks. Quite safe not slower, maybe even a slight bit faster:
Screenshot 2025-07-03 at 19 22 17

PR #1901, opened 2025-07-07
Allow "unflagged" selection when clicking "Which frameworks?"

twharmon:
Added an "Unflagged" button to the "Which frameworks?" modal, alongside "All" and "None". It filters out frameworks with issues/notes, showing only unflagged ones.


leeoniya:
nice, i've been wanting this :)

PR #1902, opened 2025-07-18
Update thyn

twharmon:
Should reduce memory usage by 15% or more. Reduces bundle size slightly. Might improve duration results as well.


krausest:
Thanks you (also for removing the other two implementations). Results are very close so it's hard to say whether it's noise or an actual improvement.
Screenshot 2025-07-21 at 21 41 18
Memory usage decreased significantly as you said.

PR #1903, opened 2025-07-21
Update spheres to v0.24

brian-watkins:


krausest:
Thanks you. Results are very close so it's hard to say whether it's noise or an actual improvement.
Screenshot 2025-07-21 at 21 41 18

PR #1906, opened 2025-07-24
Fix Crank to use non-deprecated key prop

brainkim:
The usage of `crank-key` is probably causing regressions because of console deprecation warnings. 0.7 will be released soon and I’d love to have an accurate comparison.

Thanks for this project!


krausest:
A lot faster!

PR #1907, opened 2025-07-27
Endr@0.2.0

caseywebdev:


krausest:
Somehow I missed updating endr. So I ran it immediately after noticing my fault.
Here are the results:
Screenshot 2025-08-03 at 19 50 54

PR #1908, opened 2025-07-30
update pota

titoBouzout:
I've dealt with some pending stuff, as refactoring the transform and getting rid of the shore placeholders. Tested the benchmark but really didn't test many other cases, I expect an improvement but hope nothing else is broken.

Thanks!


krausest:
Yes, your expectation was right.

PR #1910, opened 2025-07-31
Update to latest hydro-js version

Krutsch:
Performance and size will be almost the same but added some features and fixed bugs since 1.5.14


krausest:
Looks like you suggested:

PR #1911, opened 2025-08-04
Update pota

titoBouzout:
I've removed the `postinstall` script that compiles the babel transform from `es` to `cjs`, and instead NPM already host the compiled version. Hoping this doesn't cause any problem. Besides, there are some minor regression fixes and improvements.

Thanks!


krausest:
(Results will be posted here soon)

krausest:
Here's a screenshot of the incremental run (left updated versions, right previous versions):
Screenshot 2025-08-05 at 22 03 09
Too close to say if it's different.

PR #1912, opened 2025-08-05
uhtml 5

WebReflection:
This MR would like to update *uhtml* to its latest/greatest version **5** which includes:

  * the `key` attribute, *JSX* style, for keyed nodes in lists
  * the new *React* / *Preact* like **components**
  * the **signals** based update

This also:

  * removed the event delegation, it uses some tiny utility but that's actually how I would solve this "problem" or benchmark for real ...
  * the code is 100% ideomatic and there to showcase here and there features and potentials of the library ... I don't know if 772 flag is still needed but I don't care leaving it in
  * dependencies to create prod have been reduced and updated, currently there is no *npm* warning whatsoever, zero security issues/concerns + it's faster to bootstrap and test
  * the keyed and non-keyed folders share everything but the *table* which is indeed keyed

Hopefully I haven't forgotten anything else in here, looking forward to see this benchmarked.

The library now is more toward 6K and has everything included, included keys, reactivity via signals, new/different logic so I am not expecting it to be faster but surely as test-case/demo is way more elegant and easy to reason about.

Thank you.


krausest:
Thanks. I tried to rebuild, but the pages didn't load. They looked like that:
Screenshot 2025-08-05 at 20 46 08

After finding not much difference between keyed and non-keyed I decided to copy jumbotron and utils to the keyed directory. It worked then and I'd like to keep it that way.

krausest:
(Results will be posted here soon)

krausest:
Here's a screenshot of the incremental run (left updated versions, right previous versions):
Screenshot 2025-08-05 at 22 03 09
I'm glad to hear it's an ideomatic implementation. Nevertheless a bit slower than the old version.

I think the following code still justifies #772:
https://github.com/krausest/js-framework-benchmark/blob/master/frameworks/keyed/uhtml/src/utils.js#L33-L39

WebReflection:
@krausest thanks, too bad I've published `5.0.5` in the meantime and tried to force-push my MR but I guess that was late ...

> After finding not much difference between keyed and non-keyed I decided to copy jumbotron and utils to the keyed directory. It worked then and I'd like to keep it that way.

the whole point was to show that everything is the same *except* for the `key` attribute in the *Table* component, but I don't mind duplicated code as long as it does show the same happens indeed beside that attribute.

> I'm glad to hear it's an ideomatic implementation. Nevertheless a bit slower than the old version.

heh, changing the whole logic behind which removed native DOM parsing (enabling more features) and somehow simplified the diffing logic by dropping entirely the previous *unroll* orchestration, plus the fact it's now *signals* driven, and not just events with 772 and 801 all over the place as it was before, I couldn't expect better results **but** the first paint metric *should* be faster than before, which is at the end all that matters for fast/reactive libraries out there, curious to see next updated table where would position *uhtml* 5 but so far I am extremely happy by its improved DX and its "*batteries included*" approach, with signals and components, which also added extra logic to handle.

I guess one cannot have the cake and eat it too in here, yet the cake is much better than before in v5.

> I think the following code still justifies ...

yes, but also I've removed a single delegated event in keyed (one op) with two events per each row of the list, multiplying that operation up to 20K times -1 on each update, so again, this time I wanted to show *ideomatic code* as opposite of showing just raw perfs if things are "*cheating*" around, hence still happy it landed and it shows what it can do (hopefully still in a competitive way, otherwise I'll be back to the board thinking what to improve, which I also know, but it would make the final size a bit too big and maintainability a bit too harder than it needs to: that is avoiding 2 loops to resolve holes per each update, it could be one, it needs a deeper change though and I am not sure I want it to happen 😅)

WebReflection:
If you don't mind, I'll file another MR just updating the version after some detail will be nailed ... right now I think this position *uhtml* behind *lit* in the keyed table (not sure about non-keyed) and that's something I do care about (it's the only real competitor, *pota* requires tools while *uhtml* or *lit* doesn't).

krausest:
Feel free to submit a PR. I'm experimenting with some AI assisted merging. I think he's really eager and enjoys it so much more than me :) https://github.com/krausest/js-framework-benchmark/blob/master/.github/copilot-instructions.md

PR #1914, opened 2025-08-15
feat: add native-document implementation for benchmark

afrocodeur:


krausest:

krausest:
Seems like append rows is the bottleneck...

PR #1915, opened 2025-08-15
Update Hellajs to 0.14

omilli:
Added JSX support. Should be smaller, more memory efficient and (hopefully) faster.


krausest:

krausest:
I added a empty td since otherwise the isKeyed check rejects the html structure.

PR #1916, opened 2025-08-18
Update preact to v10.27.1

vasylenkoval:
Hey @krausest, big fan of this repo, thank you for all of your hard work!

I wanted to submit a small PR to bump all of the Preact implementations to `10.27.1` because the latest [release](https://github.com/preactjs/preact/releases/tag/10.27.1) has a small fix related to re-ordering of memoized rows.

It should remove majority of `swap-rows` cost in `preact-classes` and `preact-kr-observable`, since these are using memozied rows and were affected by the edge case that was fixed.

Other implementations, `preact-hooks` and `preact-signals`, do not use memoized rows and currently have faster swap performance (before this release), but end up having to eat the cost of not being able to skip processing rows in `select-row` and `partial-update`. 

Screenshot 2025-08-18 at 7 28 51 PM

I am guessing that the reason why hooks and signals implementations do not use memoized rows is because of the said swap deopt that was fixed. So I was also planning on submitting another PR to update `preact-hooks` and `preact-signals` to use memoization. Please let me know if you'd rather have it all in one PR, I can also push it here later.

### DIFF
--- a/frameworks/keyed/preact-classes/package-lock.json
+++ b/frameworks/keyed/preact-classes/package-lock.json
@@ -10,7 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@babel/plugin-transform-react-jsx": "7.25.9",
-        "preact": "10.25.2"
+        "preact": "^10.27.1"
       },
       "devDependencies": {
         "@babel/core": "7.26.0",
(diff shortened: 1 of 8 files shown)


krausest:

krausest:
For preact signals there's PR #1917 which I'll merge next. We first should take a look at those results...
Feel free to submit a PR experimenting with memoization...

PR #1918, opened 2025-08-25
Update keyed/mettle

maomincoding:
Updated version to v1.7.0!


krausest:
Since vanillajs was also slower it might be just noise, not really a performance regression.

Issue #1789, opened 2024-12-12
Update all implementations (at least the keyed ones)


krausest:
I'll try to update the implementations (starting with keyed, I've not yet decided if non-keyed are worth it) if there's a newer minor or major version (I'll skip patch versions) 

Implementations that fail to build or run will be moved to broken-frameworks and the last committer will be asked to fix the implementation.

krausest:
It's really annoying to do this by hand.
I changed the cli to support a command for automating the task:
`node cli.js update-frameworks --type keyed &> log.txt`
The script iterates over all frameworks, checks if there's a newer major or minor version for. a package specified in the js-framework-benchmark.frameworkVersionFromPackage property in the package.json (patch versions are ignored).
If a newer version is found we'll do a `ncu -u` and rebuild and check the implementation.
If that fails the implementation is restored via `git restore .``

What's currently missing (maybe I'l add that later):
* Automatically moving broken builds to the broken-frameworks folder
* Building broken builds after restoring (necessary if they are not moved. The build remains broken otherwise)
* Automatically creating and assigning an issue :)

krausest:
Here's the log (yeah with a spelling error...):

Sucessfully updated frameworks/keyed/alpine
Sucessfully updated frameworks/keyed/apprun
Retire keyed/art - js-art is older than a year
Retire keyed/dojo - @dojo/framework is older than a year
Retire keyed/ef-js - ef-core is older than a year
Failed to update frameworks/keyed/ember
Sucessfully updated frameworks/keyed/fntags
Failed to update frameworks/keyed/fre
Sucessfully updated frameworks/keyed/hono
Retire keyed/hyperapp - hyperapp is older than a year
Retire keyed/incremental-dom - incremental-dom is older than a year
Failed to update frameworks/keyed/ivi
Sucessfully updated frameworks/keyed/karyon
Retire keyed/knockout - knockout is older than a year
Failed to update frameworks/keyed/ko-jsx
Failed to update frameworks/keyed/legend-state
Sucessfully updated frameworks/keyed/lwc
Retire keyed/marionette - marionette is older than a year
Retire keyed/marionette-backbone - marionette is older than a year
Failed to update frameworks/keyed/marko
Failed to update frameworks/keyed/maverick
Error checking keyed/mikado-proxy
Failed to update frameworks/keyed/million
Failed to update frameworks/keyed/mobx-jsx
Sucessfully updated frameworks/keyed/owl
[...]

PR #1919, opened 2025-08-25
`karyon` update

syduki:
for the sake of #1789


krausest:
Thanks - here's the diff for the incremental run (left new keyed results, right old keyed results). Please note that vanillajs was a bit slower than in the last full run.
Since vanillajs was also slower it might be just noise, not really a performance regression.

PR #1921, opened 2025-08-28
Added non-keyed/vode framework

ryupold:
i hope the format fits.
there is no 'key' support for my framework, so i can only provide a non-keyed version.


krausest:
The results are now available in https://krausest.github.io/js-framework-benchmark/current.html
Since I'm skipping non-keyed results for odd chrome versions this has an odd effect on the results: Only updated frameworks (i.e. only vode) is displayed in the non-keyed table. You can compare against older chrome versions to get a feeling where it'd land or wait until chrome 140 where all non-keyed will be included ... or you could submit a keyed version 😄

PR #1924, opened 2025-09-09
`karyon` update

syduki:
This fixes the minor performance regression in `clear rows` observed in the [Chrome 140 results](https://krausest.github.io/js-framework-benchmark/2025/table_chrome_140.0.7339.81.html), with a slight improvement in `append rows`.

### DIFF
--- a/frameworks/keyed/karyon/package-lock.json
+++ b/frameworks/keyed/karyon/package-lock.json
@@ -8,16 +8,16 @@
       "name": "js-framework-benchmark-karyon",
       "version": "1.0.0",
       "dependencies": {
-        "karyon": "^4.0.0"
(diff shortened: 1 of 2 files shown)


krausest:
Here's a screenshot of an incremental run (left new keyed results, right old keyed results):
Screenshot 2025-09-13 at 21 45 14
Maybe a small improvement considering it's faster and vanillajs is a bit slower in this incremental run.

syduki:
Thanks.
"One small step for man, one giant leap for mankind" 😉, it may look small nevertheless that aligns with reality and intention.
Now `append` is faster than `replace` as it logically should be, actually ~15% improvement in that scenario which looks better than expected.

PR #1925, opened 2025-09-09
update non-keyed/vode

ryupold:
- fix version read-out
- memo rows


krausest:
Thanks. Results were (incrementally) updated: https://krausest.github.io/js-framework-benchmark/current.html

PR #1926, opened 2025-09-19
Update HellaJS to v1

omilli:


krausest:

krausest:
I had to fix the html structure (the empty td at the end of the row was missing).
And it looks like it doesn't run with a CSP. Maybe you want to check that.

NIce improvement for clear rows.

PR #1927, opened 2025-09-20
update ripple to 0.2.56

trueadm:


krausest:

krausest:
Same slowdown as for vanillajs (0.02), so I'd expect it to perform equally to the old version:

trueadm:
It should be quite a bit faster than the old version, at least in my testing, I was seeing `1.04` - `1.05`. I guess we can wait for the full run again at the next major version release – it's a shame Ripple missed out on that, as the entire reactivity syntax changed a few days after my last PR landed and that had a major impact on performance (we were essentially using Proxies before, and now we are not).

PR #1928, opened 2025-09-30
endr@0.2.1

caseywebdev:
Bugfixes, not expecting much perf impact


krausest:

krausest:
Here are the results (left new keyed, right previous keyed results):
vanillajs was 0.02 slower, so both versions should be pretty close.

PR #1932, opened 2025-10-09
update non-keyed/vode

ryupold:
maybe a little better performance, hopefully not worse :)


krausest:
Just for updated non-keyed frameworks the results are updated: https://krausest.github.io/js-framework-benchmark/current.html

PR #1933, opened 2025-10-12
Shorten leptos build_data to more idiomatic Rust

daniel-pfeiffer:


daniel-pfeiffer:
Please note that the 4th commit potentially changes the order of ids (explained within.) Feel free to drop that commit if it changes your test assumptions!

krausest:

PR #1935, opened 2025-10-15
update lui, adding noeval variant

L3P3:
I did not improve much but just updated to the latest lui version.
And to make the result slightly smaller, added minification. I minify with bun but in case you don't like my approach, just use some other simple minifier.
And to remove the CSP issue, added a non-eval variant.

**Very** surprisingly, it performs almost the same on my machine, throwing me into an existential crysis.
There must be some advanced pattern matching in v8 detecting object key iteration and transforming it into struct lookups.
I wonder how Firefox and Safari would perform.
Maybe in future, I will make the noeval variant default and remove the other benchmark.

Thanks.


krausest:

Almost indistinguishable. I'm a big fan of keeping only the ~noevil~ noeval version.

PR #1936, opened 2025-10-16
Bump sprae to v12

dy:


krausest:
Just for updated non-keyed frameworks the results are updated: https://krausest.github.io/js-framework-benchmark/current.html

Issue #1795, opened 2024-12-14
Updating keyed/marko failed


krausest:
As in #1789 I'm updating all implementations. Implementations that failed to update are moved in to the broken-frameworks folder.
Here's what I'm trying to update:
```
 @marko/run                  ^0.2.0  →   ^0.5.8
 @marko/run-adapter-static   ^0.1.5  →   ^0.2.1
 marko                      ^5.31.0  →  ^5.37.1
 ```
The build fails with the following error:
```
[marko-vite:pre] 
    at node_modules/.marko/60b37072/__marko-run__route._index.entry.marko:11:8
       9 | }
      10 | static const assets = ["__marko-run__route_95kF"];
    > 11 | <const/writeSync=addAssets($global, assets) || setFlush($global)/>
         |        ^^^^^^^^^ Tag does not support a variable.
      12 | -- $!{writeSync && getPrepend($global)}
      13 | <Template ...input/>
      14 | -- $!{writeSync && getAppend($global)}
file: /home/stefan/js-framework-benchmark/frameworks/keyed/marko/node_modules/.marko/60b37072/__marko-run__route._index.marko?marko-server-entry:undefined:undefined

   9 | }
  10 | static const assets = ["__marko-run__route_95kF"];
> 11 | <const/writeSync=addAssets($global, assets) || setFlush($global)/>
     |        ^^^^^^^^^ Tag does not support a variable.
  12 | -- $!{writeSync && getPrepend($global)}
  13 | <Template ...input/>
  14 | -- $!{writeSync && getAppend($global)}
```
@birkskyum Can you take a look at this?

PR #1937, opened 2025-10-17
Fix Marko 5 example, add Marko 6

kanashimia:
I decided to keep both, as they are basically different frameworks, named them as marko and marko-classes respectively.

This benchmark doesn't really test what Marko is actually good at - subcomponent hydration (resumability) and out-of-order streaming, but that kind of a client side performance is still a pretty cool achievement for a SSR-focused framework right? :)

image

Closes https://github.com/krausest/js-framework-benchmark/issues/1795


kanashimia:
readme has this "Please use fixed version numbers, no ranges, in package.json. Otherwise the build will break sooner or later - believe me. Updating works IMO best with npm-check-updates, which keeps the version format."

but a lot of frameworks here still seem to use semver ranges, what's up with that?

krausest:
I guess I'm not too good enforcing that policy...
To be honest: Since I stopped rebuilding the frameworks regularly I care less about it.

Nice results (left new keyed, right old keyed results). The incremental run was a bit slower for vanillajs, so subtracting 0.02 from the weighted mean might be a plausible results from a full run:

PR #1943, opened 2025-11-21
updated: Riot.js version

GianlucaGuarini:
Update Riot.js to v10
🙏


krausest:
Thanks. Here's a screenshot of the incremental results (left new keyed results, right old keyed results):
image
That incremental run wasn't particularly fast. I'd expect it to perform almost like the old version.

PR #1944, opened 2025-12-01
Remove unneeded memo wrapper in the react-compiler-hooks

MrFlashAccount:
This PR removes the unnecessary memo wrapper, that adds overhead by doubling the comparison of the props:
1. Inside the memo func
2. Inside the React component

Note: rerendering of the individual rows is fine - if nothing changed, React will immediately return with the cached result

### DIFF
--- a/frameworks/keyed/react-compiler-hooks/src/main.jsx
+++ b/frameworks/keyed/react-compiler-hooks/src/main.jsx
@@ -1,4 +1,4 @@
-import { useState, memo } from "react";
+import { useState } from "react";
 import { createRoot } from "react-dom/client";
 import { buildData } from "./utils";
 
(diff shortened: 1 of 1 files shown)


krausest:
Sorry, seems like select row is hit badly by your PR... (left new results, right old results). The other benchmarks are slightly faster, but select is significantly worse.

I suggest rolling back the change. Ok?

Screenshot 2025-12-15 at 07 39 51

MrFlashAccount:
@krausest that's strange. At least on my pc, the difference is negligible. If you don't mind, I'd like to ping the react core team according the need to use manual `memo` for list items.

UPD: at least in docs, if you already using the compiler, you don't need any extra memoization on top of it (except a few edge-cases): 

https://react.dev/learn/react-compiler/introduction#after-react-compiler

UPD2: reproduced on my machine with a very large slowdown.

MrFlashAccount:
@krausest waiting for a confirmation from the core team. I think, in case the memo here is not needed with the compiler (as per docs), we can ignore the perf degradation, as it's intended. Thoughts?

PR #1948, opened 2025-12-14
chore: update all angular to latest version (21)

nigrosimone:
update all angular to latest version (21):
```
frameworks\keyed\angular-*
```


krausest:
Thanks, very welcome. Here are the results (left new keyed, right old results):
Screenshot 2025-12-15 at 07 39 51

I'd say overall slightly slower, new no-zone is still worse than the old zoneless implementation. At least signals are as fast as the classic approach.

The angular-cf-signals-optimized is angular-cf-signals with #1949 applied. I'm very happy to say that those micro-optimizations get lost in the measurement noise for this benchmark.

PR #1989, opened 2026-03-28
Add keyed Lattice implementation

xlreon:
## What is Lattice?

[Lattice](https://github.com/nicoobc/lattice) is a compiled UI framework with a Rust-based compiler that transforms `.lat` templates into optimized vanilla JavaScript at build time.

**Key characteristics:**
- **No virtual DOM** — the compiler generates direct DOM operations
- **Zero runtime overhead** — compiled output is plain JS with no framework runtime
- **Template cloning** — compiler emits `cloneNode` patterns instead of `createElement` chains
- **Compiler-generated event delegation** — not manually written, emitted by the compiler

Since Lattice compiles away at build time (similar to Svelte's approach), the benchmark entry is the pre-compiled output. The Rust compiler is not yet published to npm, so `build-prod` is a no-op — the compiled JS is checked in directly.

## Structure

```
frameworks/keyed/lattice/
├── index.html           # Standard benchmark HTML
├── package.json         # js-framework-benchmark config
├── package-lock.json    # Minimal (no runtime deps)
└── src/
    └── Main.js          # Compiled output from lattice-rs
```

## Notes

- **Issues:** Self-declaring [#772](https://github.com/krausest/js-framework-benchmark/issues/772) — the compiled output uses direct DOM class manipulation for selection, which is the idiomatic pattern for a compiler-emitted framework
- **Keyed:** Implementation maintains a 1:1 mapping between data items and DOM nodes via `data_id` stored on each `<tr>`
- **No dependencies:** Zero npm dependencies — the compiled output is self-contained
- **Version:** 0.1.0 (pre-release, hardcoded since the compiler isn't on npm yet)

## Checklist

- [x] `npm ci` succeeds (no dependencies to install)
- [x] `npm run build-prod` succeeds (exit 0 — compiled output checked in)
- [x] All 7 benchmark operations implemented (create 1K, create 10K, append, update, clear, swap, select/remove)
- [x] Uses `/css/currentStyle.css` from root
- [x] Correct button IDs (`run`, `runlots`, `add`, `update`, `clear`, `swaprows`)
- [x] Preload icon included (`<span class="preloadicon glyphicon glyphicon-remove" aria-hidden="true"></span>`)
- [x] Fixed version numbers in package.json
- [x] No gzipped files in output
- [x] Only files in `frameworks/keyed/lattice/` are modified

### DIFF
--- a/frameworks/keyed/lattice/index.html
+++ b/frameworks/keyed/lattice/index.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
+    <title>Lattice-keyed</title>
+    <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+<div id='main'>
+    <div class="container">
+        <div class="jumbotron">
+            <div class="row">
+                <div class="col-md-6">
+                    <h1>Lattice-keyed</h1>
+                </div>
+                <div class="col-md-6">
+                    <div class="row">
+                        <div class="col-sm-6 smallpad">
+                            <button type='button' class='btn btn-primary btn-block' id='run'>Create 1,000 rows</button>
+                        </div>
+                        <div class="col-sm-6 smallpad">
+                            <button type='button' class='btn btn-primary btn-block' id='runlots'>Create 10,000 rows</button>
+                        </div>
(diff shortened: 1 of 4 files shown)


tarsa:
https://github.com/nicoobc/lattice returns http 404 currently

PR #2000, opened 2026-04-03
Update Ember framework version to 6.12.0

NullVoxPopuli:
oops 🙈


krausest:
Thanks - here are the results. Please note that updated frameworks were run with chrome 147 while the rest of the table (those hidden in the screenshot) was run with chrome 146. A full run will happen for chrome 148.
image

PR #2001, opened 2026-04-04
Add roqa implementation

hawkticehurst:
Submitting implementation for [Roqa](https://github.com/roqajs/roqa/). Let me know if you need any specific info from me to help with merging.

Thanks again @krausest!

### DIFF
--- a/frameworks/keyed/roqa/.gitignore
+++ b/frameworks/keyed/roqa/.gitignore
@@ -1,0 +1,24 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+lerna-debug.log*
+
+node_modules
+dist
+dist-ssr
+*.local
+
+# Editor directories and files
+.vscode/*
+!.vscode/extensions.json
+.idea
+.DS_Store
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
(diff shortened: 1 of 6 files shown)


krausest:
Pretty impressive - are there any bad tricks I didn't notice 😄 ?
Please note that updated frameworks were run with chrome 147 while the rest of the table (those hidden in the screenshot) was run with chrome 146. A full run will happen for chrome 148.
image

hawkticehurst:
Nice! 🥳 Also everything _should_ be above board! 

The only "trick" that I expect will be controversial (but is, to my knowledge, legal for this benchmark) is that Roqa caches DOM references similar to what Mikado does. The big difference is that Roqa does this as a compiler optimization rather than an explicit developer-facing API. So, I'm hoping to argue that what Roqa does is "better" or feels less "cheat-y" than what Mikado does because it lives in an interesting grey zone where the compiler is deriving an optimization from clean declarative code rather than asking developers to hand-author these optimizations.

Beyond that, Roqa implements all the "typical" strategies that frameworks like Solid, Svelte, Ripple use to achieve performance. Perhaps the only other big / unique optimization is that Roqa _aggressively_ inlines code / code paths. For example, this declarative code...

```jsx
const count = cell(0);
console.log(get(count));
set(count, get(count) + 1);
```

...will be compiled down to plain lightweight JS objects, with direct (very fast) object property access:

```js
const count = { v: 0, e: [] }
console.log(count.v);
count.v = count.v + 1;
```

You could maybe think of it as Roqa effectively compiling away its reactivity system; so it's achieving a lot of perf by doing less than other frameworks at runtime.

PR #2003, opened 2026-04-10
Update Qingkuai implementation and dependencies

mlgq:
This PR updates the Qingkuai keyed and non-keyed implementations with the following changes:
- Bump `qingkuai` from `^1.0.55` to `^1.0.61`
- General performance optimizations:
  - Optimize data generation to reduce unnecessary allocations and per-iteration overhead
  - Simplify row append, remove, update, and selection logic
  - Switch selection handling to state-driven updates instead of imperative DOM mutations
- Remove note `#772` (previously mislabeled as `#801`) by eliminating manual DOM manipulation from the implementation:
  - Row selection is now handled declaratively through framework state and template bindings
  - This better reflects how the framework works rather than relying on custom DOM patching

### DIFF
--- a/frameworks/keyed/qingkuai/package-lock.json
+++ b/frameworks/keyed/qingkuai/package-lock.json
@@ -6,7 +6,7 @@
         "": {
             "name": "js-framework-benchmark-qingkuai",
             "dependencies": {
-                "qingkuai": "^1.0.55",
+                "qingkuai": "^1.0.60",
                 "vite": "^8.0.3",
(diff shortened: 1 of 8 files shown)


krausest:
Nice! Here are the results. Please note that updated frameworks were run with chrome 147 while the rest of the table (those hidden in the screenshot) was run with chrome 146. A full run will happen for chrome 148.
image

mlgq:
Thanks for merging the PR. Looking forward to the full benchmark results. Let me know if there's anything else you need from me.

PR #2006, opened 2026-04-27
AkashJS: rewrite using framework primitives

phpirate:
## Summary

- Rewrites the AkashJS benchmark to use proper framework primitives instead of manual DOM manipulation
- Replaces manual reconciler with `<For>` keyed list component (addresses note 772)
- Replaces event delegation on tbody with per-row `onClick` handlers (addresses note 801)
- Uses `.akash` single-file component format with `class:danger` directive and `signal()` state

## Changes

- `index.html` — simplified to a shell (buttons moved into the component)
- `src/main.ts` — reduced from 222 lines to 3 lines (imports and mounts the component)
- `src/App.akash` — new single-file component with all logic and template

## Build

- Bundle: 5.71 KB gzip
- Build time: ~200ms


krausest:
Just a quick info: It'll take a few more days until I can look at the PR...

krausest:
Thanks - all flags are gone 😄 

image

PR #2010, opened 2026-05-08
Update pota dependency version to ^0.20.233

titoBouzout:
Just a bump.

On running time it may have got a bit worse, memory probably improved but was already pretty good. 

Thanks!


krausest:
Just a quick info: It'll take a few more days until I can look at the PR...

krausest:
Thanks - results are updated...
image

PR #2015, opened 2026-05-19
Add plastron-dom v0.0.2 (keyed + non-keyed)

rheophile10:
Adds keyed and non-keyed framework dirs for **plastron-dom**, a DOM-rendering segment built on a reactive cel substrate. Repo: https://github.com/rheophile10/plastron (pinned at `66575305d030`).

## Build shape

Pre-built bundle (vanillajs pattern). plastron and plastron-dom aren't published to npm yet, so the framework dirs can't resolve source-time deps via the standard `npm install + npm run build-prod` flow. `main.js` is the production artifact produced by `vite build` on the scaffold at `bench/krausest/{keyed,non-keyed}/plastron-dom-v0.0.2/src/main.ts` in the plastron monorepo. `src/main.ts` is included verbatim from that scaffold for reviewer reference; its deep-relative imports (`../../../../../plastron/src/...`) point at the plastron monorepo and aren't resolvable inside this directory. `package.json`'s `build-prod` is `exit 0` — same shape as vanillajs.

## Validation

- `npm run isKeyed -- keyed/plastron-dom-v0.0.2`: classifies as **keyed** for run / remove-row / swap-rows.
- `npm run isKeyed -- non-keyed/plastron-dom-v0.0.2`: classifies as **non-keyed** for the same.
- `npm run rebuild-ci -- {keyed,non-keyed}/plastron-dom-v0.0.2`: both pass.

## CSP

Both variants declare `issues: [1139]` — the plastron-sheet formula compiler uses `new Function()` for codegen.

## Architectural notes

plastron-dom is one segment in a broader substrate (reactive DAG kernel + dehydratable cels + segment-based packaging). For krausest specifically the relevant detail is the diff strategy: positional fallback for unkeyed children, LIS-based minimal-edit reconciliation for keyed children. The `05_swap1k` benchmark exercises this directly — the keyed variant turns a 1000-row swap into 2 `insertBefore` calls.

Local 15-iteration medians (Chrome 148, Ubuntu 26.04) place the keyed variant in the React/Angular tier — geomean 1.91× vanilla, ahead of React on `05_swap1k` (3.2×) and `07_create10k` (1.3×), behind React on `04_select1k` (~4×) because the render tree is one lambda over the whole table rather than per-row components (architectural choice, documented in our cookbook).

Real numbers will come from this benchmark's CI machine; the above is sanity-check ballpark.


krausest:
BTW I removed the version from the directory names.

krausest:
Thanks - here's a screenshot from the [results](https://krausest.github.io/js-framework-benchmark/current.html). 
image

rheophile10:
thank you sir I am much obliged to you. i would like to try future versions eventually

PR #2016, opened 2026-05-19
Update LiteralJS to v8.0.5

Mikeysax:
This PR updates the LiteralJS benchmark implementations:

## Changes
- **New keyed implementation**: Added frameworks/keyed/literaljs/ with explicit key props on Row components for proper keyed reconciliation
- **Updated non-keyed implementation**: Migrated from v7.0.2 to v8.0.5 API
  - Replaced render(App, 'root', {}) with new App(App).mount('root')
  - Updated HTML structure (DOCTYPE, lang attribute)
  - Removed node-sass dependency (incompatible with Node 25)

## Bundle Sizes
- Keyed: ~17.02 KB
- Non-keyed: ~16.88 KB

Both implementations build successfully with Parcel.


krausest:
Thanks. 

1) I'm afraid the keyed implementation isn't right.

Keyed test for swap failed. Swap must add the TRs that it removed, but there were 2 new nodes
literaljs-v8.0.5-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and non-keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
ERROR: Framework literaljs-v8.0.5-keyed is not correctly categorized

Please fix the implementation and open a new PR. Please don't forget to run `npm run isKeyed keyed/literaljs` in webdriver-ts.

2) Please add the following key to js-framework-benchmark in the package.json to point the benchmark driver to the right directory.
    "customURL": "/dist"

PR #2017, opened 2026-05-21
feat: Add astro vanilla + astro react

Johannsis:
Screenshot 2026-05-20 at 11 26 23 PM


krausest:
Thanks. But I'm afraid I won't be able to keep running this benchmark if we add all possible permutations.
Can you please pick two combinations that are particularly noteworthy? Maybe astro and astro-react?

Johannsis:
> Thanks. But I'm afraid I won't be able to keep running this benchmark if we add all possible permutations. Can you please pick two combinations that are particularly noteworthy? Maybe astro and astro-react?

Left astro and astro-react!

krausest:
Here's a screenshot of the incremental run (left new keyed, right previous results - so not relevant here):
image

I'm excited to see how astro-react in the next full run competes with the other react implementations.

PR #2018, opened 2026-05-23
Add solarite non keyed

eric-frost:
I fixed the previous issue that was causing the 04_select1k validation to fail.  Sorry about that : )


krausest:
Thanks -  [results](https://krausest.github.io/js-framework-benchmark/current.html) are updated.
Any chance for a keyed version?

eric-frost:
I'll see what I can do, but it won't come quickly.  The architecture isn't immediately amenable to being keyed.  

Thanks for the accept : )

PR #2019, opened 2026-05-24
update ripple to latest version

trueadm:


krausest:
Thanks. The incremental run was a bit slower for vanillajs and ripple in comparison (on the left are the new results, on the right the previous results):
image

trueadm:
For me the results should be neutral to before, so likely just an off incremental run. :)

PR #2020, opened 2026-05-24
AkashJS: bump packages with perf fixes

phpirate:
## Summary

- Bump `@akashjs/runtime` to 0.2.8, `@akashjs/compiler` to 0.1.60, `@akashjs/vite-plugin` to 0.2.4
- LIS-based keyed list reconciler — swap/remove operations now produce minimal DOM moves instead of O(n) iteration
- `class:` directive caches previous boolean value, skips `classList.toggle` when unchanged
- Signal `.set()` uses inline batch instead of closure allocation

## Expected impact

| Benchmark | Before (Chrome 149) | Expected |
|-----------|-------------------|----------|
| swap rows | 82.7ms | ~12ms |
| select row | 6.4ms | ~2.5ms |
| remove row | 41.0ms | ~10ms |

### DIFF
--- a/frameworks/keyed/akashjs/package.json
+++ b/frameworks/keyed/akashjs/package.json
@@ -7,12 +7,12 @@
     "build": "vite build"
   },
   "dependencies": {
-    "@akashjs/runtime": "0.2.7",
-    "@akashjs/compiler": "0.1.59",
-    "@akashjs/vite-plugin": "0.2.3"
+    "@akashjs/compiler": "0.1.60",
+    "@akashjs/runtime": "0.2.8",
+    "@akashjs/vite-plugin": "0.2.4"
   },
   "devDependencies": {
-    "vite": "5.4.21",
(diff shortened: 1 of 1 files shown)


krausest:
Thanks. Your implementation lacks a few thing to allow running it. I had added them in the repo, but your PR undid those changes. I added them again - please make sure to **keep them for future PRs** as it's not particular funny to find out why the benchmark runner fails (and even fix a merge conflict).

The important pieces are: js-framework-benchmark section in package.json and it seems like vite.conf.ts needs a base: './' otherwise it'll try to load the bundled js from "/".

phpirate:
@krausest Thanks for the fixes and merge, and apologies for the extra overhead. I've noted the js-framework-benchmark requirements and base configuration for future PRs.

krausest:
Here's a screenshot of the incremental run (left new keyed, right previous results). Looks like a very nice improvement:
image

PR #2023, opened 2026-05-27
LiteralJS v8.0.5 keyed implementation

Mikeysax:
This PR fixes the issues reported in #2016:

## Fixes

1. **Keyed reconciliation on swap rows**: Added `key={item.id}` to the <tr> element inside the Row component. Previously the key was only passed to the Row component itself, but the benchmark test requires the key to be on the actual DOM element for proper keyed reconciliation during swap operations.

2. **Missing customURL in package.json**: Added `"customURL": "/dist"` to the keyed package.json to point the benchmark driver to the correct directory.

## Verification

Ran `npm run isKeyed keyed/literaljs` and all tests pass:
- run benchmark (keyed)
- remove row benchmark (keyed)  
- swap rows benchmark (keyed)

The implementation is now correctly categorized as keyed.


krausest:
Nice. Keyed implementations are always welcome.

Here's a screenshot of the incremental run (left new keyed, right previous results - not relevant here):
image

PR #2025, opened 2026-06-06
Update spheres to v0.27.2

brian-watkins:


krausest:
Thanks. No big change (left new keyed results from chrome 149 and on the right the previous results from chrome 148 - so take care with the comparison. Though vanillajs changed less than expected...)
image

PR #2026, opened 2026-06-07
feat(angular): upgrade to version 22

nigrosimone:


krausest:
Thanks, here are the result:
Left new keyed results from chrome 149 and on the right the previous results from chrome 148 - so take care with the comparison. Though vanillajs changed less than expected...
image

As things move IMO in the right direction and newer approaches are as fast or faster than the old ones: Can we consider which implementations to keep? 
ngfor is slower than cf (yeah!). Can we remove it?
Can we also remove cf-signals-nozone now that cf-new-nozone is as fast?

nigrosimone:
> which implementations to keep?

I would just keep the latest (angular 22) default implementation, with:
- Zoneless
- OnPush
- Control flow
- Signal

krausest:
Thanks. Removed in 3d59532b9955c5fc4dae8eac65c852433c309dff

PR #2027, opened 2026-06-08
AkashJS: bump runtime to 0.2.9

phpirate:
## Summary

- Bump `@akashjs/runtime` from 0.2.8 to 0.2.9
- Fix memory leak: effects and signal subscriptions in `<For>` are now disposed when rows are removed
- Reduce bundle size: 18.3 KB → 16.0 KB (tree-shaking improvements)

Only `package.json` is changed (runtime version). The `js-framework-benchmark` section and `vite.config.ts` with `base: './'` are preserved.


krausest:
image

PR #2030, opened 2026-06-09
Update LiteralJS to v8.1.8

Mikeysax:
Updates the LiteralJS benchmark implementation to v8.1.8.

Changes:
- Update non-keyed LiteralJS from 8.0.5 to 8.1.8
- Add keyed LiteralJS implementation

Following the pattern from #2028.


krausest:
The keyed version works, but the non-keyed doesn't. 
The non-keyed version looks OK, shows no error in the console, but clicking on the buttons doesn't have any impact.
Can you take a look at it?

Mikeysax:
@krausest Okay, thank you for pointing that out. I believe I fixed the issue.

krausest:
image

PR #2031, opened 2026-06-10
Add keyed Micra implementation

denisfl:
This adds a keyed implementation for [Micra.js](https://github.com/denisfl/micra.js) — a ~5 KB (gzip) reactive library for server-rendered pages: shallow-proxy state, data-* directives, keyed list diffing (LIS-based reordering), no build step.

**Implementation notes**:

- Fully idiomatic, no escape hatches: the table is one <template data-each="data" data-key="id">, all mutations replace state.data with a new array (the library's documented pattern — the proxy is intentionally shallow), selection is data-class="danger:item.id === selected".
- No build: build-prod is a no-op; the page loads node_modules/micra.js/dist/micra.min.js directly (same pattern as vanillajs-lite).
- Event handlers take bare method names (@click="select"); row identity is passed via a bound data-id attribute — Micra deliberately has no call-expression syntax in bindings.

Happy to adjust anything to match current contribution guidelines.

### DIFF
--- a/frameworks/keyed/micra/index.html
+++ b/frameworks/keyed/micra/index.html
@@ -1,0 +1,112 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Micra-keyed</title>
+    <link href="/css/currentStyle.css" rel="stylesheet" />
+  </head>
+  <body>
+    <div class="container" id="main" data-component="main">
+      <div class="jumbotron">
+        <div class="row">
+          <div class="col-md-6">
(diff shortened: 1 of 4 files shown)


krausest:
Thanks. Seems like it doesn't work with CSP, so I added #1139

denisfl:
Thanks. That's the correct categorization. Micra compiles non-trivial directive expressions via `new Function`, so it triggers a CSP eval violation under default-src 'self'. Simple property bindings work, but anything with operators/calls doesn't. A CSP-safe expression evaluator is on the roadmap. I hope I'll open a PR to drop the 1139 flag once it lands.

PR #2033, opened 2026-06-14
Solarite 0.6.0 which adds keyed/solarite

eric-frost:
I worked on improving Solarite performance and added a keyed version.  Happy to adjust anything that's needed.


krausest:
image

PR #2034, opened 2026-06-15
Bumps the pinned micra.js dependency 2.3.2 → 2.5.0

denisfl:
The #1139 strict-CSP note — Micra 2.5.0 replaced its new Function-based directive-expression compiler with a tokenizer + Pratt parser + AST interpreter, so 2.5.0 needs no unsafe-eval. 

I verified the keyed entry locally under the exact policy staticRouter applies (default-src 'self'; report-uri /csp), exercising create/update/swap/append/clear/select — zero CSP violations. Happy for you to drop #1139 if you agree.

### DIFF
--- a/frameworks/keyed/micra/package-lock.json
+++ b/frameworks/keyed/micra/package-lock.json
@@ -9,13 +9,13 @@
       "version": "1.0.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "micra.js": "2.3.2"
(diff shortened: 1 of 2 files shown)


krausest:
image

#1139 is gone and I added a generic mechanism to add some files to the build.zip:
https://github.com/krausest/js-framework-benchmark/blob/master/frameworks/keyed/micra/package.json#L9

PR #2035, opened 2026-06-15
Upgrade native document to 1.0.172

afrocodeur:
Upgrade native document 1.0.172

ElementCreator.processChildren and getChild - refactored for better performance
nd-element-extensions - all prototype extensions converted to non-enumerable defineProperty
NDElement events - inline handler (on*) used when no existing handler and no options; falls back to addEventListener for subsequent handlers or when options are provided


krausest:
image

PR #2038, opened 2026-06-18
Add Nuclo keyed benchmark scaffold with UI implementation

dan2dev:
This pull request adds a new implementation of the JS Framework Benchmark using the Nuclo framework. 

https://nuclo.dev/
https://github.com/dan2dev/nuclo


krausest:
image

PR #2041, opened 2026-06-29
Add kerfjs (keyed)

brianwestphal:
Adds a **keyed** implementation for [kerfjs](https://github.com/brianwestphal/kerf) — a ~12 KB (min+gzip) reactive UI framework: fine-grained signals (`@preact/signals-core`) + a keyed-list reconciler + a tiny JSX runtime. No build step beyond the consumer's existing one.

### Implementation notes
- **Keyed.** Rows live in kerf's `arraySignal`, so create / append / update / swap / remove drive the keyed-list reconciler in O(changes).
- **Selection** is tracked as a single `selectedId` signal (one id for the whole table — no per-row flag). The selected/​deselected rows are morphed in place.
- **Event handling** uses kerf's idiomatic `delegate()` (event delegation), so the *manual event delegation* note (#801) applies — declared in `package.json`.
- Deps pinned to fixed versions (`kerfjs@0.14.0`, `typescript`, `vite`) per the contribution guidelines.

Happy to adjust anything to fit the contribution requirements — thanks for maintaining this benchmark!

### DIFF
--- a/frameworks/keyed/kerfjs/index.html
+++ b/frameworks/keyed/kerfjs/index.html
@@ -1,0 +1,15 @@
+<!doctype html>
+<html>
+  <head>
+    <title>kerfjs-keyed</title>
+    <meta charset="UTF-8" />
+    <link
+      href="/css/currentStyle.css"
+      rel="stylesheet"
+    />
+  </head>
+  <body>
+    <div id="main"></div>
+    <script type="module" src="/src/main.tsx"></script>
+  </body>
+</html>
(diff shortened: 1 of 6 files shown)


krausest:
I'm afraid there's an error in the app. Row selection doesn't work once a row is deleted. Can you fix that please?

brianwestphal:
Fixed — thanks for catching that. It was a genuine reactivity bug in kerf: deleting a row applied a granular list patch that dropped the selection signal out of the render's dependency set, so subsequent selections stopped re-rendering. Released the fix in kerfjs 0.16.0 and bumped this entry's pin accordingly (also fixes an append-after-clear case). Verified locally that create → delete a row → select now highlights correctly.

PR #2042, opened 2026-06-30
Update knno-jsx dependencies to version 2.5.1 and adjust package-lock…

ThorQin:
### Changes
1. Upgrade all knno-jsx dependencies from previous version to 2.5.1
2. Regenerate package-lock.json to lock the new dependency versions

### Reason
Apply latest bug fixes & minor performance improvements.


krausest:
image

PR #2043, opened 2026-07-03
Add sprae 13.4.0

dy:
Adds sprae 13.4.0 in keyed and non-keyed.

Keyed uses object rows, which sprae keys by object identity. Non-keyed iterates primitive ids and keeps labels separately, so rows are reused positionally.

Checked locally:
`npm run rebuild-ci keyed/sprae non-keyed/sprae`


krausest:
Finally! This implementation worked nicely. Here are the results (keyed results - left new, right old for updated implementations):
image
Performs a little bit faster than react and angular.

PR #2045, opened 2026-07-05
Update to latest hydro-js version

Krutsch:
Traded file size for some perf and memory gains. Looks like this for me:
image


krausest:
Just as a reminder for me: Couldn't merge yet. The benchmark runner fails for two test cases - can't confirm with manual testing.

Executing frameworks/non-keyed/hydro-js and benchmark 02_replace1k failed: checkElementContainsText tbody>tr:nth-of-type(1)>td:nth-of-type(1) failed. expected 1001, but was 1
Executing frameworks/non-keyed/hydro-js and benchmark 05_swap1k failed: checkElementContainsText tbody>tr:nth-of-type(999)>td:nth-of-type(1) failed. expected 2, but was 999
run was not completely sucessful Benchmarking failed with errors
rebuild-check-single failed for non-keyed/hydro-js

Update: Seems to run locally. Check if performance might be too slow
Update 2: Works locally also for the smoketest and puppeteer...

Krutsch:
Hey, will Check it! I actually did not check the non-keyed version and will fix it if I spot anything. 🙏🏼

krausest:
Thanks! Here are the results (keyed results - left new, right old for updated implementations):
image

Running non-keyed failed on my laptop too so it's not only the virtual server that failed. Not sure yet if it's the test driver or the implementation.

krausest:
Not sure why I thought I saw it work yesterday. It doesn't. Here's the problem:
Open http://localhost:8080/frameworks/non-keyed/hydro-js/index.html, click "create 1,000 rows". Those rows appear starting with id 1.
If you click  "create 1,000 rows" again the labels change, but the id stay the same. Thus the test fails that expects to see 1,001 as the ID after the 2nd click.

The same is true for swap rows. The labels swap, but not the ID.

I moved the non-keyed/hydro-js to the broken-frameworks folder.  IMO there should be no pride to also have a (grubby) non-keyed version. So if you just want to keep the keyed version I'm fully happy with that.

Krutsch:
Fixed it in https://github.com/krausest/js-framework-benchmark/pull/2055
Sorry for the bad code in the first place!

PR #2047, opened 2026-07-08
update ivi v5.1.0

leeoniya:
no perf changes expected, hopefully no regressions.


leeoniya:
ok, added a couple small data store optimizations (i hope)

krausest:
Thanks and nice to see you! Here are the results (keyed results - left new, right old):
image
Vanillajs as the baseline was 0.2 slower, so I'd say ivi 5.1 is almost as fast as ivi 4.

PR #2048, opened 2026-07-08
Add spark-html (keyed) implementation

wilkinnovo:
Adds a keyed implementation for **spark-html** ([spark-html.dev](https://spark-html.dev), [npm](https://www.npmjs.com/package/spark-html)) — a no-build reactive HTML framework: the `.html` file the author writes is what runs in the browser (no compiler, no virtual DOM, single-file components). This implementation uses spark-html 1.1.0 from npm.

**Implementation notes**
- Keyed via `<template each="row in rows" key="row.id">` — the framework's standard keyed-list syntax (LIS-based reconciliation internally; a swap is 2 DOM moves).
- Row markup matches the vanillajs reference byte-for-byte, including `aria-hidden="true"`.
- State updates are the idiomatic immutable style (`rows = next`); `select` is plain state (`selected = id`) read by a per-row `:class` binding — no per-row selected flags (#800), no client-code event delegation (#801; handlers are per-row `onclick` attributes in the template), no direct DOM manipulation in app code (#772).
- Built with the framework's own tool (`spark-html-bun`) via `npm run build-prod`; `customURL` points at `dist`. Framework version resolves from npm via `frameworkVersionFromPackage`.
- `isKeyed` passes for run/remove/swap locally.

Happy to adjust anything that doesn't match current conventions.

### DIFF
--- a/frameworks/keyed/spark-html/index.html
+++ b/frameworks/keyed/spark-html/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8"/>
+    <title>spark-html-keyed</title>
+    <link href="/css/currentStyle.css" rel="stylesheet"/>
+</head>
+<body>
+<div id='main' import="components/app"></div>
+<script type="module" src="/src/main.js"></script>
+</body>
+</html>
--- a/frameworks/keyed/spark-html/package-lock.json
+++ b/frameworks/keyed/spark-html/package-lock.json
@@ -1,24 +1,24 @@
 {
   "name": "js-framework-benchmark-spark-html",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "js-framework-benchmark-spark-html",
-      "version": "1.1.0",
+      "version": "1.2.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "spark-html": "^1.1.0"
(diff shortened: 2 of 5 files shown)


wilkinnovo:
Bumped to spark-html 1.2.0 (published on npm). Internal-only performance release — no changes to the benchmark implementation itself, just the framework version.

wilkinnovo:
Bumped the implementation to spark-html 1.5.0 (npm latest). Since the PR was opened the runtime's local paired numbers vs the vanillajs reference improved from 3.46× to 1.29× CPU geomean (count=15, windowed, official webdriver-ts harness) — same implementation shape, keyed, no build-time codegen. Happy to adjust anything needed for review.

wilkinnovo:
Bumped the implementation to spark-html 1.6.0 (published today). Local paired run vs the vanillajs reference (count=15, windowed, official webdriver-ts harness): CPU geomean 1.239x, run-memory 1.45x. No implementation-file changes beyond the dependency bump — same keyed impl, no build step.

krausest:
Thanks! Here are the results (keyed results - left new, right old for updated implementations):
image
Not in the top tier, but ahead of react and angular.

wilkinnovo:
Thanks for running it and for the merge! Those numbers were v1.7.0 — I've opened #2051 to bump the impl to **1.8.1**. The main release since (1.8.0) narrows in-place mutations to a dirty-key update lane, which targets the weakest op in your run (partial update / every-10th-row, 1.57) and helps swap. Same single no-build 18 KB artifact, no impl-code change — a re-run when convenient would be much appreciated. 🙏

PR #2049, opened 2026-07-12
feat: add pure-vapor

shenjunjian:
Pure Vapor is a reimplementation of Vue's Vapor logic built with AI, intended for learning and research purposes. It only rewrites the code for runtime-vapor, allowing it to reuse the existing Vue ecosystem.


krausest:
What's the point of pure-vapor? Why would anyone use it instead on vue-vapor?

PR #2050, opened 2026-07-12
chore: upgrade vue versions

shenjunjian:
Use version 3.5.39  for standard Vue mode.
Use version 3.6.0-beta.17 for Vapor mode.


krausest:
image

PR #2051, opened 2026-07-12
spark-html (keyed): bump to 1.8.1

wilkinnovo:
Bumps the spark-html keyed implementation from 1.7.0 → 1.8.1 (dependency + lockfile). No implementation-code changes — the app builds unchanged (`build-prod` verified locally with 1.8.1).

Thanks again for running #2048! The results you posted were for v1.7.0 (geomean 1.32). The releases since then target exactly the ops that were spark-html’s weakest there:

- **1.8.0** — in-place array/object mutation of a top-level state value now rides a narrow dirty-key update lane instead of a full component pass. This directly improves **partial update** (the "every 10th row" op, spark-html’s worst relative factor at 1.57 in the v1.7.0 run) and helps **swap rows**.
- **1.8.1** — a correctness fix (same-event `bind:` + `onXXX` handler ordering); no perf change.

The runtime is still the same single no-build 18 KB-gzip artifact, API-frozen for 1.x — so this is a safe drop-in re-benchmark. Would appreciate a re-run when convenient. 🙏

### DIFF
--- a/frameworks/keyed/spark-html/package-lock.json
+++ b/frameworks/keyed/spark-html/package-lock.json
@@ -1,24 +1,24 @@
 {
   "name": "js-framework-benchmark-spark-html",
-  "version": "1.7.0",
+  "version": "1.8.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "js-framework-benchmark-spark-html",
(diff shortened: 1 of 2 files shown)


krausest:
image

PR #2052, opened 2026-07-12
Update sprae to 13.5.0

dy:
Follow-up to #2043 — updates sprae to 13.5.0 in keyed and non-keyed.

Perf-focused release: bulk row teardown on clear/replace (single `replaceChildren` instead of per-row removes), direct list store patching without per-index proxy traps, `Symbol.unscopables` fast path in scope proxies, inert event teardown on disposed rows.


krausest:
image

PR #2056, opened 2026-07-15
Fix react-supergrain: use published @supergrain/kernel package

### LINKED ISSUE #2053: Supergrain needs an update
I can't build supergrain anymore.
The packages @supergrain/core and @supergrain/react don't exist in npmjs.org anymore

I moved react-supergrain to the broken-frameworks folder.

@scottmessinger Can you take a look at it?

scottmessinger:
Fixes #2053. The @supergrain/core and @supergrain/react packages were unpublished from npm; supergrain now ships as @supergrain/kernel (with a react subpath export). Move react-supergrain back out of broken-frameworks and port it to @supergrain/kernel@7.1.0.

@krausest I think this should fix it!


krausest:
image

PR #2058, opened 2026-07-19
Add No.JS keyed implementation

ErickXavier:
Adds a keyed implementation for [No.JS](https://no-js.dev), an HTML-first reactive framework where state and behaviour are declared as HTML attributes with no client-side build step.

### Implementation notes

- `index.html` is served from the framework root, no `dist/`, no `customURL`.
- The framework bundle is vendored at `src/no.js` (v1.20.0). No.JS is distributed via CDN rather than npm, so `frameworkVersion` is pinned literally as documented in README section 4.3. `build-prod` is therefore a no-op, and `includeInBuild` is set to `index.html:src` so `npm run zip` picks the sources up.
- Benchmark logic lives in `src/main.js`. It is deliberately external rather than inline so the implementation runs under the CSP test's `default-src 'self'` policy.

### Rules checklist

- DOM structure matches the vanillajs reference, including `aria-hidden` on the remove glyphicon and the `preloadicon` hint.
- All button ids are unchanged.
- Selection is kept as a single row id in the component state and used for rendering, rather than a per-row flag (#800).
- No explicit event delegation (#801), no `requestAnimationFrame` in client code (#796), no manual DOM manipulation (#772).
- No dependencies, no pre/post-install scripts, no gzipped files.
- Only files under `frameworks/keyed/no-js/` are touched.

### Validation

`npm run rebuild-ci keyed/no-js` passes:

```
no-js-v1.20.0-keyed is keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as keyed in the results
======> Please rerun the benchmark: npm run bench keyed/no-js
```

The CSP check passes, so no `1139` flag is set.


krausest:
image

ErickXavier:
Thank you for letting me know.

PR #2062, opened 2026-07-24
Add Magnet

vilarfg:
Adds a keyed implementation for [Magnet](https://gitlab.com/magnetjs/monorepo), a signal-based UI framework that renders real DOM nodes directly — no VDOM, no compiler. Published as [@magnet/ui](https://jsr.io/@magnet/ui) on JSR and [@magnet-js/ui](https://www.npmjs.com/package/@magnet-js/ui) on npm.

## Implementation notes

- Rows are real `<tr>` elements created once per data item; Magnet's keyed list update (`updateList`) reuses and moves nodes by object identity, so append/remove/swap touch only the affected rows.
- Signal implementation is pluggable; this build uses the Preact-signals adapter from [@magnet-js/signal](https://jsr.io/@magnet/signal).
- Expected review notes: identity-keyed DOM-node reuse is the framework's core mechanism (same family as other keyed implementations), and row selection state is held as a per-row signal (Note #800-style).


krausest:
Thanks. Seems like I have to add two issues:
Note #800 (view state on model): YES — Each row object holds . The selected state is stored per-row on the data object, not as a single top-level variable.

Note #1261 (manual DOM node caching): YES —  pre-creates the DOM element for each row and stores it as row.el = html.tr(...). The sync() function reuses these cached elements: xrows.set(data.map((row) => row.el)). This avoids recreating DOM nodes on operations like swap/clear/add.

krausest:
image

PR #2063, opened 2026-07-26
Update Ripple to 0.3.112

trueadm:
- update `ripple` and `@ripple-ts/vite-plugin` from 0.3.65 to 0.3.112
- migrate the benchmark component to the current TSRX function, statement-container, `@for`, text, and expression syntax


krausest:
Even a bit ahead of octane. Here's a preview of the results:
image

PR #2064, opened 2026-07-26
Add Octane

trueadm:
Octane benchmark using Octane 0.1.16


krausest:
Not bad! Here's a preview of the results:
image

PR #2065, opened 2026-07-27
Add grainular/nord (keyed) implementation

IamSebastianDev:
Adds a keyed benchmark implementation for [Nørd](https://docs.nordjs.dev) using:

- `@grainular/nord`@2.0.0
- `@grainular/grains`@2.0.0

Implementation details:

- Rows are keyed by their stable item ID using `$each(...).$withKey(...)`.
- Selection is stored as a single selected ID and derived for each row.
- Labels use writable grains for fine-grained partial updates.


krausest:
Thanks. Here's a preview of the results:
image

PR #2066, opened 2026-07-27
Update sprae to 13.7.1

dy:
Follow-up to #2052 addressing the select/swap/create feedback — updates sprae to 13.7.1 in keyed and non-keyed.

Engine changes since 13.5.0:
- Read-only expressions compile to destructured property reads instead of `with` — identifier resolution becomes optimizable (select-row JS −68%)
- `:each` rows: prototype-chain scope replacing the per-row proxy, generation-stamp keyed diff, flat directive-program replay for clones (create/replace JS −25%+, swap framework JS −63%)
- 2-row swap moves siblings directly instead of cycling a text-node placeholder through `<tbody>`
- Single-directive fast path and direct bare-event listeners — less per-row machinery, same DOM semantics


krausest:
Thanks. Here's a preview of the keyed results:
image

PR #2068, opened 2026-07-28
react-supergrain: update to @supergrain/kernel 7.3.0

scottmessinger:
Hi @krausest! Thanks again for all your work on this. This is a small update for react-supergrain.

### DIFF
--- a/frameworks/keyed/react-supergrain/package-lock.json
+++ b/frameworks/keyed/react-supergrain/package-lock.json
@@ -1,14 +1,14 @@
 {
   "name": "js-framework-benchmark-react-supergrain",
-  "version": "2.0.0",
+  "version": "2.1.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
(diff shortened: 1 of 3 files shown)


krausest:
`npm run isKeyed -- --runner playwright --headless true keyed/react-supergrain`yields an error:
```
react-supergrain-v19.2.4 + 7.3.0-keyed is non-keyed for 'run benchmark' and keyed for 'remove row benchmark' and keyed for 'swap rows benchmark' . It'll appear as non-keyed in the results
ERROR: Framework react-supergrain-v19.2.4 + 7.3.0-keyed is not correctly categorized
```
Can you take a look at it? Typically this means you're not really using a keyed approach.

scottmessinger:
@krausest Sure thing! Sorry about that!

scottmessinger:
@krausest I think it's fixed now!

krausest:
image

PR #2070, opened 2026-07-28
Update sprae to 13.8.4

dy:
Follow-up to #2066 — updates sprae to 13.8.3 in keyed and non-keyed.

Engine changes since 13.7.1:
- reactive keyed row index (moved/spliced rows update index bindings)
- effect re-runs dedupe signal subscriptions with a single membership probe; single-subscriber signals hold their effect inline, no Set allocation (select-row JS −24%, leaner rows)
- `:each` defers post-mount list updates to a microtask — multi-write mutations like swap diff in one pass; initial mount stays synchronous
- fix: out-of-range list reads no longer extend the raw array


krausest:
image

PR #2072, opened 2026-07-30
Add ValyrianJS keyed and non-keyed implementations

Masquerade-Circus:
This PR adds keyed and non-keyed implementations for Valyrian.js.

Both versions use the same TSX implementation. The only difference is that the keyed version adds `key={item.id}` to each row.

The implementation uses Valyrian.js directives such as `v-for`, `v-text`, and `v-keep`.

The build uses Valyrian.js’s inline build tool to produce an optimized bundle with minimal configuration.


krausest:
Here's the result of the incremental run. Seems to be slower than react & angular. 
image

PR #2074, opened 2026-08-01
bump preact to latest

JoviDeCroock:
Bump preact to 10.29.8 as we've done some perf work


krausest:
Thanks - here's a comparison (left new keyed results from incremental run, on the right previous run):
image
Maybe a tiny bit faster...

PR #2075, opened 2026-08-02
Keyed magnet v0.4.0

vilarfg:
Thank you @krausest for the review and merge of the first submission of Magnet.

It was tagged #800 and #1261 .

This new submission uses Magnet's own list utility which, I believe, should do away with both tags.

cheers


krausest:
image

vilarfg:
cool, thanks!

I've noticed the two tags: 800 and 1261 show up in the image.

Is it because the latest submission still warrants them?

cheers!

krausest:
Yes, I think you're right. I'll remove them.

PR #2076, opened 2026-08-04
bump hydro-js to 1.9.4

Krutsch:
Hope I am not spamming too much recently. Did some more promising work:
image


krausest:
image

PR #2077, opened 2026-08-05
Add keyed domphy implementation

huukhanhnguyen:
## Summary

Adds a keyed implementation for **Domphy** (`@domphy/core` 0.21.1, published on npm) — a patch-based, framework-agnostic UI runtime: plain-object element trees keyed by HTML tag, listener-based fine-grained reactivity, no JSX, no virtual DOM.

- Implementation notes: one keyed `State<Row[]>` list with `_key` per row; per-row `label` states so `update every 10th` touches only those rows; a single table-level `selected` id state (per the contribution notes on selection style); per-row descriptors cached so unchanged rows skip patching. Markup mirrors `frameworks/keyed/vanillajs` including the `aria-hidden` / glyphicon preload span.
- `npm install && npm run build-prod` installs purely from npm (lockfile included, `@domphy/core` 0.21.1 from the registry).
- `isKeyed` check passes (keyed for run / remove row / swap rows); official webdriver-ts playwright runner completed with a successful plausibility check.

Local measurement (official runner, headless Chrome on Windows, no throttling — indicative only): select/swap/remove/update land within ~0.95–1.6× of vanillajs-keyed; create/clear ops at ~2–3.4×. Full numbers available on request; a re-run on the reference hardware will be more meaningful.

Project: https://github.com/domphy/domphy · https://domphy.com

### DIFF
--- a/frameworks/keyed/domphy/README.md
+++ b/frameworks/keyed/domphy/README.md
@@ -1,0 +1,25 @@
+# Domphy — keyed implementation
+
+[Domphy](https://domphy.com/) is a patch-based, framework-agnostic UI runtime.
+No JSX, no virtual DOM: UIs are plain objects keyed by HTML tag and reactivity
+is listener-based (`toState`).
+
+This implementation uses the framework's idiomatic fine-grained style:
+
+- one keyed `State<Row[]>` for the row list (`_key` per row drives the keyed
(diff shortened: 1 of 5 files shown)


krausest:
Here's the result of the incremental run. Could be a little faster though...
image

PR #2078, opened 2026-08-05
Add WildflowerJS keyed benchmark entries (data-list and data-pool)

wfjs-admin:
Two entries for WildflowerJS (https://wildflowerjs.com), a no-build-step reactive framework with no virtual DOM and two reactivity models:

- wildflowerjs: the general-purpose data-list reactivity model, the default declarative way to render lists in the framework.
- wildflowerjs-pool: a separate data-pool reactivity model built for high-frequency mutation workloads, also a valid and often simpler choice for ordinary lists.

Each entry's README explains why the two are kept separate rather than folded into one. Both are CSP-safe with zero reported violations (data-csp-safe on the framework script tag), pass npm run isKeyed for run/remove/swap, and validate cleanly end to end.


krausest:
Here's the result of the incremental run. Not bad!
image

wfjs-admin:
Excellent, thank you! These results are in line with my experience, so I'm glad to see it holding up in your testing!

Much appreciated!

PR #2079, opened 2026-08-07
fix(svelte): build with the production export condition

baseballyama:
Without `production` in the export conditions, `esm-env` resolves to its dev fallback, which reads `process.env.NODE_ENV` at runtime. `DEV` then stops being a compile-time constant, so terser keeps every `if (DEV)` branch and the whole Svelte development runtime ends up in the production bundle.

Reduces the brotli-compressed bundle by ~18-20% for keyed/svelte, keyed/svelte-classic and non-keyed/svelte-classic.


krausest:
Thanks. I'm usually a bit skeptical, but you're right (left new results, right previous results):
image
The bundle really size got smaller. I don't know why first paint got worse (can't say wether it's a artifact of the incremental run or really got worse).

PR #2081, opened 2026-08-09
Update Solarite to 0.8.0

eric-frost:
New Solarite that's faster but a little larger.


krausest:
image

PR #2086, opened 2026-08-17
Add keyed PyJS implementations

eukreign:
## Summary

Adds two keyed PyJS implementations:

- `pyjs` uses idiomatic Python with direct DOM operations.
- `pyjs-ui` uses PyJS custom elements and reactive state.

Both are ahead-of-time compiled from Python to JavaScript. No Python runtime is shipped to the browser. The builds use the fixed `@pyjsdev/cli` 0.3.0 release.

## Validation

Ran the upstream focused validation from a clean dependency install:

```console
npm run rebuild-ci -- keyed/pyjs keyed/pyjs-ui
```

The smoke benchmarks and plausibility checks passed. Both implementations passed keyed classification for run, remove, and swap, and both passed CSP validation.

The upstream size check reported:

- `pyjs`: 7,910 bytes uncompressed; 2,577 bytes compressed.
- `pyjs-ui`: 8,336 bytes uncompressed; 3,127 bytes compressed.

All changes are contained within `frameworks/keyed/pyjs` and `frameworks/keyed/pyjs-ui`.


krausest:
There's one issue with my new build system: 
`pyjs: The PyJS npm prototype does not support linux-arm64 yet.`
Can you provide a prebuilt binary (keyed/sauron can serve as an example)?

eukreign:
> There's one issue with my new build system: `pyjs: The PyJS npm prototype does not support linux-arm64 yet.` Can you provide a prebuilt binary (keyed/sauron can serve as an example)?

Thanks for considering my PR. I have added a linux-arm64 npm build and updated this PR to point to the new version. Please give it another try. Thank you!

krausest:
image
Amazing performance for cross-compilation! 
image
Even memory and startup look pretty good!

eukreign:
Thanks, it's able to achieve this through very deep static analysis of the Python source and outputting close to handwritten JavaScript. For the DOM performance, it should be comparable to Lit, since pyjs.ui is just thin wrapper of over native custom elements like Lit.

I'm also re-using the same static analyzer to compile Python with LLVM to native code; separately I can produce JVM bytecode and then run that on Android. That will be in subsequent releases of pyjs!

PR #2087, opened 2026-08-21
chore(vue): drop data-label and delegate the per-row click handlers

edison1105:
### Why `data-label` is removed

The `<tr>` in the Vue implementations carried a `:data-label="label"` binding
that no other implementation has. Out of 253 implementations in this repo, only
the Vue-family entries (and alins) have it — the vanillajs reference row is just
`<td class='col-md-1'>…` with no data attribute, and nothing in the repo reads it:
not the test driver, not the results app, not the CSS.

### DIFF
--- a/frameworks/keyed/vue-jsx-vapor/src/App.jsx
+++ b/frameworks/keyed/vue-jsx-vapor/src/App.jsx
@@ -106,7 +106,7 @@ export default defineComponent({
 
         <table class="table table-hover table-striped test-data">
           <tbody>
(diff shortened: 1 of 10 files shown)


krausest:
I'll apply that after the next full run (~ soon) to measure the effect.

krausest:
I merged your PR and plan to run a comparison old vue/vue-vapor vs your updated versions. I think that's more interesting than just the results of an incremental run.

krausest:
It helps indeed a bit:
image
The comparisons show that the difference is statistically signifcant (white column vs. column to the right. The red benchmarks are significantly slower):
image
image

PR #2094, opened 2026-09-09
chore(marionette): update keyed implementation to v5 beta1

paulfalgout:
## What

- Update the existing keyed Marionette entry from `5.0.0-alpha.2` to the published `5.0.0-beta.1` packages. Keep version detection through `frameworkVersionFromPackage`, without an adapter-specific name or extra entry.
- Replace the hand-written collection shim with `@mnjs/data` and the local renderer with the published `@mnjs/adapters/dom/morphdom` adapter and a string row template. Remove Underscore and the obsolete renderer file.
- Use beta1's public APIs for collection moves, child-view swapping, model lookup, and declarative root-state updates. Selection remains one ID on the collection; events use Marionette's event delegation.

## Why

Marionette v5 beta1 is now available on npm. This brings the alpha2 entry up to date with its supported data and DOM adapters rather than maintaining benchmark-specific substitutes. The benchmark continues to exercise Marionette Views, CollectionView, model events, and lifecycle management.

## Scope

Only `frameworks/keyed/marionette` changes. The Backbone.Marionette v4 entry, other frameworks, benchmark harness, and result files are unchanged.

## Deployment Impact

Rebuild only `keyed/marionette` through the usual benchmark build process. The published beta1 packages require Node.js 24 or newer. No shared application or other framework deployment changes are needed.

## Testing

- `npm run compile-webdriver-ts`
- `npm run rebuild-ci keyed/marionette` — clean lockfile install, production build, headless benchmark smoke tests, keyed/HTML checks, and CSP checks.
- `git diff --check`
- Earlier local full headless CPU, memory, size, and first-paint measurements used an identical production JavaScript bundle; these are local exploratory results, not official benchmark results, and are not included in this PR.


krausest:
image

PR #2095, opened 2026-09-09
Add Farm.js (React) keyed implementation

Kinfe123:
adds a keyed implementation for farm.js (https://farmjs.dev), a full-stack framework whose react renderer includes an experimental ahead-of-time compiler. the component source is idiomatic react (useState, keyed map, row-local ternaries) with no benchmark-specific code; the compiler proves DOM targets at build time and emits direct bindings for them. the build uses onUnsupported: "error" so it can never silently fall back to baseline react, which keeps the entry honest about what it measures.

- builds from the published npm packages with npm install and npm run build-prod (vite)
- verified locally: run/runlots/add/update/select/remove/swaprows/clear all behave per the contract, and swap moves the actual TR nodes (keyed identity checked by marking dom nodes before swapping)
- bundle is 69.3 kB gzip
- follows the vue entry layout (vite + customURL /dist)

happy to adjust anything to fit the repo conventions.


krausest:
image

Kinfe123:
thanks for the preview run! that swap number flagged a real problem: beta.3 predates the runtime's keyed reorder/removal fast paths, so a 2-row swap was falling back to rebuilding rows. opened #2102 bumping the entry to 0.1.0-beta.11, which has those paths. rebuilt this exact entry locally both ways to confirm: swap drops from ~117ms to ~32ms under 4x throttle, remove row roughly halves, create ~1.4x faster. would appreciate it if the bump could land before the next full run so the table reflects the current runtime.

PR #2096, opened 2026-09-12
Update WildflowerJS entries to v1.5.1

wfjs-admin:
Updates both WildflowerJS entries to the 1.5.1 release (https://www.npmjs.com/package/wildflowerjs/v/1.5.1).

- `wildflowerjs` (data-list): bundle updated to `wildflower.mini.min.js` from 1.5.1. 1.5.1 is a performance release for the list renderer's create, replace and clear paths.
- `wildflowerjs-pool` (data-pool): now loads the framework's `mini-pool` tier (`wildflower.mini-pool.min.js`, 213 KB uncompressed) instead of the `lite` tier (301 KB). It is the build an application choosing `data-pool` would ship, so the entry now measures what users get. README updated to say so.
- `frameworkVersion` set to 1.5.1 in both `package.json` files.

No changes to `Main.js` or the markup; the implementations are unchanged. Both bundles are the published npm files, byte for byte.

Validated locally: `npm ci` and `npm run build-prod` for both entries, the headless smoke test over benchmarks 01 to 09, 21, 22, 25 and 40 (PlausibilityCheck: successful run), and `npm run isKeyed` (both keyed for run, remove and swap).


krausest:
image

wfjs-admin:
Sweet :-) This holds against my benchmark runs here. Look forward to the next official run. Thank you!

PR #2097, opened 2026-09-15
Keyed Places.js v0.1.0

UnknownOrange:
Places.js is a framework for building HTML-first websites with progressive enhancements to provide interactivity. It has these key features.

-     No build step required.
-     HTML-first templating with fine-grained reactivity.
-     Cross-component state management using signals.
-     Data fetching.

Here are the results from running benchmarks on my computer.
Memory_and_Size_Benchmarks
DurationBenchmarks

More details about Places.js can viewed [here](https://codeberg.org/createthirdplaces/places-js/src/branch/main/README.md)


krausest:
Thanks. What's the purpose of not publishing a npm package? This would allow the user to actually manage the dependency (e.g. checking for CVEs in the specified version) instead of including something...

krausest:
image

UnknownOrange:
> Thanks. What's the purpose of not publishing a npm package? This would allow the user to actually manage the dependency (e.g. checking for CVEs in the specified version) instead of including something...

I agree with you about the value of a dependency management tool. For now, I have the framework on a [jsDelivr CDN](https://cdn.jsdelivr.net/gh/createthirdplaces/places-js@0.1.0/places-js-latest.js) while I look for an alternative. 

 I’m not sure if npm is the best fit for Places.js. I'm thinking about using [jsr.io](http://jsr.io/) or [Artifactory](https://jfrog.com/artifactory/)

PR #2098, opened 2026-09-15
Update keyed/thane to 0.2.0

timlouw:
Dependency bump to the newly published thane@0.2.0 (pinned, no range) with a regenerated
package-lock.json. The benchmark implementation itself (benchmark.ts, main.ts, index.html)
is unchanged from the 0.1.1 submission; vite.config.js is removed because the folder has
built with the Thane CLI since 0.1.1 and nothing reads it.

What changed in the framework: 0.2.0 is a compiler and runtime release. The keyed reconciler
gained fast paths (identity-first matching for same-length updates, an append path, an
identity-aware reorder path, batch row creation, a lazily built key map), row updates are
guarded so unchanged bindings are not written, and sole-content text is updated through
Text.nodeValue. The benchmark code keeps one selected id in state and uses no manual DOM
manipulation, no explicit event delegation and no requestAnimationFrame; delegation happens
inside the compiled output, not in client code.

Checked locally: npm install, npm run build-prod, npm run rebuild-ci keyed/thane
("All checks are fine"), npm run isKeyed keyed/thane.


krausest:
image

timlouw:
> Thanks. Here's a preview from an incremental run:
> image
> 

Thanks a million. That's an exciting result!

krausest:
Indeed. Excited to see the next full run!

PR #2100, opened 2026-09-19
Update DriftJS to 0.0.14 and point framework URL to driftjs-org

hrutavmodha:
This PR updates DriftJS to its latest optimized version


krausest:
image

PR #2101, opened 2026-09-19
Refresh the results website and add SEO metadata

nigrosimone:
Redesign website with better (better for my eye) UI and SEO. Take your time, it's a huge PR

Screenshot 2026-09-19 130901

Screenshot 2026-09-19 130928

Screenshot 2026-09-19 130948

Screenshot 2026-09-19 134240

Screenshot 2026-09-19 134302

Screenshot 2026-09-19 134346


UnknownOrange:
As someone who recently started using this repo for benchmarking my framework, I prefer the existing design [krausest](https://github.com/krausest) decided to use. 

I think your design has too much filler text, and I found that the UI for the existing version was easy to understand as a new user.

docs/WEBSITE.md
```diff
@@ -0,0 +1,57 @@
+# Results website
+
+The official website is hosted at <https://krausest.github.io/js-framework-benchmark/> in the separate `krausest.github.io` repository. This repository contains its static overview and the interactive results application.
```
UnknownOrange, line 3: As someone who recently started working with the code in this repo, I found the author's docs to be easy to follow.

I think these just add confusion.

webdriver-ts-results/public/sitemap.xml
```diff
@@ -0,0 +1,216 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
UnknownOrange, line 1: What is the purpose of this file?

webdriver-ts-results/overview.html
```diff
@@ -0,0 +1,589 @@
+<!doctype html>
+<html lang="en">
+  <head>
```
UnknownOrange, line 3: I often refer to the benchmarking site as a user, and don't want to see another page adding UI clutter.

webdriver-ts-results/public/sitemap.xml
nigrosimone: Used by search engine for indexing alla pages

nigrosimone:
@UnknownOrange Thank you for the review. Let's see if any other opinions come in. However, I want to emphasize the value of the SEO-related changes; right now, the site isn't very readable for indexing bots. The visual design is a matter of taste and might not appeal to everyone.

krausest:
Thanks a lot. Very appreciated - I'll take a deeper look at it after the next incremental run.

krausest:
I'll post a few comments. Please let me know if any of my ideas commit crimes against your design.
I adjusted the header since I noticed I called this project jsfb sometimes.
image

krausest:
The table is too large for me to take a screenshot that shows the CPU result. I'm posting this regularly and I think the whole CPU table should fit on one screen.

To fit on the screen this would mean adjusting padding e.g. to 1px 4px
image

krausest:
The current temporary results are updated :)

Looking for feedback...

krausest:
I pushed the changes I'd propose.

nigrosimone:
Fine for me. The only two things I would add on top of what is there now:
- A composite score with a final ranking, to make the reading simpler. It could be the sum of the scores of each area (CPU, memory, startup), where the best gets 100 and the worst 0.
- Highlight the default implementation of each framework, the one you get installing it from npm, so the user understands what they get without special configuration.

krausest:
Once again thanks for the styling. I think we'll switch to the new layout (also for the overview page, but I put my focus on the result table so far).
I put some decisions regarding the new style here and hope to get some feedback what others think:
https://github.com/krausest/js-framework-benchmark/issues/2103

krausest:
(and whether we include the llms.txt I'll decide when Anthropic and OpenAI will have decided regarding my free open source license application 😄 )

